PR #121, opened 2020-09-05
add pundit gem and a base application policy

### LINKED ISSUE #120: add the Pundit Gem! Tessa!
install Pundit gem

tnewmanheggie:
Resolves #120 
This adds the pundit gem and a base application policy. 
Needs rules.


robbkidd (approved):
we should have rules

PR #122, opened 2020-09-05
Strip formatting when searching for a phone number

reid-andrew:


jim (approved):
Thanks for tackling this. It looks good!

A nice-to-have would be to add a test to `tests/models/members_test.rb` that directly verifies the `Member.matching` scope returns the members we expect it to.

reid-andrew:
> Thanks for tackling this. It looks good!
> 
> A nice-to-have would be to add a test to `tests/models/members_test.rb` that directly verifies the `Member.matching` scope returns the members we expect it to.

Good catch - I'm happy to refactor to add in model tests before this gets merged.

jim:
Awesome!

PR #124, opened 2020-09-05
add concept of roles to Member

robbkidd:
We'll be figuring out member access privileges through Member instances. This is a minimum viable role implementation to determine that a Member is a member. More roles to come after linking Members to User logins.


cflipse:
lgtm

jim (commented):
👍

Issue #105, opened 2020-09-02
Member views profile/user info


rkim00:
As a member, I want to be able to view my user information, so that I can see if it's incorrect or so that I can look up my member information.

What constitutes user info?
editable
- name
- pronouns
- e-mail
- telephone
- zip code
- street address?

not-editable (display only)
- membership status
- membership no.
- membership expiration date

See babywearing repo for code to reuse. 

See bottom of [Member portal mock-ups](https://drive.google.com/file/d/172gIc-yOgd-yAxbv4_qnVnC0JwMqacXo/view).

Related to #106.

jim:
This was closed by #131

PR #131, opened 2020-09-05
Member profile view creation #105

johncarlocerna:
[https://github.com/rubyforgood/circulate/issues/105](https://github.com/rubyforgood/circulate/issues/105)
Authors: @jim and @johncarlocerna 
Creating member profile page view for accounts. URL at /member_profile


jim (approved):
This LGTM, but since I worked on it, I would love to get a 👍 from someone else before merging it!

johnksawers (approved):

johnksawers:
In another PR, we created /member/loans and /member/history - it might be worth considering moving this to /member/profile (or /member) or something like that. Not a big change, and we can even do it after this is merged.

jim:
@johnksawers Good catch. I'm going to make a new story to go back and fix all the routes once and organize all the new controllers once we get the member UI in place.

PR #133, opened 2020-09-06
Remove broken migration with missing tags relation

cattywampus:
When running the `db:migrate` task from beginning to end, the migration
breaks on `20200420175849_rename_tags_to_categories.rb` with an error
that there is no `tags` relation. Digging into this a bit more, it looks
like this migration exists to counteract the from commit 96d3342 which
set out to rename categories to tags. That work was subsequently
reverted in commit 897d43a which deleted that migration. This remaining
artifact is to rename tags back to categories but there is no tag table
as the code was never long lived in the baseline from what I can tell.

After deleting this migration and re-running the rake task, there appear
to be no functional changes to any of the content in the `schema.rb`.
The few changes that do appear are adding trailing commas and re-arrange
the custom views, but the actual sql view code appears to be consistent.

Outside of the motivation for having a full working history of
migrations, this bug was blocking the ability to deploy to heroku
because the `release` target in the `Procfile` was also running the
`db:migrate` task on deployments, which failed. This should unblock that
and allow the heroku work to continue.

Reference:

https://github.com/rubyforgood/circulate/commit/897d43a0a84f2b8e849fac2cf5191c16471d57b0#diff-a043fb9a41e05dcf884cd74e4ec0a3e5

Co-authored-by: Robb Kidd <robb@thekidds.org>

### DIFF
--- a/db/migrate/20200420175849_rename_tags_to_categories.rb
+++ b/db/migrate/20200420175849_rename_tags_to_categories.rb
@@ -1,9 +1,0 @@
-class RenameTagsToCategories < ActiveRecord::Migration[6.0]
-  def change
-    rename_table :tags, :categories
-    rename_column :categories, :taggings_count, :categorizations_count
-
-    rename_table :taggings, :categorizations
-    rename_column :categorizations, :tag_id, :category_id
-  end
-end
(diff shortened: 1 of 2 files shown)


robbkidd (approved):
you can remove it

Issue #41, opened 2020-04-22
Deleting a category breaks the item history page


jim:
Item history pages attempt to load any categories that were previously assigned to the item. If one of these isn't found, the page breaks:

`Couldn't find all Categories with 'id': (1, 4, 18, 50) (found 3 results, but was looking for 4).`

This situation should be handled gracefully and ideally display `deleted category`.

PR #137, opened 2020-09-06
Swap find for where to grab category names

britth:
Using `find` throws an error on an item's edit history page when the category has been deleted. Updates the query to use `where` instead. If the number of categories found is less than the number of keys, add the string `deleted category` to signify that there was also a now missing value associated with the item. Adds a test to exercise this behavior.

Issue #41

ruby -I ./test test/system/items_test.rb output:

Finished in 10.779663s, 0.6494 runs/s, 1.4843 assertions/s.
7 runs, 16 assertions, 0 failures, 0 errors, 1 skips

### DIFF
--- a/app/helpers/admin_helper.rb
+++ b/app/helpers/admin_helper.rb
@@ -67,7 +67,12 @@ module AdminHelper
       value = BorrowPolicy.find(value).complete_name
     when "category_ids"
       key = "categories"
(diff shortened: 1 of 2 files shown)


jim (approved):
Thanks for finding this 🐛 and for fixing it. This looks great.

Issue #126, opened 2020-09-05
Member views loan history


rkim00:
As a Member, I want to be able to view my loan history so that I can make choices about what tools to check out again.

Tool name should link to View Tool page.

Mock-up shows default sort as checked-out, asc, but it should probably be checked-in, desc.

Open questions:
- should currently checked out items also be displayed in this list? They don't need to be for sure, but there may some UX reasons in favor of doing so.
- should member be able to request a pick-up / check-out directly from this list? If so, we may want to display whether the item is available. Or this could happen from the View Tool page.

See [Member portal mock-ups](https://drive.google.com/file/d/172gIc-yOgd-yAxbv4_qnVnC0JwMqacXo/view).

PR #138, opened 2020-09-06
Creates member loan history view

amygurski:
Members can go to `/member/history` to view their personal loan history.

Linked to issue [#126]( https://github.com/rubyforgood/circulate/issues/126)

Regarding the open questions in the issue:
- Should currently checked out items also be displayed in this list? They don't need to be for sure, but there may some UX reasons in favor of doing so.

    Currently checked out items are shown too. Happy to change it if we only want items that have been returned to show.

- Should member be able to request a pick-up / check-out directly from this list? If so, we may want to display whether the item is available. Or this could happen from the View Tool page.

    Added display for current item availability so a future PR could expand to include a button to check-out previously loaned 
    items from this page.

Co-authored-by: Johnksawers  <john@johnksawers.com>
Co-authored-by: Andrew <dev.andrewreid@gmail.com>


amygurski:
Once the member's [checked out items](https://github.com/rubyforgood/circulate/pull/143) PR is merged, I will merge and finish this since there's some overlap in the setup of the member's controller & I think I need to rewrite my tests.

reid-andrew (approved):

cflipse:
I would recommend not putting all of these into a single, shared controller.  You're going to have a _lot_ of things that are just gated on /member, and putting them into one controller is going to result in a very large and convoluted space.

You can define them as nested under a specific namespace, so you can get the same urls:  /member/history and the like, by using the `namespace` or `scope` blocks in the routes file:

```
# routes /member/history to Member::HistoriesController
namespace :member do
  resource :history
end

# routes /member/history to HistoriesController
scope "/member" do
  resource :history
end
```

amygurski:
Hi @cflipse Thanks for the feedback. I read that @jim is planning to standardize the new member routes - Issue [#158](https://github.com/rubyforgood/circulate/issues/158). Therefore, I am inclined not to make this change right now and leave it for that PR which will do not only member history but take into account all the other new member routes. Does that sound OK to you? Thanks again!

jim:
@cflipse Thanks for the review. We'll definitely make that adjustment shortly.

PR #140, opened 2020-09-06
Ensure that single category changees are saved in audit history

britth:
When an item has no categories (whether on creation or if the categories are all cleared), the audit history will not include an entry when a category is subsequently added. Two changes must be made before it is saved. This is because there is a check for presence of cached category ids before initiating the audit addition. This PR updates the logic to ensure that an entry is created anytime the new category ids do not match the cached category ids.

Co-authored with @johnksawers 

Finished in 5.790894s, 29.5291 runs/s, 96.8762 assertions/s.
171 runs, 561 assertions, 0 failures, 0 errors, 2 skips

### DIFF
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -103,7 +103,7 @@ class Item < ApplicationRecord
 
   # called when item is updated
   def audited_changes
-    if @current_category_ids.present? && @current_category_ids != category_ids.sort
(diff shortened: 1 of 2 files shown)


jim (approved):
Looks good! Tightening up this auditing functionality is really wonderful.

PR #141, opened 2020-09-06
Add install step

UnicodeRogue:
Add install step to clarify for folks setting up their dev environment


jim (approved):
Documentation improvements! This is awesome, and future contributors will thank you.

Issue #94, opened 2020-08-27
Deprecation warnings from activerecord-postgres_enum


jim:
Running `rails` commands prints out some deprecation warnings like these:

```
/Users/jimb/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/activerecord-postgres_enum-1.0.0/lib/active_record/postgres_enum/extensions.rb:27: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/jimb/.asdf/installs/ruby/2.7.1/lib/ruby/gems/2.7.0/gems/activerecord-6.0.3.2/lib/active_record/connection_adapters/abstract/schema_definitions.rb:364: warning: The called method `column' is defined here
```

We might need to update the `activerecord-postgres_enum` gem if there is an updated version available.

may be superseded by #117

robbkidd:
After recent updates—maybe [this upgrade to `activerecord-postgres_enum`](https://github.com/rubyforgood/circulate/pull/146/files#diff-e79a60dc6b85309ae70a6ea8261eaf95R68)—I am not seeing these warnings when running `rails db:<anything>`, `rails server`, or `rails console`. Maybe we fixed it?

robbkidd:
Closed by @Sinrock's upgrade of many gems, including `activerecord-postgres_enum`, in #146

robbkidd:
This appears to still be an issue. Deprecation warning still appears when running `rails db` tasks involving the schema (`db:migrate`, `db:schema:load`, `db:setup`, and `db:reset`). Triggered by calls to `create_enum` provided by the `activerecord-postgres_enum` gem.

Related, the `t.monetize` method from `money-rails` used in a past migration has a similar deprecation warning about keyword parameters:

```
== 20190526205528 CreateBorrowPolicies: migrating =============================
-- create_table(:borrow_policies)
/Users/rkidd/code/r4g/circulate/.direnv/ruby/gems/money-rails-1.13.3/lib/money-rails/active_record/migration_extensions/table_pg_rails4.rb:8: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/rkidd/code/r4g/circulate/.direnv/ruby/gems/activerecord-6.0.3.2/lib/active_record/connection_adapters/abstract/schema_definitions.rb:364: warning: The called method `column' is defined here
   -> 0.0158s
[...]

Issue #117, opened 2020-09-02
Upgrade all outdated gems


jim:
`gem outdated` shows a decent number of gems that could be updated.

PR #146, opened 2020-09-06
Updated the outdated bundled gems to newest versions

robbkidd:
From @Sinrock's fork, here's an update for all the gems!

PR #147, opened 2020-09-06
Rails errors

Sinrock:
Checked to make sure the previous rails errors were no longer occurring after the fix of #117 fixing the outdated gems.


robbkidd:
@Sinrock I've updated #94 with [a comment that the deprecation warnings are gone](https://github.com/rubyforgood/circulate/issues/94#issuecomment-687933314) after your upgrade of the `activerecord-postgres_enum` gem in #146 (my redo of your #117).

robbkidd:
Confirmed #146 fixes #94. The changes in this PR are already merged in #146.

PR #162, opened 2020-09-12
User edits password

### LINKED ISSUE #107: Member edits/changes password
As a member, I want to be able to edit my password, so that I can change my password if I'm worried about security or if I'm having trouble remembering my old password.

Note: do we want members to be able to straight type in a new password in a form, or should we have them go through a Forgot Password flow (reset password link, etc.)?

Also -- Babywearing makes this integrated on the same screen as editing other account settings. Is that good/bad for our purposes?

See [Member portal mock-ups](https://drive.google.com/file/d/172gIc-yOgd-yAxbv4_qnVnC0JwMqacXo/view).

See babywearing repo for code to reuse, see mock-ups for general sense of the thing.

See issue re other account settings: #106

reid-andrew:
This allows users to update their passwords. Users will see a button on their "Member Profile" screen to update their password. Users are required to submit a password and confirmation. Messages are displayed for either errors (password mismatch, too short) or success in updating. Also adds tests for happy and sad paths.

Closes #107


app/views/member_profiles/show.html.erb
```diff
@@ -29,7 +29,7 @@
 
     <button class="btn">Edit Account Settings</button>
     <br> <br>
-    <button class="btn">Edit Password</button>
+    <%= button_to "Edit Password", edit_member_profiles_password_url, method: :get, class: 'btn' %>
```
jim, line 32: Since this is a GET request, I think that `link_to` can do the job here:

```suggestion
    <%= link_to "Edit Password", edit_member_profiles_password_url, class: 'btn' %>
```

The CSS should be able to make a link look like a button.

app/views/member_profiles/passwords/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= form_for(@user, :url => { :action => "update" } ) do |f| %>
+  <div class="field">
+    <%= f.label :password, "Password" %><br />
+    <%= f.password_field :password, :autocomplete => "off"  %>
+  </div>
+  <div class="field">
+    <%= f.label :password_confirmation %><br />
+    <%= f.password_field :password_confirmation %>
+  </div>
+  <div class="action_container">
```
jim, line 10: I don't see the class `action_container` mentioned in the CSS anywhere. I might have missed it, though. What does it do?

jim (approved):
This looks good. I found a few small things to potentially tweak.

PR #163, opened 2020-09-12
Users can select more than one pronoun

### LINKED ISSUE #78: Member can specify more than one pronoun set
Currently, members can select `he/him`, `she/her`, `they/them`, or provide custom pronouns when they signup. We should allow users to select more than one pronoun.

This is slightly complicated by the current implementation, which uses [ActiveRecord::Enum](https://api.rubyonrails.org/classes/ActiveRecord/Enum.html) to model this attribute.

I think the path forward is probably to:

* Rename the `pronoun` column `pronouns`.
* Change the column to store an array of values in the database.
* Remove the use of `enum` in `Member` and write a method to return the options.
* The edit form for member records needs to allow for multiple selections.
* The conditional field logic (that toggles the display of the `custom_pronoun` field) might need to change.

tnewmanheggie:
closes #78 
This pull request allows users to select more than one pronoun, and specify a custom pronoun in addition. 
This creates a partial for the pronoun selection that can be added to other member/user signup or edit forms. 
This also adds a test for this feature and fixes the user signup test. 

Screen Shot 2020-09-13 at 10 11 08 AM


jim (approved):
Our members will be so happy to have this improvement. Thanks for tackling this one.

PR #173, opened 2020-09-13
Issue 151 member renews a tools

### LINKED ISSUE #151: Member renews A tools
As a member, I want to be able to renew an "A-tool" item that is due via a button, so that I don't need to send an e-mail to do so and get instantaneous feedback.

This will also greatly relieve administrative burden, possibly as many as 50% of renewals!

For a member, put a button next to checked-out items in member history that says "renew" for when the item is an A tool.

For A tools, it should appear when 
- the item is checked-out,
- the person's membership is still valid, and
- within the term of the check-out time for the item (should be 7 days) (check with Jim on this)
- (UPDATED) item has been renewed < than max number defined by borrow policy

Ideally this logic is NOT hard-coded, since it will change based on future tenants, but that's probably a stretch for now.

After submitted, button should change to "renewed" and not be pressable (or similar functionality), and item should be renewed for the next due date. Members should receive a new e-mail receipt that reflects their renewal -- RK: I *think* this functionality exists in the admin side already, but if not, let me know and we can figure out what do next.

See admin Member page view for similar renewal functionality, e.g., https://ctl-staging.herokuapp.com/admin/members/399.

See #110 for more context.

RK note: B & C tools will need to be approved for renewal, so functionality is more complex. Still working that out in #123 .

dmbertoglio:
Allows members to renew loans. A `member_renewable` flag was added to the borrow_policy that enables members to renew loans for items with those borrow_policies. Members can renew their loans from the member loans page. 

Fixes #151

Co-authored by: @britth


jim:
Thanks again for this PR! I'm working through all the code from the past two weeks that needs to be merged and should get to this one shortly.

jim (approved):
This looks good. I merged in the latest code from `development` and will merge this PR once all the tests pass.

Thanks again!

PR #179, opened 2020-09-16
Docker setup

### LINKED ISSUE #92: Add Docker configuration for development
It would be nice to have a Dockerfile for developers to use for development. We'd also need a `docker-compose.yml` file that would start up Postgres, etc.

There is some preliminary work in the [docker-development](https://github.com/chicago-tool-library/circulate/tree/docker-development) branch that might be a good jumping-off point.

littleforest:
closes #92 

* Add Docker setup for development environment
* Add documentation for using Docker

What version of Postgres is being used in production? I'd like to update the postgres database service to use the same version.


jim:
@littleforest This is great! We're running 12.4 in production right now.

jim (approved):

jim:
@littleforest Rad! Thanks for setting this up for us.

PR #214, opened 2020-09-21
Displays the members items that are on hold

gregblake:
Closes https://github.com/rubyforgood/circulate/issues/205

#### When an item is placed on hold, but there is at least one active hold on that item, the item is on a wait list:

Screen Shot 2020-09-22 at 1 17 12 PM

#### When an item is placed on hold, and there are no active holds on that item, then the member can schedule a pickup any time within 7 days after the hold was placed:

Screen Shot 2020-09-22 at 2 10 24 PM

#### When the member attempts to remove a hold, there's a confirm dialog:

Screen Shot 2020-09-22 at 1 17 22 PM

### DIFF
--- a/app/controllers/members_controller.rb
+++ b/app/controllers/members_controller.rb
@@ -10,4 +10,9 @@ class MembersController < ApplicationController
     @loans = current_member.loans.includes(:item).order(:due_at)
     @holds = current_member.holds.includes(:item)
   end
+
+  def delete_hold
+    current_member.holds.find(params[:id]).destroy!
+    redirect_to member_loans_path
+  end
 end
--- a/app/controllers/members_controller.rb
+++ b/app/controllers/members_controller.rb
@@ -8,5 +8,6 @@ class MembersController < ApplicationController
 
   def loans
     @loans = current_member.loans.includes(:item).order(:due_at)
+    @holds = current_member.holds.includes(:item)
   end
 end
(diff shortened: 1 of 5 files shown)


app/views/members/loans.html.erb
```diff
@@ -36,3 +36,25 @@
...
+    </tr>
+  </thead>
+
+  <tbody>
+    <% @holds.each do |hold| %>
+      <tr>
+        <td><%= hold.item.name %></td>
+        <td><%= hold.item.number %></td>
+        <td><%= hold.item.status&.titleize %></td>
+        <td><%= link_to "Remove Hold", admin_member_holds_path(hold.id), method: :delete, data: { confirm: 'Are you sure you want to remove this hold?' } %>
```
gregblake, line 56: This currently uses the `admin_member_holds_path`, but we need to use a different path.

gregblake:
We may also need to add the ability for the member to see what position they're in to borrow that particular tool.

app/controllers/members_controller.rb
```diff
@@ -8,5 +8,11 @@ def history
 
   def loans
     @loans = current_member.loans.includes(:item).order(:due_at)
+    @holds = current_member.holds.includes(:item)
+  end
+
+  def delete_hold
```
jim, line 14: Eventually we might want to move this to a `Member::HoldsController#destroy` action, but until we get there, this works.

jim (approved):

app/controllers/members_controller.rb
rafbgarcia: Agreed, I almost did it but it'd conflict with the other team's work.

kskilt:
Tested with Steve B and looks good to us.

PR #217, opened 2020-09-22
Login Page Refresh

bh247484:
#### What does this PR do?
Refreshes the login and the forgotten password pages.  Only a modest redesign.  Didn't want to depart too dramatically from the existing style of the site.  Mostly spacing and alignment fixes.  The page is responsive and scales well on mobile.  The buttons have been restyled to align with the design systems color patterns.

PR also creates a sass index and partials folder for style scoping and clean per-page imports.

#### Things to Test, Lingering Concerns
I had to generate the devise views and change some relative paths.  Definitely worth double checking my math there. I only manually overrode `sessions/new` and `passwords/new`.

Finally there is only one Devise model, User.  It is used for both the Member Login and the Admin Login route.  Per Richard's recommendation I generalized the login title to accommodate both use cases.

#### Screenshots
login

forgotten

#### Mobile
mobile


config/initializers/devise.rb
```diff
@@ -232,7 +232,7 @@
   # Turn scoped views on. Before rendering "sessions/new", it will first check for
   # "users/sessions/new". It's turned off by default because it's slower if you
   # are using only default views.
-  # config.scoped_views = false
+  # config.scoped_views = true
```
jim, line 235: Does this line need to be uncommented?

jim (approved):
This looks good. The organization of the newly added styles is a nice improvement over putting everything in one giant file 😄

config/initializers/devise.rb
bh247484: Good catch!  Can actually revert this to false and leave it commented out.  This is config for using per model file paths for devise.  They're all pulling from the same views so I can revert this.

PR #222, opened 2020-09-22
docker dev improvements

benlangfeld:
- Improved flow of setup instructions for Docker
- Docker: Run the webpack dev server by default
- Allow dropping the DB in Docker

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -54,14 +54,6 @@ running `docker-compose build web` first.  If you have been making gem updates
 to your container without rebuilding the image, then the one-off container will
 be out of date.
 
-### Running webpack dev server
-To speed compiling of assets, run the webpack dev server in a separate terminal
-window:
-
-```
-$ docker-compose exec web bin/webpack-dev-server
-```
-
 ### Viewing logs
 To view the logs, run:
 ```
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -11,14 +11,13 @@ Dockerfile, or if gems or npm packages have been added or updated.
    you don't have permission to commit directly to this repo.
 4. Change into the application directory: `cd circulate`
 5. Run `docker-compose build` to build images for all services.
-6. Run `docker-compose up -d database` to start the database service.
-7. Run `docker-compose run --rm web rails db:reset` to create the dev and test
+6. Run `docker-compose run --rm web rails db:reset` to create the dev and test
    databases, load the schema, and run the seeds file.
-8. Run `docker-compose up -d` to start all the remaining services.
-9. Run `docker-compose ps` to view status of the containers. All should have
+7. Run `docker-compose up -d` to start all the remaining services.
+8. Run `docker-compose ps` to view status of the containers. All should have
    state "Up". Check the [logs](#viewing-logs) if there are any containers that
    did not start.
-10. The web application will be available at http://localhost:3000
+9. The web application will be available at http://localhost:3000
 
 ## For ongoing development:
 1. Run `docker-compose up -d` to start all services.
@@ -40,7 +39,7 @@ Dockerfile, or if gems or npm packages have been added or updated.
 ### Running commands
 In order to run rake tasks, rails generators, bundle commands, etc., they need to be run inside the container:
 ```
-$ docker-compose exec web rails db:migrate
+$ docker-compose run web rails db:migrate
 ```
 
 If you do not have the web container running, you can run a command in a one-off container:
@@ -73,15 +72,15 @@ $ docker-compose exec database psql -h database -Upostgres circulate_development
 
 #### Rails console
 ```
-$ docker-compose exec web rails c
+$ docker-compose run web rails c
 ```
 
 ## Testing Suite
 Run the testing suite from within the container:
 
 ```
-$ docker-compose exec web rails test
-$ docker-compose exec web rails test:system
+$ docker-compose run web rails test
+$ docker-compose run web rails test:system
 ```
 
 System tests will generate a screenshot upon failure. The screenshots can be
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,6 +13,8 @@ services:
       - config/docker.env
     stdin_open: true
     tty: true
+    depends_on:
+      - database
 
   assets:
     build: .
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,17 @@ services:
     stdin_open: true
     tty: true
 
+  assets:
+    build: .
+    command: bin/webpack-dev-server
+    volumes:
+      - .:/usr/src/app
+      - /usr/src/app/node_modules
+    env_file:
+      - config/docker.env
+    stdin_open: true
+    tty: true
+
   database:
     image: postgres:12.4
     env_file:
--- a/README.md
+++ b/README.md
@@ -9,6 +9,7 @@
 - [Requirements](#requirements)
 - [Integrations](#integrations)
 - [Development](#development)
+  * [Fetching the source code](#fetching-the-source-code)
   * [Setting up Circulate on your machine](#setting-up-circulate-on-your-machine)
   * [Running tests](#running-tests)
   * [Setup pre-commit checks](#setup-pre-commit-checks)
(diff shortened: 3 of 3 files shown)


DOCKER.md
```diff
@@ -81,15 +72,15 @@ $ docker-compose exec database psql -h database -Upostgres circulate_development
 
 #### Rails console
 ```
-$ docker-compose exec web rails c
+$ docker-compose run web rails c
```
robbkidd, line 75: I suspect that changing these from `exec` to `run` will result in conflicts when trying to run two or more things based on the `web` service at the same time. I'll give it a try.

robbkidd: This console command is OK with `web` already running, so I'll resolve this comment. But see later for the tests ...

DOCKER.md
```diff
@@ -81,15 +72,15 @@ $ docker-compose exec database psql -h database -Upostgres circulate_development
...
 ```
 
 ## Testing Suite
 Run the testing suite from within the container:
 
 ```
-$ docker-compose exec web rails test
-$ docker-compose exec web rails test:system
+$ docker-compose run web rails test
+$ docker-compose run web rails test:system
```
robbkidd, line 83: Changing the system tests command from `exec` (do it in same container) to `run` (do it in a new container) will result in conflicts when trying to run at the same time as the development `web` service.

```
Errno::EADDRINUSE: Address in use - bind(2) for "0.0.0.0" port 4000
```

Related to `exec` vs `run` also, `docker-compose run web` without a `--rm` option will leave behind a bunch of stopped containers with random names.

```
» docker-compose ps -a
             Name                           Command               State                        Ports
-------------------------------------------------------------------------------------------------------------------------
circulate_assets_1               ./docker-entrypoint.sh bin ...   Up       3000/tcp
circulate_database_1             docker-entrypoint.sh postgres    Up       0.0.0.0:5435->5432/tcp
circulate_selenium_chrome_1      /opt/bin/entry_point.sh          Up       4444/tcp, 0.0.0.0:5900->5900/tcp
circulate_web_1                  ./docker-entrypoint.sh bin ...   Up       0.0.0.0:3000->3000/tcp, 0.0.0.0:4000->4000/tcp
circulate_web_run_11d00c616bc5   ./docker-entrypoint.sh rai ...   Exit 1
circulate_web_run_13fa5fa5e61e   ./docker-entrypoint.sh rai ...   Exit 1
circulate_web_run_1c450e235f6d   ./docker-entrypoint.sh rai ...   Exit 0
circulate_web_run_637eeb1cf223   ./docker-entrypoint.sh rai ...   Exit 1
circulate_web_run_a53439855ff1   ./docker-entrypoint.sh rails c   Exit 0
circulate_web_run_e59160a4e8fe   ./docker-entrypoint.sh rai ...   Exit 1
```

littleforest:
I've never used `run` without `--rm` -- what are the benefits?

docker-compose.yml
```diff
@@ -13,6 +13,19 @@ services:
...
+  assets:
+    build: .
+    command: bin/webpack-dev-server
+    volumes:
+      - .:/usr/src/app
+      - /usr/src/app/node_modules
+    env_file:
+      - config/docker.env
+    stdin_open: true
+    tty: true
```
littleforest, line 28: Doesn't this just end up running the `webpack-dev-server` in a separate container that is completely isolated from the web application (and therefore not doing anything)?

littleforest: I did try to set up the webpack dev server as a separate service by opening port 3035 and setting the `WEBPACKER_DEV_SERVER_HOST` variable on both the webpack and web services, but I couldn't get it to work. If you've had success with that, please give it a shot in getting it set up. But I think just running the webpack server in the same container as the web application is not a big deal.

benlangfeld: It shares the same mounted volume...

benlangfeld: Ah, but that's not the point because this is actually serving assets itself. Sorry, my misunderstanding.

littleforest: It's actually using an anonymous volume...

benlangfeld: > But I think just running the webpack server in the same container as the web application is not a big deal.

It wouldn't be if it were started automatically. When one has to stop the web container to run DB migrations, it's actually quite a drag to have to start everything up again in multiple terminal panes, which is what I was trying to address.

littleforest: If you use `exec`, you don't need to stop the terminal to run the DB migrations. You just can run one container the whole time, which is also helpful for not having to immediately rebuild your image if someone adds a new gem. 

What benefit do you see in using `run` vs `exec`?

robbkidd: If you'd like to keep webpack-dev-server in a separate `assets` container service but still start it automatically when the `web` container is run, you could add `assets` to the `web` service's `depends_on:` key.

```
depends_on:
  - database
  - assets
```

littleforest: @robbkidd I'm using an anonymous volume in order to be able to run `yarn install` in the build stage of the Dockerfile. So locally the `/node_modules` folder does not sync up to the container `node_modules`. Won't your suggestion still have the issue of the assets getting compiled nicely in a separate container, but the web container won't be able to see them at all?

robbkidd: 🤔 Probably.

One option could be to unanonymize (nymize?) a volume for `node_modules` and share it between the web and asset services.

```
  app: &app
    build: .
    image: rubyforgood/circulate
    volumes:
      - .:/usr/src/app
      - node_modules:/usr/src/app/node_modules
    env_file:
      - config/docker.env
    stdin_open: true
    tty: true
    depends_on:
      - database

  web:
    <<: *app
    ports:
      - '127.0.0.1:3000:3000'
      - '127.0.0.1:4000:4000'
    depends_on:
      - database
      - assets

  assets:
    <<: *app
    command: bin/webpack-dev-server

volumes:
  db_data:
  node_modules:
```

(Er, that has a bunch of my tinkering included.)

Like the anonymous volume, if the named volume is empty, the `node_modules` from the image build will be used. Once something changes in the containers so that `node_modules` is written to, both containers will use that new content.

benlangfeld: > If you use exec, you don't need to stop the terminal to run the DB migrations or db:reset. 

Unfortunately this is not true:

```
➭ docker-compose exec web rails db:reset
I, [2020-09-23T12:53:34.731258 #278]  INFO -- Mjml: Using mjml binary at /usr/src/app/node_modules/.bin/mjml
PG::ObjectInUse: ERROR:  database "circulate_development" is being accessed by other users
DETAIL:  There is 1 other session using the database.
Couldn't drop database 'circulate_development'
rails aborted!
ActiveRecord::StatementInvalid: PG::ObjectInUse: ERROR:  database "circulate_development" is being accessed by other users
DETAIL:  There is 1 other session using the database.
```

```
database_1         | 2020-09-23 12:53:41.100 UTC [66] ERROR:  database "circulate_development" is being accessed by other users
database_1         | 2020-09-23 12:53:41.100 UTC [66] DETAIL:  There is 1 other session using the database.
database_1         | 2020-09-23 12:53:41.100 UTC [66] STATEMENT:  DROP DATABASE IF EXISTS "circulate_development"
```

littleforest: Ah right, sorry. You can keep it up when you run db migrations, but not `db:reset`. Are you running `db:reset` every time you pull code in case there's been a change to the seeds file...? Why not just run the migrations (or `bin/update`)?

littleforest: @robbkidd @benlangfeld I tried to use the named volume approach suggested by @robbkidd:

```
services:
  web:
    build: .
    image: circulate/web
    volumes:
      - .:/usr/src/app
      - node_modules:/usr/src/app/node_modules
    env_file:
      - config/docker.env
    ports:
      - '3000:3000'
      - '4000:4000'
    depends_on:
      - database
      - assets
    stdin_open: true
    tty: true

  assets:
    image: circulate/web
    volumes:
      - .:/usr/src/app
      - node_modules:/usr/src/app/node_modules
    env_file:
      - config/docker.env
    command: bin/webpack-dev-server

...

volumes:
  db_data:
  node_modules:
```

which initially looks like it works, but in the logs, I see that the web app is still compiling the assets into the local `/node_modules` folder (which is not the same data as `/usr/src/app/node_modules`). I'm confused why that is happening. Any ideas?

littleforest: Actually, I don't think using a shared named volume will work unless we compile the assets after the container is created since currently they are getting compiled in the build stage of the Dockerfile. Using the anonymous volume gets around the local `node_modules` overriding the container `node_modules`, but I don't think using the shared name volume will work with the current Dockerfile configuration.

PR #223, opened 2020-09-22
Implement place on hold button functionality for items

ghaydarov:
This PR adds Place on hold button for members to add holds on inventory screen.

[Link](https://github.com/rubyforgood/circulate/issues/193) to the story

Not Logged in
Screen Shot 2020-09-24 at 11 01 41 AM

Place hold one uniquely numbered item
Screen Shot 2020-09-24 at 10 53 10 AM
Screen Shot 2020-09-24 at 10 53 47 AM
Placing hold on non-uniquely numbered item
Screen Shot 2020-09-24 at 10 55 43 AM
Screen Shot 2020-09-24 at 10 55 53 AM
Screen Shot 2020-09-24 at 10 56 04 AM


app/controllers/member_holds_controller.rb
```diff
@@ -0,0 +1,19 @@
+class MemberHoldsController < ApplicationController
+  def create
+    if new_hold.save
+      redirect_to item_path(item), success: "Hold placed."
+    else
+      redirect_to item_path(item), error: "Something went wrong!"
```
gregblake, line 6: If there is a validation error on the new `Hold`, will the validation error(s) be included on the error message that is rendered in the view?

app/controllers/member_holds_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+  private
+
+  def item
+    @item ||= Item.find(params[:item_id])
+  end
+
+  def new_hold
+    @new_hold ||= Hold.new(item: item, member: current_member, creator: current_user)
+  end
+end
```
gregblake, line 19: Not sure if it's required for the style guidelines of this project, but a few files on this PR are missing a a newline at the end.

```suggestion
end

```

jim: Good catch! In general we like to run things through [standardrb](https://github.com/rubyforgood/circulate#setup-pre-commit-checks) to deal with this kind of code formatting issue.

app/controllers/application_controller.rb
```diff
@@ -11,6 +11,10 @@ def current_member
     current_user.member
   end
 
+  helper_method def current_user?
```
jim, line 14: Devise has a [user_signed_in?](https://github.com/heartcombo/devise#controller-filters-and-helpers) helper that provides similar functionality.

app/controllers/items_controller.rb
```diff
@@ -17,9 +17,7 @@ def index
     @pagy, @items = pagy(item_scope)
   end
 
-  def show
-    @item = Item.listed_publicly.find(params[:id])
-  end
+  def show; end
```
jim, line 20: I think this change will disable displaying the item on the public inventory page.

app/controllers/member_holds_controller.rb
ghaydarov: It has only the item, member and creator associations which we are passing to it thru current user and item. And now there is a generic success and error messages for hold and they will be displayed once the button is clicked.

app/controllers/items_controller.rb
ghaydarov: We added it as a method [below]( https://github.com/rubyforgood/circulate/pull/223/files#diff-2eb2bf110a48bee6c79d3bbb377a549eR41) in that controller.

app/controllers/member_holds_controller.rb
terryfinn: @gregblake I don't think we would have a error we could show them that would be more specific. Really if an error happens, it it not something the did wrong. The only thing they can do is click a button.

app/models/borrow_policy.rb
```diff
@@ -27,6 +27,14 @@ def complete_name
     "#{name} (#{code})"
   end
 
+  def allow_multiple_holds_per_member?
+    code == "A"
```
jim, line 31: `uniquely_numbered` should be `false` for A tools, so this could be:

```suggestion
     !uniquely_numbered
```

terryfinn: Thanks. I see uniquely_numbered and was wondering if that was the right way to do this.

jim (approved):

PR #226, opened 2020-09-23
Set PARALLEL_WORKERS to 1 for Docker system tests

### LINKED ISSUE #225: Look into parallel issues with running system tests in Docker
`PARALLEL_WORKERS=1` in docker.env is a workaround for now.

https://rubyforgood.slack.com/archives/C019C2178BD/p1600786559027800

littleforest:
The parallel system tests are causing port binding issues. Set the parallel workers to 1 to avoid this. 

closes #225


jim (approved):
Thanks for the fix!

PR #230, opened 2020-09-23
Members can create a new appointment

gregblake:
Closes https://github.com/rubyforgood/circulate/issues/204.

##### Members can return items they previously checked out, and pick up new items:

Screen Shot 2020-09-25 at 10 39 10 AM

##### Available appointment times are based on the library's hours:

Screen Shot 2020-09-25 at 10 39 44 AM

### DIFF
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -1,7 +1,7 @@
 class AppointmentsController < ApplicationController
   def new
     @holds = Hold.active.where(member: current_user.member)
-    @loans = current_user.member.loans.includes(:item)
+    @loans = current_user.member.loans.includes(:item).checked_out
   end
 
   def create
--- a/app/controllers/appointments_controller.rb
+++ b/app/controllers/appointments_controller.rb
@@ -1,6 +1,7 @@
 class AppointmentsController < ApplicationController
   def new
     @holds = Hold.active.where(member: current_user.member)
+    @loans = current_user.member.loans.includes(:item)
   end
 
   def create
@@ -8,8 +9,8 @@ class AppointmentsController < ApplicationController
     @appointment = Appointment.new
     @appointment.comment = appointment_params[:comment]
     @appointment.member = member
-
     @appointment.holds << Hold.where(id: appointment_params[:hold_ids], member: member)
+    @appointment.loans << Loan.where(id: appointment_params[:loan_ids], member: member)
 
     if appointment_params[:time_range_string].present?
       appointment_times = appointment_params[:time_range_string].split("..")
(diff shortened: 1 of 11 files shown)


app/views/members/loans.html.erb
```diff
@@ -66,3 +66,8 @@
     <% end %>
   </tbody>
 </table>
+<div class="button">
+  <% if @holds.present? %>
+  <%= link_to 'Schedule a Pick Up', new_appointment_path %>
```
gregblake, line 71:
```suggestion
    <%= link_to 'Schedule a Pick Up', new_appointment_path %>
```

test/lib/open_days_test.rb
```diff
@@ -0,0 +1,72 @@
+require "test_helper"
+
+class ActivityNotifierTest < ActiveSupport::TestCase
```
jim, line 3: I think this should probably be:

```suggestion
class OpenDaysTest < ActiveSupport::TestCase
```

test/lib/open_days_test.rb
```diff
@@ -0,0 +1,72 @@
...
+  TIME_SLOTS = [
+    OpenStruct.new(day: "Thursday", from: 18, to: 20),
+    OpenStruct.new(day: "Saturday", from: 11, to: 13)
+  ].freeze
+
+  test "returns possible time slots" do
+    Timecop.freeze(DateTime.new(2020, 9, 21)) do
```
jim, line 10: The [built-in Rails travel_to helper](https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel) can probably be used here instead of adding Timecop to the app.

jim (approved):
I left a couple notes, but I think this is ready to merge once those are addressed.

test/lib/open_days_test.rb
```diff
@@ -0,0 +1,72 @@
...
+    assert_equal slots[5].to_s, Time.zone.local(2020, 10, 1, 19, 0, 0).to_s
+    assert_equal slots[6].to_s, Time.zone.local(2020, 10, 3, 11, 0, 0).to_s
+    assert_equal slots[7].to_s, Time.zone.local(2020, 10, 3, 12, 0, 0).to_s
+    travel_back
+  end
+
+  # This handles an edge case:
+  # We're asking for time slots for Thursday and Saturday,
+  # if today is Friday, it's going to parse first the date for the
+  # next Wednesday (which is only next week) and then the date for
```
rafbgarcia, line 29:
```suggestion
  # next Thursday (which is only next week) and then the date for
```

app/controllers/appointments_controller.rb
```diff
@@ -0,0 +1,33 @@
+class AppointmentsController < ApplicationController
+  def new
+    @holds = Hold.active.where(member: current_user.member)
+    @loans = current_user.member.loans.includes(:item).checked_out
```
rafbgarcia, line 4: Hey @jim, these are the loans to be returned, we just wanted to confirm that we're listing them correctly here by calling the `checked_out` scope, does that seem right?

jim: That looks right to me!

PR #231, opened 2020-09-23
Create table, model for appointment and appointment holds

ghaydarov:
This adds appointment and appointment holds tables and models with all the relationships.

We came up with the names for models and we can discuss it here if we want a different naming.

We currently have `appointment_holds` joint table to store multiple holds for an appointment. 

In conversation with Greg, we thought it would be ok for us to use this joint table for pickup and returns. We would need to check for `loan_id` on the hold and if nil, we assume it's a hold without loan yet, otherwise, we assume it's a loan and it's being returned.

All of these are open to discussion.

Story [link](https://github.com/rubyforgood/circulate/issues/227)

### DIFF
--- a/app/models/appointment.rb
+++ b/app/models/appointment.rb
@@ -1,0 +1,7 @@
+class Appointment < ApplicationRecord
+  has_many :appointment_holds
+  has_many :holds, through: :appointment_holds
+  belongs_to :member
+
+  validates :starts_at, :ends_at, presence: true
+end
--- a/app/models/appointment_hold.rb
+++ b/app/models/appointment_hold.rb
@@ -1,0 +1,4 @@
+class AppointmentHold < ApplicationRecord
+  belongs_to :appointment
+  belongs_to :hold
+end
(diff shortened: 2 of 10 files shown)


app/models/appointment.rb
```diff
@@ -0,0 +1,7 @@
+class Appointment < ApplicationRecord
+  has_many :appointment_holds
+  has_many :holds, through: :appointment_holds
+  belongs_to :member
+
+  validates :starts_at, :ends_at, presence: true
```
gregblake, line 6: I think we should add one more validation on the `starts_at` and `ends_at` date ranges, to make sure the `starts_at` is earlier, like this:

```
validate :ends_at_later_than_starts_at

private
  def ends_at_later_than_starts_at
    if ends_at < starts_at
      errors.add(:ends_at, "must be after the starts_at date")
    end
 end
```

gregblake (approved):

jim (approved):

PR #239, opened 2020-09-25
Fix puma config to work with debugging

terryfinn:
When workers are specified, Puma will run in a clustered mode. Worker
processes will be killed if they do not report back to the parent
process that manages the cluster. This makes using debugging tools very
frustrating and less than productive. When the worker processes is
killed the debugging session is broken, and the standard in is
closed.

To address this issue, only specify a worker configuration when the
environment is not development.


jim (approved):
Thanks for fixing this! Future developers thank you.

PR #240, opened 2020-09-25
Upcoming Appointments Page

gregblake:
Closes https://github.com/rubyforgood/circulate/issues/232

Screen Shot 2020-09-25 at 1 16 55 PM

#### "My Appointments" link added to the navigation menu:

Screen Shot 2020-09-25 at 1 54 12 PM


app/views/appointments/index.html.erb
```diff
@@ -0,0 +1,97 @@
...
+  <% @appointments.each_with_index do |appointment, index| %>
+    <section class="appointment">
+      <div class="appointment-date">
+        <h2 class="title_bold"><%= index + 1 %>.</h2>
+        <h2 class="title_bold">
+          <%= appointment.starts_at.strftime("%a") %>
+          •
+          <%= appointment.starts_at.strftime("%b %-d, %Y") %>
+          <br />
+          <small class="gray-dark"><%= "#{appointment.starts_at.strftime("%I:%M%P")} - #{appointment.ends_at.strftime("%I:%M%P")}" %></small>
```
gregblake, line 13: The appointment times are currently displaying based on the UTC time zone. They should be displayed in either the Central time zone--or more ideally--in the current user's time zone, to support mutli-tenancy.

What's the best way to change this so that the times are displayed correctly?

gregblake: Thanks @rafbgarcia!

PR #246, opened 2020-09-25
CSS fixes

evalouderback:
**This PR includes:**

- The renaming of the of the `$link-color` so it doesn't interfere with the old styling still in place

- Changing the opacity of `caption` and `caption_lg` from `60%` to `.5` so that it doesn't display weird in heroku

- The addition of `card` to the style guide

- Styling of Account Settings page

Screen Shot 2020-09-25 at 11 51 08 AM


jim (approved):
Looks good! Thanks for fixing that `opacity` issue. We'll have to research more what is going on there, but I'm glad that we could work around it for now.

Issue #253, opened 2020-10-01
Fix item link on My Loans so that it goes to the item page.


rkim00:
Currently link to Items on "My Loans" page (https://circulate-staging.herokuapp.com/) under "My Loans & Holds" doesn't go anywhere. Link this to the corresponding item page, e.g., https://circulate-staging.herokuapp.com/items/67

Rajdeepc:
@rkim00 There is no "My Loans & Holds" section in the homepage https://circulate-staging.herokuapp.com/
Please be more specific.

PR #254, opened 2020-10-01
Fix item link on loans page

britth:
Fix item link from `/members/loans` checkout table such that clicking
on it will take the member to that item's overview page.

Issue #253

`rails test`
208 runs, 624 assertions, 0 failures, 0 errors, 2 skips

`rails test test/system/member_checked_out_items_test.rb`
1 runs, 3 assertions, 0 failures, 0 errors, 0 skips
(getting a couple of errors for other system tests, but appear unrelated)

### DIFF
--- a/app/views/members/loans.html.erb
+++ b/app/views/members/loans.html.erb
@@ -27,7 +27,7 @@
             <% end %>
           <% end %>
         </td>
-        <td><%= link_to loan.item.name %></td>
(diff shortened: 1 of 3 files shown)


mdworken (approved):
Test error appears to be on other branches as well so approving.

jim (approved):

PR #257, opened 2020-10-02
Add power source field to items

### LINKED ISSUE #152: Add "power source" field to item.
We want to track what is powering items (if they are powered). To do this, we'd like to add a `power_source` column to the `items` table.

Items don't have to have a value (a blank value will be interpreted as that item not being powered). For those items that do have a value for `power_source`, it can be one of: solar, gas, air, electric (corded), or electric (battery).

This value should be editable for admins on the item edit form, and displayed on that item's page both in the admin view and the public inventory.

swanson:
Resolves #152

Add `power_source` to `Item`.

image
image
image
image

### DIFF
--- a/app/controllers/admin/items_controller.rb
+++ b/app/controllers/admin/items_controller.rb
@@ -108,7 +108,7 @@ module Admin
     def all_item_params
       params.require(:item).permit(
         :name, :other_names, :description, :size, :brand, :model, :serial, :number, :image, :status, :strength,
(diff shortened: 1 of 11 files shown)


swanson:
The build seems to be failing on unrelated tests. Please advise if I need to get those passing :)

mdworken (approved):
Looks great, thanks for the thoughtful job on this!

PR #261, opened 2020-10-03
fix admin fixture, use admin user, fix redirect

pineappledreams:
### Issue

Fixes https://github.com/rubyforgood/circulate/issues/218

### Summary

1. Implement `require_admin` check at `app/controllers/admin/base_controller.rb`
2. When failed, it will redirect to root_url (tentative)
3. `test/fixtures/users.yml` the admin user is now an actual admin
4. Make admin tests actually use the admin-user.

### DIFF
--- a/app/controllers/admin/base_controller.rb
+++ b/app/controllers/admin/base_controller.rb
@@ -5,6 +5,7 @@ module Admin
     include ActiveSupport::Testing::TimeHelpers
 
     before_action :authenticate_user!
+    before_action :require_admin
     around_action :override_time_in_development, if: -> { Rails.env.development? }
 
     layout "admin"
@@ -23,7 +24,7 @@ module Admin
 
     def require_admin
       unless current_user.admin?
-        redirect_to admin_items_url, warning: "You do not have access to that page."
+        redirect_to root_url, warning: "You do not have access to that page."
       end
     end
   end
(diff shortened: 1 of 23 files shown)


app/controllers/admin/base_controller.rb
```diff
@@ -23,7 +24,7 @@ def override_time_in_development(&block)
 
     def require_admin
       unless current_user.admin?
-        redirect_to admin_items_url, warning: "You do not have access to that page."
+        redirect_to root_url, warning: "You do not have access to that page."
```
jim, line 27: Good catch!

jim (changes requested):
Thanks for tackling this one!

We've been trying to move away from fixtures and toward using [FactoryBoty](https://github.com/thoughtbot/factory_bot)factories instead. This wasn't documented anywhere, so there is no way you could have known.

In any case, if you can update the calls to `users(:admin)` to `create(:admin_user)`, we should be able to merge this in.

jim:
Also, the test failure on this branch has been fixed on the `development` branch, so you can either merge that into this one or we can take care of it when it comes time to merge.

pineappledreams:
@jim gotcha bud, I will get right to it.

pineappledreams:
That's strange, I am running the `docker-compose exec web rails test:system` and `rails test` too on my computer but it won't get any errors. Am I missing something here?

jim:
@pineappledreams Everything is passing locally for me as well. It might be a CI issue, as we've been having some build instability on Circle recently. I'm going to merge this.

jim (approved):
Thanks for making those revisions!

PR #267, opened 2020-10-04
219 unify login

pineappledreams:
### Issue
Fixes https://github.com/rubyforgood/circulate/issues/219

### Summary
- Remove admin login button since it leads to the same page anyway
- Add a check when user is logged in, it won't show the button to register.


jim (approved):
Looks great. Thanks!

PR #269, opened 2020-10-04
added POSTGRES_PASSWORD env variable

italomatos:
This PR allow you setup password in test and development environment in .env file.


jim (approved):
Thanks!

PR #279, opened 2020-10-05
Rearranged nav bar links

ariane-h:
Swapped around the nav bar links in the layout requested.


jim (approved):
Looks good! Thanks for the contribution.

PR #281, opened 2020-10-05
[Librarian] Show loans and holds stats on members index [#154]

### LINKED ISSUE #154: Librarian: Add number of active loans and holds to member index
The admin members index shows a list of all members and some of the information associated with their account.

Add two columns to this listing that shows how many items the member is currently borrowing and how many holds they have. The new columns can be added to the right side of the table.

If either of the values is 0, display a `-` for that value so that they page is easier to scan. Otherwise, show the value.

matisnape:
Resolves #154 

Current view:
image

### DIFF
--- a/app/javascript/stylesheets/application.scss
+++ b/app/javascript/stylesheets/application.scss
@@ -165,6 +165,7 @@ header.navbar {
   grid-row-gap: .5rem;
   grid-column-gap: .3rem;
   grid-template-columns: .3fr 1fr 1fr 1fr 1fr;
+  grid-template-columns: 0.3fr 1fr 1fr 1fr 1fr 0.5fr 0.5fr;
 
   .members-table-header {
     font-weight: bold;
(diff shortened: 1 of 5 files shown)


app/javascript/stylesheets/application.scss
```diff
@@ -165,6 +165,7 @@ header.navbar {
   grid-row-gap: .5rem;
   grid-column-gap: .3rem;
   grid-template-columns: .3fr 1fr 1fr 1fr 1fr;
```
jim, line 167: There are two `grid-template-columns` declarations. I think we only need the second (newer) one.

jim (changes requested):
This is really going to help staff see what is going on at a high level! Thanks for doing this work.

There is a duplicate CSS declaration in one of the files. Once we fix that, this is ready to go.

app/javascript/stylesheets/application.scss
matisnape: yeah, my editor's formatter formatted the whole file and I didn't want to commit all the changes, because I don't know if that formatting is OK. Apparently I selected too much for commit :)

matisnape: fixed

jim (approved):
Looks good!

app/javascript/stylesheets/application.scss
jim: Thank you!

PR #284, opened 2020-10-05
Update errors -> error in calendaring module

britth:
In https://github.com/rubyforgood/circulate/commit/826da0b35259a506811a67616e4b7540332296ee, it looks like `Result#errors` was renamed `Result#error`. One 
spot that appears to have been missed is in `calendaring.rb`. Locally, I was
getting an error `Undefined method errors for <result...>` when trying
to access the volunteer shifts report. Updating to `error` allowed the
page to load and revealed a permissions error. This PR updates these
remaining calls to `.errors`. Potentially it might make sense to further
update the error text that is flashed, but I did not make that update as
yet.

before
Screen Shot 2020-10-05 at 6 42 49 PM

after
Screen Shot 2020-10-05 at 7 48 52 PM

(I also updated two unrelated system tests in this commit, as without
the `wait` they occasionally fail for me locally. Can back that out if 
desired though!)

unit tests:
208 runs, 624 assertions, 0 failures, 0 errors, 2 skips

system tests:
58 runs, 278 assertions, 0 failures, 0 errors, 4 skips

### DIFF
--- a/app/controllers/concerns/calendaring.rb
+++ b/app/controllers/concerns/calendaring.rb
@@ -9,7 +9,7 @@ module Calendaring
       @events = result.value
     else
       @events = []
-      flash.now[:error] = result.errors
+      flash.now[:error] = result.error
     end
   end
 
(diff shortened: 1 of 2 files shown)


jim (approved):
Good catch. I'll pull in the system test changes, too, since I've seen those ones time out as wel.

PR #285, opened 2020-10-06
Add columns to loan summary index view for item number and categories…

### LINKED ISSUE #148: add item # to displays of loans in admin screens
As an librarian/admin, I want to be able to see item #s next to loans, so that I can see more details on the specific loan, and get a sense of what categories of tools are loaned out.

For all three views (checked out, overdue, returned), add two columns, one for category (by letter) and one for item # next to loans.

loans pages annotated

cjilbert504:
…. fixes #148
Added columns for item number and categories in loan summary view.

There may be more work to be done around this issue but I wanted to submit this PR and get feedback as to what other work may need to be done. 

image


rkim00:
Hey Collin. Thanks for this!

Just looking at your screen shot, the Category is A, B, or C, the Number is just the number. So you can split the Number into one column and the Category into another.

It's probably a little more intuitive to have the Category first, then the Number, left-to-right.

jim:
@rkim00 @cjilbert504 At CTL we generally include the Borrow Policy number when speaking about the tools, so we'll say `A-143` or `B-1442`. We combine the letter and number almost everywhere, so treating it differently on this page might be a little strange for users. What appears in the screenshot above looks good to me in terms of these bits of information.

An item's "category" would be something like `Hand Tools` or `Drills and Drivers`, but it sounds like the terminology might not be consistent everywhere. I _think_ what @rkim00 is envisioning is the letter and number from the item's identifier and not its categories, but it's probably best if he clarifies what he was looking for here 😄

cjilbert504:
Ok no problem. Once there is some more detail around this I can adjust and resubmit a new PR addressing this.

rkim00:
So by "category" I meant "borrow policy," yes.

I checked my notes on this; this is something I think I discussed with Tessa. The general idea was that by having the borrow policy and the numbers, you could (1) see generally what groups ("borrow policy") of tools were being checked-out more often, (2) eventually be able to sort by borrow policy, e.g., to see all the "A" tools together, and (3) eventually be able to sort by tool number, to find a specific tool number more easily and to see all of the associated loans with a tool more easily. Obviously we're not implementing sorting here, but separating out the columns was the first step.

Does that make sense? If @jim thinks that would be confusing without any obvious benefit, however, then I think it makes sense to use the combined number—that still could eventually allow sorting by borrow policy, since it will start with A, B, or C. Ultimately I'm not too attached—seems like something we can see if people like and can always tweak later.

cjilbert504:
So just to throw this out there, if the end vote is to go with using the Borrow Policy number, would it be better to just make one new column called "Borrow Policy Number" or something? And then, as you stated @rkim00, if it turns out that down the road it needs to be further split out into other columns then that could be done at that time.

rkim00:
> So just to throw this out there, if the end vote is to go with using the Borrow Policy number, would it be better to just make one new column called "Borrow Policy Number" or something? And then, as you stated @rkim00, if it turns out that down the road it needs to be further split out into other columns then that could be done at that time.

Yes, that's right. I'd label the column "Inventory Number" or just "Number". If you've got cycles to do that now before Jim has a chance to respond, then I'd just do that—that still delivers improved user value. We can likely get someone to split out towards the end of the month if that seems more valuable.

cjilbert504:
This PR has been updated, here's the updated screen shot:
image

Also there is a failing test that I have locally but it is unrelated to the changes in regard to the code changes in this PR as it was failing from the start before I started working on this issue. This is the failure I have locally:
`MembershipTest#test_creates_a_membership_for_a_member`

Let me know if anything else needs to change. 😁😁😁

mdworken (changes requested):
Looks great! If we can just remove the package_lock.json should be ready to merge!

cjilbert504:
I removed the package lock file. Hopefully correctly 😂 Let me know if there's anything else or different I need to do.

PR #290, opened 2020-10-07
added message when no have items on hold and loan

italomatos:
This PR have a Goal solve the problem in Member Hold list.
When member has no item on hold, show message: 'You have no items on hold'
The behavior was replicated to Loan, in the same screen.

card: https://github.com/rubyforgood/circulate/issues/288

image

### DIFF
--- a/app/views/members/loans.html.erb
+++ b/app/views/members/loans.html.erb
@@ -4,80 +4,87 @@
 
 <h1 class="new_title">My Loans & Holds</h1>
 
-<h3 class="title_bold">Checked-out Items</h3>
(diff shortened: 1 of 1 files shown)


rkim00:
Thanks! I wrote the wrong page in the title of the issue (sorry about that, now fixed), but you found the right page. And thanks for fixing the Loans as well; I hadn't checked if that needed a message as well, but suspected it might. Great attention to detail!

mdworken (approved):
This looks excellent to me! I'll line-by-line it in the morning when I have more energy, but this should really be good to go.

jim (approved):
This looks great.

PR #291, opened 2020-10-08
Update setup script to run yarn

tmr08c:
Running `bin/setup` on a new project was failing with the following
error:

```
error Couldn't find a binary named "mjml"
rails aborted!
Couldn't find the MJML 4. binary. Have you run $ yarn add mjml?
```

This was failing when attempting to create the database. I believe this
is related to bootsnap loading the environment and the `mjml-rails` app
expecting the `mjml` binary to exist.

By running `yarn` first the `mjml` npm package will be installed and the
binary will be available, enabling `bin/setup` to complete
successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,8 @@ After you have the application running, here are some places to explore:
 
 Use the standard Rails test commands: `rails test`, `rails test:system`, etc.
 
+Note, in order to get system tests to run, you will need `chromedriver` installed. See [Requirements section](#requirements) above.
+
 ### Setup pre-commit checks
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -129,6 +128,8 @@ After you have the application running, here are some places to explore:
 
 Use the standard Rails test commands: `rails test`, `rails test:system`, etc.
 
+Note, in order to get system tests to run, you will need `chromedriver` installed. See [Requirements section](#requirements) above.
+
```
tmr08c, line 132: I thought it may be useful to move the whole `requirements` header to be a subsection of the `development` header, but I didn't know if that would be too far-reaching. Let me know if you're open to seeing what that would look like.

mdworken (approved):
I really like this - bin/setup is our friend!

README.md
jim: That sounds good to me! If you want to make that change in this PR, I can wait to merge it.

jim (approved):
This is great. I'm on board wit moving the requirements section as you suggested.

PR #292, opened 2020-10-08
Add `webdrivers` gem to automatically install `chromedriver`

tmr08c:
# What 

This gem will automatically install webdrives needed to run Selenium
tests.

Since the test setup only uses `chromedriver`, we only `require` that in
our `application_system_test_case` setup.

# Why

When getting set up and creating #291, I ran into system test failures because
I didn't have chromedriver installed. I thought this gem could help make it easier 
for other developers to get set up to contribute more quickly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ There is content and information hard-coded in many of the views that is specifi
 
 Circulate is a fairly basic Rails application. The main application requires a recent version of Ruby, a PostgreSQL database, and a modern version of Node and Yarn to build assets.
 
-* A version of chromium (Google Chrome is fine) and a compatible `chromedriver` are required to run application tests.
+* A version of chromium (Google Chrome is fine) and a compatible `chromedriver` are required to run application tests. This will be downloaded automatically for you when running system tests.
 * Imagemagick needs to be installed for gift memberships and item thumbnails to be generated.
 
 ## Integrations
--- a/Gemfile
+++ b/Gemfile
@@ -82,6 +82,7 @@ group :test do
   # Adds support for Capybara system testing and selenium driver
   gem "capybara", ">= 2.15"
   gem "selenium-webdriver"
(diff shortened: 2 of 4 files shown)


mdworken (approved):
I seem to recall this not working fully in the past for linux users - happy to try it though!

README.md
```diff
@@ -44,7 +44,7 @@ There is content and information hard-coded in many of the views that is specifi
 
 Circulate is a fairly basic Rails application. The main application requires a recent version of Ruby, a PostgreSQL database, and a modern version of Node and Yarn to build assets.
 
-* A version of chromium (Google Chrome is fine) and a compatible `chromedriver` are required to run application tests.
+* A version of chromium (Google Chrome is fine) and a compatible `chromedriver` are required to run application tests. This will be downloaded automatically for you when running system tests.
```
tmr08c, line 47: > I seem to recall this not working fully in the past for linux users - happy to try it though!

Thank you for pointing this out! This is an example of the danger of only thinking about your own setup 🤦 

Fortunately, their [README](https://github.com/titusfortner/webdrivers#description) claims support for the following:

> Works on macOS, Linux, Windows, and Windows Subsystem for Linux (WSL) v1.

Hopefully, this doesn't mess up anyone's workflow 🤞 

This following suggestion updates the README to add a bit more caveats and information. Maybe that can help if this does fail for some users? 

```suggestion
* A version of chromium (Google Chrome is fine) and a compatible `chromedriver` are required to run application tests. Thanks to the [`webdrivers` gem](https://github.com/titusfortner/webdrivers), this should be downloaded automatically for you when running system tests. If you into issues, you may want to check their README or manually install `chromedriver`.
```

jim: I also feel like I had an issue once, but it's been a while, so I agree: let's give it a shot.

jim (approved):

PR #297, opened 2020-10-10
Adds Renew All Button

### LINKED ISSUE #156: Librarian: Add "renew all" button to admin view member page that renews all borrowed items.
Often a librarian renews all of the items for a member. We'd like to add a "Renew All" button to the admin member page so that staff can click one button to accomplish this instead of having to click renew on each item that is currently checked out. We've had some issues where one item gets missed, and hopefully having an easy way to renew all items will make this less common.

Renewing items in this way should have the same effect as if a staff member had clicked "Renew" manually on each item: a single item renewal should be able to be undone, etc.

This button can be places above the list of checked out items on the "Current Loans" tab of the admin member page. If you have an idea for a better place to put it, please go ahead and do so!

chaserx:
Closes #156 

This PR is a good first iteration on bulk loan renewal in the admin panel.
It also includes some updates to seed data for ease of testing this functionality locally. 

Google Chrome -  Circulate 2020-10-09 20_01_22

### DIFF
--- a/app/controllers/admin/bulk_renewals_controller.rb
+++ b/app/controllers/admin/bulk_renewals_controller.rb
@@ -1,0 +1,21 @@
+module Admin
+  class BulkRenewalsController < BaseController
+    include ActionView::RecordIdentifier
+
+    before_action :set_member, only: [:update]
+
+    def update
+      @member.loans.checked_out.each do |loan|
+        loan.renew! if loan.renewable?
+      end
+
+      redirect_to admin_member_url(@member.id)
+    end
+
+    private
+
+    def set_member
+      @member = Member.find(params[:id])
+    end
+  end
+end
--- a/test/controllers/admin/bulk_renewals_controller_test.rb
+++ b/test/controllers/admin/bulk_renewals_controller_test.rb
@@ -1,0 +1,32 @@
+require "test_helper"
+
+module Admin
+  class BulkRenewalsControllerTest < ActionDispatch::IntegrationTest
+    include Devise::Test::IntegrationHelpers
+
+    setup do
+      @admin = create(:admin_user)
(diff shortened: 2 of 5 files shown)


app/controllers/admin/bulk_renewals_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Admin
+  class BulkRenewalsController < BaseController
+    include ActionView::RecordIdentifier
+
+    before_action :set_member, only: [:update]
+
+    def update
+      @member.loans.checked_out.each do |loan|
+        loan.renew! if loan.renewable?
+      end
```
chaserx, line 10: This probably isn't optimal if there is the case where a member has a very large number of items out on loan. However, I think this is a decent first iteration.   

Other thoughts include moving the singular or bulk renewal process out to a service object to avoid having that logic here.

jim: Good call-out. The most we ever have to renew at once is ~12 items, so I think this is OK to start with.

test/controllers/admin/bulk_renewals_controller_test.rb
```diff
@@ -0,0 +1,32 @@
...
+      @admin = create(:admin_user)
+      sign_in @admin
+    end
+
+    test "should renew all renewable loans" do
+      @member = create(:verified_member)
+      @loans = create_list(:loan, 5, member: @member)
+
+      # NOTE(chaserx): ensure that at least on of the items on loan is not renewable.
+      @loans[2].item.borrow_policy.update(renewal_limit: 0)
```
jim, line 17: Nice!

jim (approved):
This looks good to me! Thanks for the contribution.

PR #299, opened 2020-10-11
update category nav helper and styles to make labels more readable

metamoni:
This PR only updates styling to make labels more readable. It doesn't deal with the suggested dropdown, as the details of what that would look like were a bit hazy.

To look at implementing a dropdown menu a bit of design is needed, I think. At the moment, categories can allow any number of nested sub-categories. For that reason, we'd need to consider where we want the caret/arrow icon to be: only on the first parent or on all nested categories that in their turn have subcategories?

For the moment, I think a simple implementation like the one in this PR is cleaner and more readable.

### DIFF
--- a/app/helpers/items_helper.rb
+++ b/app/helpers/items_helper.rb
@@ -36,10 +36,11 @@ module ItemsHelper
   def category_nav(categories, current_category = nil)
     return unless categories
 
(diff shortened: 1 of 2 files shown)


jim:
This is a solid improvement. Thank you.

rkim00:
Great, looks much better!

Issue #207, opened 2020-09-21
Admin can View / Edit an appointment & check tools out


sboligitz:
As an CTL Admin, I want to be able to see the detailed view of an appointment, as well as check tools out, so that I can "initiate" the borrowing process.

- [x] Should be able to check out tools individually
- [x] Should be able to check out all tools at once
- Should be able to remove items from appointment
--- UPDATE: I think we want to deviate slightly from the mock ups here.  Let's have two buttons: 
1. [x] One to remove the item from that appointment but to leave it on Hold, and 
2. [ ] a second to remove the item from the appointment AND from the Hold list.

- [x] Should be able to add items to appointment (if not on Hold)

- [x] Once an item has "Checked Out" then this should cause that item to be displayed in the "My Loans & My Holds" page under the "My Loans" section

https://drive.google.com/file/d/1fK-AI8Cf5xe6-Bf5nIE1DiZ_UWDw6gQK/view

PR #300, opened 2020-10-11
Allow admins to cancel holds from an appointment

fosterv2:
This PR attempts to fix one of the bullets on #207

In an effort to allow admins to cancel the hold for an item at the same
time as they remove an item from an appointment, a button was added to
the admin appointment show page.

We opted to add another parameter to the existing
remove_appointment_hold controller action and base the canceling of a
hold on that. We chose to do this instead of making a new action to
perform both destroys.

* Turned the "remove item" link into a button.
* Made the "Cancel Hold" action a button.
* Moved these buttons to appear below the existing "Check-out" button

Possible To-Dos

* [ ] Add tool-tips to the new buttons to explain what they each do
* [ ] Fix the button styling/spacing
* [ ] Considering security: are there guards in place to make sure only permitted people can perform this action
* [ ] Add test/s?


jim:
I think it would be nice to add a test for this behavior. A system test would be ideal to capture the entire hold flow (request and then cancel), but a simpler controller test would also suffice. Please LMK if you want some help with sketching out what that would look like!

jim (approved):
I tested this and it works as intended. I'd still like to get a test in for this, but that can be a follow-up bit of work since this feature is pretty important for librarians.

Thanks for your work on this!

PR #302, opened 2020-10-12
add nodejs and yarn to asdf tool-versions

robbkidd:
If we're including `.tool-versions` for [asdf](https://asdf-vm.com/#/) users, we ought to declare nodejs and yarn as necessary, too.

The selection of which version of these tools to include here derive from the versions of [nodejs](https://pkgs.alpinelinux.org/packages?name=nodejs&branch=v3.12) and [yarn](https://pkgs.alpinelinux.org/packages?name=yarn&branch=v3.12) currently packaged in the ruby:2.7.1-alpine docker image.


jim:
🚀

Issue #283, opened 2020-10-05
Admin Appointment Page - Add Item for check-in


rkim00:
As an librarian I should be able to add an item to a scheduled appointment for check in.

On pages like https://circulate-staging.herokuapp.com/admin/appointments/7 , should be able to add an item to the Scheduled Appointment for eventual check in. Note that the item should already be checked-out (be part of a loan) by the member who's associated with the appointment.

Rough mock-up here:
https://drive.google.com/file/d/1fK-AI8Cf5xe6-Bf5nIE1DiZ_UWDw6gQK/view

UX implementation could happen a number of ways. Easiest is probably just to have a pulldown of all of the member's loans that are not already associated with the appointment, and have a button next to it that says "Add item for return." 
Selecting an item and clicking the button adds the item to the list of Items to be checked-in above the pulldown and button, and updates the contents of the pull-down.

PR #303, opened 2020-10-12
Add ability to add items for check-in to admin appointment page

fosterv2:
Solves #283 

In an effort to allow admins to add a currently loaned item to an existing appointment, a dropdown form with items currently loaned to the user was added to the admin appointment show page. A helper method to find the items loaned to the user was added to the admin appointments controller. Added a route for creating an appointment loan, and made associated controller actions.

Possible To-Do: Add test/s?


jim:
This looks good! Let's add a test to cover this functionality.

fosterv2:
I'm pretty new to open source, and I haven't written many tests before, but I can give it a try. I'm assuming that I would only add system and/or controller tests for this functionality, but I don't see any obvious files for these tests to go in so I think I would need to make new test files. Would the new test file be for appointments in general, or for this functionality specifically?

app/views/admin/appointments/_checkins.html.erb
```diff
@@ -31,5 +31,23 @@
...
   <% end %>
+
+  <div class="columns">
+    <p>Find and add another item to the appointment for check-in.</p>
+    <%= form_with(model: [:admin, AppointmentLoan.new], url: admin_appointment_loans_path(current_appointment), method: :post, builder: SpectreFormBuilder) do |form| %>
+      <%= form.text_field :item_id, list: "items2", autocomplete: :off %>
+      <datalist id="items2">
+        <%= items_avalable_to_add_to_dropoff.find_each do |loan| %>
+          <% if !appointment_return_items.include?(loan) %>
+            <%= tag(:option, value: loan.item.complete_number_and_name) %>
```
jim, line 42: I think we probably want to send the `item_id` to indicate which item instead of the name and number
```suggestion
            <%= tag(:option, value: loan.item_id) %>
```

app/controllers/admin/appointment_loans_controller.rb
```diff
@@ -11,6 +16,37 @@ def remove_appointment_loan
...
+        loan: loan_for_item_to_add
+      )
+    end
+
+    def loan_for_item_to_add
+      @item_to_add_loan = appointment.member.loans.find_or_initialize_by(item: item_to_add)
+    end
+
+    def item_to_add
+      Item.find_by_complete_number(params.require(:appointment_loan)[:item_id].to_s)
```
jim, line 47: With the change suggested below, this can become:

```suggestion
      Item.find(params.require(:appointment_loan)[:item_id].to_s)
```

jim (changes requested):
Sorry about not getting back to you sooner!

I think the only blocking change I'd like to see is changing the data that we send to the server to identify the item to add to the `AppointmentLoan`.

In terms of tests, you're right that there aren't any system tests covering this. It would be a pretty big lift to add those for this small change. A good exercise for sure, but a lot of work for the scope of the change in this PR.

A controller test would probably be a better place to start. I think [this one](https://github.com/rubyforgood/circulate/blob/development/test/controllers/admin/gift_memberships_controller_test.rb#L23-L40) would be a good controller test to use as a template.

If you'd like to pair on writing a test or two, ping me in the Ruby For Good Slack (I'm `Jim Benton (he/him) - Circulate`) and we can find a time 😄 .

jim:
@fosterv2 I'm going to make the change I suggested above and merge this in so we can unblock users. If you'd like to keep chugging on this path and tackle writing some tests, feel free to open another PR for that work. I'm available to pair or chat async about where to proceed if that would be helpful.

jim (approved):

jim:
I couldn't push to the branch on this PR, but the changes I made can be seen in 53063f7f.

fosterv2:
@jim Thanks for all the advise!

Issue #311, opened 2020-10-13
Librarian Appointments page: Adjust display of date at top of page so that it doesn't wrap unnecessarily.


rkim00:
On pages like https://circulate-staging.herokuapp.com/admin/appointments?day=2020-10-15 , the date at the top is wrapping. Adjust HTML/CSS so that it doesn't wrap unnecessarily.

avinashupadhya99:
I can pick this up.

Issue #312, opened 2020-10-13
Librarian Appointments: Update text for when no appointments.


rkim00:
On pages where there are no appointments (e.g., https://circulate-staging.herokuapp.com/admin/appointments?day=2020-10-13 ), mainwell message is

"No matching appointsments for this day!"

Change this error message to be a little more simple and fix typo ("appointsments"):

"No appointments scheduled for this day."
[Note that we want to remove "!"]

avinashupadhya99:
This should be fairly easy, can I pick this up?

rkim00:
Thanks for being so active on our repo! I've sent you an invite to be a contributor; the invite should be in your e-mail. If you accept, you can assign issues yourself, rather than waiting to be assigned.

avinashupadhya99:
Thanks. Glad I get to help.

PR #313, opened 2020-10-13
Fixed message in appointments page and adjusted display of date. Fixes Issue #311 and #312 

avinashupadhya99:


app/views/admin/appointments/index.html.erb
```diff
@@ -5,16 +5,16 @@
 <div class="columns">
   <div class="col-12" style="justify-content: center;">
     <ul class="pagination">
-      <li class="page-item page-prev">
```
jim, line 8: I see we've already got some inline `style` tags in this template, but can you please move the new styles into a `scss` file?

jim (changes requested):
Thanks for the contribution! I'd like to try to keep the styles in SCSS files, although I see we've already had some leak out into this template.

app/views/admin/appointments/index.html.erb
avinashupadhya99: Sure, I will do that

avinashupadhya99: It is done. @jim

jim (approved):
Thanks for moving those styles!

PR #321, opened 2020-10-14
Showing member's place in line on the members hold  screen in the admin member holds view

preethinarayan:
Addresses the issue https://github.com/rubyforgood/circulate/issues/75.

@jim I'm not sure if there is a better way to do this. Open to suggestions on code improvement.

Screen Shot 2020-10-13 at 10 04 31 PM

### DIFF
--- a/app/controllers/admin/members/holds_controller.rb
+++ b/app/controllers/admin/members/holds_controller.rb
@@ -32,6 +32,10 @@ module Admin
       def new_hold_params
         params.require(:hold).permit(:item_id)
       end
+
+      helper_method def place_in_line_for_hold(hold)
+        Item.find(hold.item.id).holds.index(hold) + 1
+     end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/controllers/admin/members/holds_controller.rb
```diff
@@ -32,6 +32,10 @@ def destroy
       def new_hold_params
         params.require(:hold).permit(:item_id)
       end
+
+      helper_method def place_in_line_for(hold)
+        Item.find(hold.item.id).holds.index(hold) + 1
```
jim, line 37: This will make a query for every item on hold and instantiate a `Hold` object for every hold on that object, but most people will have less than 10 items on hold at any one time.

The optimization here would be to pull in the place in line using SQL when loading the holds, but I think that can be left until later (and if this actually becomes a performance issue).

jim (approved):
I think this works! We could optimize determining the place in line using some custom SQL, but until we actually see a performance issue, I think it's fine to defer that.

app/controllers/admin/members/holds_controller.rb
preethinarayan: Makes sense. Thanks for the explanation 👍🏽

PR #317, opened 2020-10-13
Add links to member and item pages

carterrwood:
### Description

When viewing an appointment there are now links that render for members names and items that will take you to the librarian view.

### Issues Resolved 

Fixes #314

PR #325, opened 2020-10-14
Fix "no appointment" text

carterrwood:
### Description

I noticed when working on #317 the message that appears on the admin appointments index page had a misspelling, "appointsments". 

Added tests to check this text in the future.

### Notes

This is build off of #317 so I could use the test file I created.

### DIFF
--- a/app/views/admin/appointments/index.html.erb
+++ b/app/views/admin/appointments/index.html.erb
@@ -40,16 +40,16 @@
           <% @appointments.each do |appointment| %>
             <tr>
               <td><%= l appointment.starts_at, format: :hour %> - <%= l appointment.ends_at, format: :hour %></td>
-              <td><%= appointment.member.full_name %></td>
+              <td><%= link_to appointment.member.full_name, member_profile_path(appointment.member) %></td>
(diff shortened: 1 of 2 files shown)


carterrwood:
I didn't see that #313 already took care of this, oops! I'm all good with closing this PR but we could also keep it for the test I wrote (although if #313 is merged first I'll have to update that test, which is fine).

rkim00:
Thanks for the initiative—sorry that your work ended up being duplicative.

I am not a programmer and I've written very few tests in my life, so feel free to tell me if I'm missing something, but I'm not sure we want a test that checks the specific error message text. That would mean that the test will fail if we adjust the error message test, no? Regardless, note that #312 changes this error text, so this test would fail if merged.

carterrwood:
The test is less for checking the specific text that appears and more so testing to make sure that _when_ there are no appointments, that the application is behaving as expected. Either way it's not a very important test so I'll just close this PR. Thanks for the input!

rkim00:
Okay, cool. Again, thanks for noticing... better to have more people noticing and overlapping than issues not getting noticed and voiced!

PR #326, opened 2020-10-15
Fixed internal error when navigating to pages that require login. Issue#316

avinashupadhya99:
I have added the authentication to all controllers that require them as far I know. Please let me know if I missed any controllers that required authentication.


jim (approved):
Good catch! This was definitely missing.

Issue #167, opened 2020-09-13
Make member type in old password to edit new password.


rkim00:
As a member, I want to be required to enter my current/old password when I edit my password, so that someone can't maliciously change my password without my knowledge if I leave my computer open after logging in.

jim:
@amygurski Have you had a chance to work on this? We have other folks interested in giving it a shot if you won't have time in the near future. Cheers!

amygurski:
@jim thanks for the reminder. let me see if i can do it this afternoon. if not, i'll let you know and take my name off it.

PR #327, opened 2020-10-15
Add password confirmation when user edits password

amygurski:
When members edit their password, they are now required to enter their current password.

Adds current password field, adjusts field labels for clarity (e.g. default Password -> New Password), and updates the password controller tests accordingly.

Linked to #167


jim (approved):
Looks great! Thank you for working on this.

Issue #216, opened 2020-09-22
Redirect logins to originating page


sboligitz:
As a CTL member not currently logged in, when clicking on "Member Login" or a login link from a page (e.g., an item page), I want to be redirected to the page I came from after a successful login, so that I can resume browsing where I left off. (Per @avinashupadhya99 's suggestion.)

RK note: see also #174 .

RK additional note: situations that we haven't fully worked out yet:
- if someone is going the Login page from outside the app
- if someone goes to the Login page from a direct URL (e.g., types https://circulate-staging.herokuapp.com/users/sign_in into their browser bar)
- if someone attempts to access a page that can only be accessed when logged in

avinashupadhya99:
Hey, can I work on this? If yes, please do assign this to me. Thank you.

avinashupadhya99:
Hey! I was just curious how do we know if an user who is not logged in is a CTL member? Currently there is a `Login/SignUp` option for non-logged in users in place of `Hold`. I can implement a feature where users are redirected back to the previous url after logging in. Please let me know your thoughts about this.

rkim00:
Hi @avinashupadhya99 , thanks for the question. I think that your implementation generally makes sense.

Note that there's currently a "Member Login" link in the upper-right. Currently, after clicking on this link, a member goes to https://circulate-staging.herokuapp.com/, which is incorrect; users who are just members should go to the My Loans page, and staff should go to the admin page (#174 is meant to fix this) ). You don't need to implement #174 (someone else is assigned to that), but whatever you do should largely leave that functionality the same.

So if you click on the "login" link in the listing for the item, and then sign-in, you should be redirected to the same item page. But if you click on the "Member Login" link in the upper-right, you should go to the My Loans page (or as currently wrongly implemented, to the https://circulate-staging.herokuapp.com/ ).

Does that all make sense? I'm not sure of the best way to implement that, but give it a shot. Let me know if you have any questions.

avinashupadhya99:
Yes got it. Thanks.

avinashupadhya99:
Hey @rkim00, I was thinking if we can redirect users to their previous page unless that previous page was the home page i.e, [ https://circulate-staging.herokuapp.com/ ]( https://circulate-staging.herokuapp.com/ ). If the user logs in/ signs up through the home page's `Log In/ Sign Up`, then we can redirect them to the respective pages which the issue #174 addresses.
[...]

PR #328, opened 2020-10-16
Redirect users to originating page after login

avinashupadhya99:
Users will be redirected to the previous page after login unless the previous page is the home page, in which case they will be redirected to admin dashboard or the member loans page based on their role.

Address issue #216


rkim00:
Hi @avinashupadhya99 , thanks for your work! Quick question—how will this behave if the user is coming from an external URL? Is there a risk that the user will get redirected to the external URL?

avinashupadhya99:
Hi @rkim00 , yes the user will be redirected to an external link if it is coming from that. Should I handle that part also?

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
+module Users
+  class SessionsController < Devise::SessionsController
+   def new
+     if  request.referer.present?
+     	self.resource = resource_class.create(sign_in_params)
+        store_location_for(resource, request.referer)
```
jim, line 6: Let's only store the referrer if it under the same domain that the application is running under. We don't want to redirect people to external sites.

jim (changes requested):
This is looking good. I'd like to see two things before merging this:

1. We don't want to redirect to other domains
2. It would be nice to add a test case to cover cases where a location was stored for both members and admins.

avinashupadhya99:
If the request is coming from another domain, it will redirect to the home page. Should it redirect to `/admin` or `/members/loans` respectively?

avinashupadhya99:
Hey, sorry I am not able to write the test case when there is a referer. I tried a lot of different ways of setting the `referer` for the request. I am not that used to writing test cases in rails. Any help would be appreciated. Thanks.

rkim00:
@avinashupadhya99 , yes, if a request is coming from another domain, it should redirect to `/admin` or `/members/loans` respectively.

The home page is not really used in our implementation; no user should ever really end up there except by accident or by typing in the URL directly. Most pages in the app are access through the SquareSpace portion of the site (https://chicagotoollibrary.org/), which sends users directly to different portions of the app, rather than the home page. Note that the logo in the upper-left corner sends users back to the SquareSpace site, not to the app home page.

test/controllers/application_controller_test.rb
```diff
@@ -7,15 +7,32 @@ class ApplicationControllerTest < ActionDispatch::IntegrationTest
...
-  test "should redirect to admin dashboard after login" do
+  test "should redirect to admin dashboard after login if there is no referer" do
     sign_in(@admin_user)
     post user_session_url
     assert_redirected_to admin_dashboard_path
   end
+
+  test "should redirect to previous page for admin after login if there is a referer" do
+    get new_user_session_path, headers: {referer: "http://www.example.com/original-page"}
+    post user_session_path, params: {user: {email: @admin_user.email, password: "password"}}
```
jim, line 24: I think part of the issue was that using the `sign_in` helper from Devise circumvents the code in `Users::SessionsController#new`, so that the referer, even if it was set correctly, wasn't getting checked. This test avoids that by making those requests manually.

test/controllers/application_controller_test.rb
```diff
@@ -7,15 +7,32 @@ class ApplicationControllerTest < ActionDispatch::IntegrationTest
...
+  end
+
+  test "should redirect to previous page for member after login if there is a referer" do
+    # @request.env['HTTP_REFERER'] = 'http://www.example.com/items'
+    sign_in(@user)
+    post user_session_path
+    assert_redirected_to items_path
+  end
+
+  test "ignores a poorly formed referer" do
```
jim, line 35: I added this test since parsing the referrer can fail if someone sends a request with garbage in it.

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+        if referer.host == request.host
+          self.resource = resource_class.create(sign_in_params)
+          store_location_for(resource, request.referer)  
+        end
+      end
+    rescue URI::InvalidURIError
+      # bad URI, ignore
+    ensure
+      super
+    end
```
jim, line 15: I updated this code to always use `request.referer` to access the referer and to check that the hosts are equal (before it was comparing the entire referer to just the host.

jim:
Thanks for updating this. I will take a look soon!

jim (changes requested):
I have a suggestion on how to refactor the code a bit, but the main blocker here is that there are two broken tests stemming from a route being renamedon the `development` branch. Otherwise, the tests look great!

jim (approved):

jim:
@avinashupadhya99 Thanks for working through this. I think it's good to go!

PR #333, opened 2020-10-18
Prevent the printed warning messages in system tests

kevinluo201:
# Problem
There are _Checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead._ during the system tests. The problem is the member's number is nil in those tests.

# Solution
## for test/system/member_profile_test.rb "member can view profile"
add :number to verified_member factory

## for test/system/member_verification_test.rb "verify pending member without membership"
do @member.reload after page changed so that the controller has done its action

### DIFF
--- a/test/factories/members.rb
+++ b/test/factories/members.rb
@@ -16,6 +16,7 @@ FactoryBot.define do
       address1 { "apt 3" }
 
       factory :verified_member do
+        sequence :number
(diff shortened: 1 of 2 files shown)


jim (approved):
Thank you for getting to the bottom of this!

PR #340, opened 2020-10-20
Adding the ability to lend a hold to a member as an admin. 

preethinarayan:
Addressing issue https://github.com/rubyforgood/circulate/issues/76

Also made a bug fix for a previous issue I worked on (https://github.com/rubyforgood/circulate/issues/75), and moved the helper method out of the controller into the helper. 

Screen Shot 2020-10-19 at 9 05 34 PM

@rkim00 @jim let me know if this looks alright or if I might have missed something.


jim (commented):
This looks good!

It would be nice to put the new functionality in a more RESTful controller instead of adding a POST with a custom method (`lend`). We have the `Admin::Members::HoldLoansController`, but that handles lending __all__ of the holds. Maybe we make a`Admin::Members::HoldLoanController` that has a `create` method to create the loan for a single hold?

The naming isn't ideal here in that there could be confusion between the two due to similarity. I don't have a better way to handle this in mind, but I would love any idea that you might have!

preethinarayan:
@jim I'm don't have a strong option either. I think having an `Admin::Members::HoldLoansController` works. I can move out the logic to that controller.

preethinarayan:
@jim on second thoughts, I was taking a look at the routes file here https://github.com/rubyforgood/circulate/blob/development/config/routes.rb#L69. It might be confusing to have another controller and route for a loan on an individual hold. Would it work if we used the same endpoint we currently have to lend all holds to work 'for a particular hold if specified or handle all holds for that member'? The other alternative is to have a `/bulk` endpoint to handle lending all holds. 
Thoughts on this?

jim (approved):
After thinking about this some more, let's just keep this as is. Thanks for responding to my feedback, though, and providing some additional options. Since there's not a clear best way to do this I've come around to keeping the `lend` method with the other interactions on the `Hold` like you said.

Issue #342, opened 2020-10-22
Member Appointments: Fail more gracefully when no appointments.


rkim00:
Currently on pages like `https://circulate-staging.herokuapp.com/account/appointments`, where a member has no appointments, it's simply blank beneath the "Upcoming Appointments" header.

When a Member has no appointments, display a message below this header that says, "You have no scheduled appointments."

ps9793:
I'll look into this one too. Thanks for the suggestion

ps9793:
Created a pull-request. Please check and let me know if it solves the issue

PR #349, opened 2020-10-22
Fail more gracefully when no appointments are present

ps9793:
Hey,
Just added a div which will show a message telling the user that there are no scheduled appointments when the variable `@appointments` is blank

#342

### DIFF
--- a/app/views/account/appointments/index.html.erb
+++ b/app/views/account/appointments/index.html.erb
@@ -1,6 +1,14 @@
 <div class="appointments-page">
   <h1 class="new_title">Upcoming Appointments</h1>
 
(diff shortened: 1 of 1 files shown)


ps9793:
The failure doesn't seem to be related to the changes I made, still I am not entirely sure.

Can you please check once and let me know

rkim00:
I'm re-running the tests... we've seen these errors before. Hopefully they'll pass when I re-run.

ps9793:
Ok thanks

ps9793:
Tests passed.
Thanks for the help @rkim00

jim (approved):
This is a nice improvement. Thanks!

Issue #347, opened 2020-10-22
Hide password inputs from member profile edit page (there is a separate password change page that handles this).


rkim00:
On the Member Edit Profile page (https://circulate-staging.herokuapp.com/account/member/edit) remove the Password and Password confirmation form fields and labels, since there is a separate page to edit the password. Since this form is shared with another page in the Librarian view, we'll just want to hide these two fields from this page.

ps9793:
can I start working on this ?

rkim00:
Sure, it's yours.

ps9793:
Thanks

ps9793:
Should I just remove the two fields or should I add something in it's place too ?

ps9793:
I have created a pull-request for the branch, let me know if it is as per your expectations

rkim00:
You should just hide/remove them, so that the space that they occupy is collapsed. Thanks for your contribution! @jim will review and merge when he has a second.

PR #351, opened 2020-10-22
Hide Password Inputs on Member Profile Edit Page

ps9793:
Hey,
Added an additional parameter to the partial `admin/members/_fields.html.erb` which is rendered to show the form fields

The additional argument is used to filter the password inputs from the form

This parameter is set as `false` in case of Member profile edit page and true elsewhere

#347


jim (approved):
This is great! Thank you.

ps9793:
Happy to help @jim

Issue #355, opened 2020-10-22
Member pages: Update text next to held item after scheduled.


rkim00:
Currently, on several Member's pages:
- My Holds pages (https://circulate-staging.herokuapp.com/account)
- Schedule an Appointment (https://circulate-staging.herokuapp.com/account/appointments/new)
- Upcoming Appointments (https://circulate-staging.herokuapp.com/account/appointments)

... the text next to a Hold is always displaying as "Ready for pickup. Schedule by [date]", even when the item has been scheduled for pick-up.
![appts](https://user-images.githubusercontent.com/7916357/96887058-2ffe4980-1452-11eb-97c5-2a022bd85757.png)

Fix this so that after an Item is scheduled (associated with an Appointment Date), this text instead reads
"Scheduled for pick-up [appointment date, [appointment time]".

Related to #356 , #357 .

kevinluo201:
May I try to solve this one? : )

rkim00:
Please! Thank you!

PR #363, opened 2020-10-23
#355 member pages update text

kevinluo201:
related issues:
* [https://github.com/rubyforgood/circulate/issues/355](https://github.com/rubyforgood/circulate/issues/355)
* [https://github.com/rubyforgood/circulate/issues/356](https://github.com/rubyforgood/circulate/issues/356)
* [https://github.com/rubyforgood/circulate/issues/357](https://github.com/rubyforgood/circulate/issues/357)
---
# Summary
1. add Member#upcoming_appointment_of to find the upcoming appointment of loan or hold
2. show the more suitable text for loans and holds status
3. hide the checkbox if loans or holds were scheduled

# Screen Shot
image
image
image


rkim00:
Hi @kevinluo201 . Thanks for your contribution! Generally looks good, but one quick note is that I know that our members will understand times better if they display in am / pm instead of 24-hour time. So I see that that the appointment times in the My Loans & Holds screen are displaying as "12:00 - 13:00" for example. We should probably display as "12pm - 1pm" or "12:00pm - 1:00pm", whatever is easier.

Also, we probably don't need to display the year, but the day of the week is very useful, so that folks don't need to think about what day of the week the appointment date is, and are less likely to get the wrong date. So instead of "10/24/2020," the better date format is probably "Sat, 10/24".

We can also clean these up later if necessary. Thanks again!

kevinluo201:
I'll modify that later : )

jim (approved):
This looks good to me. Thanks!

Issue #361, opened 2020-10-23
LoanHelpersTest fails intermittently


jim:
We're seeing this on a lot of CI failures:

```
Expected: "tomorrow"
  Actual: "today"
test/helpers/loans_helper_test.rb:13:in `block in <class:LoansHelperTest>'
```

I've also seen it locally. It usually goes away with a re-run. Let's make this test more reliable.

PR #364, opened 2020-10-23
#361 loan helpers test fails intermittently

kevinluo201:
# Related issues
[https://github.com/rubyforgood/circulate/issues/361](https://github.com/rubyforgood/circulate/issues/361)

# Summary
Time sensitive test is better to manually set the time to prevent timezone problem.

# Screenshots
before
image
after
image


rkim00:
I thought it might have to do with timezones! Thanks!

test/helpers/loans_helper_test.rb
```diff
@@ -8,9 +8,11 @@ class LoansHelperTest < ActionView::TestCase
   end
 
   test "humanize_due_date returns tomorrow when due date is tomorrow" do
-    @loan = build(:loan, due_at: 1.days.from_now)
-
-    assert_equal "tomorrow", humanize_due_date(@loan)
+    Timecop.freeze(Date.new(2020, 1, 1)) do
```
jim, line 11: Can we use the built-in `travel_to` method from Rails instead of timecop here?

jim (changes requested):
Thanks for taking a look at this and identifying a solution.

If possible, I'd like to avoid adding a gem dependency if we can accomplish the same thing using Rails built-in functionality. [Here's the doc page for travel_to](https://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html#method-i-travel_to).

test/helpers/loans_helper_test.rb
kevinluo201: I removed the timecop.
I also found that the root cause is the `from_now` method which use Time.current as input.
Time.current return UTC time that's why it has timezone problem
https://apidock.com/rails/ActiveSupport/Duration/from_now.

I changed to use Date.today + 1.day, which always uses local time, so we don't need change Time.now anymore.

jim (approved):

jim:
Thanks for fixing this for us!

PR #366, opened 2020-10-23
Librarian Member Notes 

chaserx:
Closes https://github.com/rubyforgood/circulate/issues/155

Screen Shot 2020-10-23 at 4 50 02 PM

Screen Shot 2020-10-23 at 4 50 13 PM

Screen Shot 2020-10-23 at 4 50 30 PM

### DIFF
--- a/app/controllers/admin/members_controller.rb
+++ b/app/controllers/admin/members_controller.rb
@@ -75,7 +75,7 @@ module Admin
       params.require(:member).permit(
         :full_name, :preferred_name, :email, :phone_number, :postal_code,
         :desires, :reminders_via_email, :reminders_via_text, :receive_newsletter, :volunteer_interest,
-        :notes, :status, :address1, :address2, pronouns: []
+        :status, :address1, :address2, pronouns: []
       )
     end
   end
--- a/app/controllers/admin/notes_controller.rb
+++ b/app/controllers/admin/notes_controller.rb
@@ -41,8 +41,10 @@ class Admin::NotesController < ApplicationController
   end
 
   def load_parent
-    @parent = if params[:item_id]
-      Item.find(params[:item_id])
+    if params[:item_id]
+      @parent = Item.find(params[:item_id])
+    elsif params[:member_id]
+      @parent = Member.find(params[:member_id])
     end
     raise ActiveRecord::RecordNotFound unless @parent
   end
--- a/app/javascript/stylesheets/application.scss
+++ b/app/javascript/stylesheets/application.scss
@@ -707,7 +707,7 @@ form.membership-amount {
 }
 
 .admin {
-  .member-lookup-items {
+  .member-lookup-items, .member-notes {
     background: white;
     padding: .7rem .5rem;
     .info-box {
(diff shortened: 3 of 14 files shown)


chaserx:
@rkim00 @jim OK. So, see the screenshots above for what I've got working - a form for notes, notes in the sidebar, and a modal that displays the notes. 

I'm unsure about form availability. That is, should it be visible all the time or only activated by the `add a note` button. Currently it's visible all the time because I'm not using the toggle functionality correctly. I will likely need some help to make the Stimulus part work as intended - I think it's something small like a typo or the wrong class that I can't see anymore that's preventing it from working properly.

rkim00:
Hi @chaserx , thanks for your contribution! I doubt we'll need formatting on the notes, but I assume you get all of those options (bolding, bullets, etc.) for free with the gem/module you're using?

I think it's probably best to hide the form when not specifically asked for, particularly since other aspects of the member details aren't immediately editable, but I don't think it's a big deal if it's available.

chaserx:
@rkim00 yep. tyring to keep to the same pattern as notes used elsewhere so yes we're getting that formatting for free.  Note that the formatting is stripped in the sidebar because rendering that rich text in the partial resulted in errors that we can fight later if needed. The formatting should appear in the modal however.

chaserx:
I'll continue to work out how to make the stimulus toggle work, but would appreciate some pointers.

rkim00:
Thanks! Wish I could help, but I don't know stimulus at all. @jim may be able to help when he has a chance to review.

chaserx:
ok. I think I better understand how the toggle feature works in stimulus now and I've noticed a glaring but with the form that I'm attempting to fix.

chaserx:
Here's the current flow.

jim (changes requested):
This all looks great. I tried things out locally and they work as intended. Thanks a lot for the contribution!

I realize now that we have a naming collision- there is already `notes` field on `Member` that is filled by a text area members see when they sign up. This field is used to ask for a little info about the member, what kind of projects they do, etc. Adding these notes overwrites that column, and I didn't realize this when I wrote this issue originally.

How about we rename the `member.notes` column in the database to `bio`, since that is a more accurate description of what goes in that field and will avoid the collision?

The other little thing I noticed is that there is an "Edit" button in the notes modal. I think we can pull editing notes out into a separate story unless you have something in mind to easiliy add that. Doing it inside the modal might be tricky, so probably just going to a new page for that is fine (it's not going to be used that frequently). I don't think seeing that "edit" test there is blocking this being merged.

chaserx:
@jim thanks for the review. I've replaced `Member.notes` with `Member.bio` good catch there. 

As for the edit link in the modal. I think something is stripping out the the link formatting for that edit button but I don't see it. Moving that to another issue would be a good idea.

jim (approved):

jim:
This is looking good. Thanks again!

jim:
And for anyone reading along, #370 is the follow up to this one that includes editing member notes.

PR #368, opened 2020-10-24
Search by item name on member's page #160

### LINKED ISSUE #160: Librarian view: on Member page, allow tool look-up to happen not just through item #, but through tool name as well.

matisnape:
Resolves #160 

TODO:

- [x] Implement separate search for item name
- [x] Render form and search results into member view
- [ ] Render search results dynamically
- [ ] Add tests
- [ ] Create a separate scope for item by name results (currently it's 10 limit)
- [ ] Maybe implement paginations of some sort
- [ ] Styling polishing


app/views/admin/members/_search_results.erb
```diff
@@ -0,0 +1,53 @@
...
+            <div class="column col-4">
+              <% if item.image.attached? %>
+                <%= link_to admin_item_path(item) do %>
+                  <%= image_tag url_for(rotated_variant(item.image, resize_to_limit: [200, 140])), class: "p-centered" %>
+                <% end %>
+              <% else %>
+                <div class="image-placeholder"></div>
+              <% end %>
+
+              <%= link_to "Cancel", admin_member_path(@member, anchor: "checkout"), class: "btn" %>
```
jim, line 47: I think we'll want to pull this out of the loop so there is only one cancel button per set of search results.

jim (commented):
This looks pretty good! I think we can proceed with the new code path you're building replacing the existing one. It will be better for the app to be able to handle rendering more than one search result, regardless of how the search results are determined. We can iterate a little on how to handle the number vs. name search when we get to that point. It's probably worth adding a scope or something to `Item` so that we can use the same behavior in other places where we look for items by number/name/etc.

jim:
This was an inadvertent close- GitHub did that when I was renaming the development branch to `main`.

jim:
@matisnape I hope things are OK with you! 🤞🏻 Is this something you still want to work on, or should I ask someone else to pick it up? No pressure at all if circumstances aren't good right now.

jim:
I'm closing this. If you want to pick it up at some point, feel free to re-open 😄

matisnape:
Hi @jim !
Sorry for no activity on my side, unfortunately I was dealing with some health issues for the past couple of months.
I intend to revisit this feature during Hacktoberfest and probably do it from scratch, since I believe that the codebase might have changed significantly during that time 😅
Will keep yiu posted ;)

PR #371, opened 2020-10-25
359-added rule to hide button

italomatos:
# What it does

Hide the "remove" link next to a hold after it's been scheduled for pick-up (i.e., associated with an appointment).

# Why it is important

issue: https://github.com/rubyforgood/circulate/issues/359

# Implementation notes

* Added appointment relation in Hold model
* Created a helper method to cover the rule in card.


jim (approved):
Looks good, thanks!

PR #372, opened 2020-10-26
include loan ending in item availability

chaserx:
# What it does

_Describe the changes that you've made to the application in this PR._
 
Closes https://github.com/rubyforgood/circulate/issues/343
This should prevent a Member from scheduling a pickup for a hold if an item has not been returned. 

# Why it is important

_Make a case for merging this work. This might include linking to a relevant issue._

This should provide a more consistent experience for the user. 

# UI Change Screenshot

Screen Shot 2020-10-25 at 7 04 45 PM

# Implementation notes

After spelunking, I needed a mechanism to test for pickup availability both in the holds helper and to control the schedule button in the account index view. So, I chose to update the `ready_for_pickup?` method on Hold versus pushing more duplicate logic into the helper and view. 

When testing locally, I found a test failure in loan_test that did not seem seem to be related. We'll see if it fails on CI

### DIFF
--- a/app/models/hold.rb
+++ b/app/models/hold.rb
@@ -27,7 +27,7 @@ class Hold < ApplicationRecord
   end
 
   def ready_for_pickup?
-    previous_active_holds.empty? #&& item.loans.last.ended?
+    previous_active_holds.empty? && item.available?
   end
 
   def upcoming_appointment
--- a/app/models/hold.rb
+++ b/app/models/hold.rb
@@ -27,7 +27,7 @@ class Hold < ApplicationRecord
   end
 
   def ready_for_pickup?
-    previous_active_holds.empty?
+    previous_active_holds.empty? #&& item.loans.last.ended?
   end
 
   def upcoming_appointment
--- a/app/helpers/holds_helper.rb
+++ b/app/helpers/holds_helper.rb
@@ -22,7 +22,7 @@ module HoldsHelper
     if appointment
       "Scheduled for pick-up at #{format_date(appointment.starts_at)} " +
       format_time_range(appointment.starts_at, appointment.ends_at)
-    elsif previous_holds_count == 0
+    elsif hold.ready_for_pickup?
       "Ready for pickup. Schedule by #{format_date(hold.created_at + 7.days)}"
     else
       "##{previous_holds_count} on wait list"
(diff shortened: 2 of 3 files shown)


app/models/hold.rb
```diff
@@ -27,7 +27,7 @@ def previous_active_holds
   end
 
   def ready_for_pickup?
-    previous_active_holds.empty?
+    previous_active_holds.empty? #&& item.loans.last.ended?
```
jim, line 30: I agree that extending this method is the way to go here. I _think_ a simpler way to handle this would be:

```suggestion
    previous_active_holds.empty? && item.available?
```

This has the nice side benefit of handling tools that aren't uniquely numbered (these are the ones where `item.borrow_policy.uniquely_numbered == false`).

jim (changes requested):
I requested one change in the `ready_for_pickup?` method.

app/models/hold.rb
chaserx: Gotcha. Thank you for the suggestion. I'll try that again. I think I encountered some test weirdness when I tried this first.

chaserx: @jim Maybe you can help me understand why using the `available` method, we get this failure. https://app.circleci.com/pipelines/github/rubyforgood/circulate/660/workflows/ca3ea2bf-f788-46c5-b8a1-9bf15cd124eb/jobs/3483/parallel-runs/0/steps/0-112

I'm puzzled. I step through it, but don't see it. I wonder if there's something with the scope checked_out_exclusive_loan that's used. WDTY?

chaserx: Or maybe it's a data problem and we need to use a slightly different factory like `ended_loan` or `nonexclusive_loan` - I think I tried nonexclusive_loan and was able to get a passing test, but I don't know enough about loan exclusivity to make an informed choice.

jim: I re-ran the tests on CI, and they passed, and then everything was looking fine locally until I ran the entire suite in the same order as it was on CI when it failed: `rails test --seed 31967`. Then I was able to reproduce the failure.

chaserx: weird. thank you for investigating. maybe that points to a test data problem then.

jim (approved):

jim:
Thanks for working on this!

app/models/hold.rb
jim: For anyone following along, here is the followup issue: https://github.com/rubyforgood/circulate/issues/382

Issue #354, opened 2020-10-22
Member can cancel appointment


rkim00:

avinashupadhya99:
Hey @rkim00 , I was thinking about having an option where members can either **cancel** their appointments or **reschedule** their appointments. Thoughts?

rkim00:
Yes, that's a perfectly good option as well. We haven't really thought about how this would work from a UI standpoint yet. What are your ideas?

avinashupadhya99:
I think we should just handle the cancelling part with this issue and take up the rescheduling part in #369

rkim00:
Hi @avinashupadhya99 , thanks for taking the initiative with this! Really appreciate it. Looks pretty good to me, although @jim may have more to say. Two things I wanted to check on:

1) Am I correct that the appointment itself is cancelled, but the associated hold is unaffected? So that after the member cancels an appointment, they keep their place in line for whatever hold they have?

2) Should we have any new tests to cover this functionality? (I may have missed them in the commit.)

avinashupadhya99:
Yes, only the appointment will be cancelled. The holds are unaffected by this. 
I'll see if I can add test cases for this.

PR #373, opened 2020-10-26
Member can cancel appointment

avinashupadhya99:
# What it does

Added cancelling of member's appointments by the member

# Why it is important

Fixes Issue #354 

# UI Change Screenshot

cancel_appointment

# Implementation notes

- Added the destroy method for appointments under the Account module
- Added test case for the same.

### DIFF
--- a/app/controllers/account/appointments_controller.rb
+++ b/app/controllers/account/appointments_controller.rb
@@ -35,7 +35,7 @@ module Account
     end
 
     def destroy
-      Appointment.find(params[:id]).destroy
+       current_user.member.appointments.find(params[:id]).destroy
       redirect_to account_appointments_path, flash: {success: "Appointment cancelled."}
     end
 
--- a/app/controllers/account/appointments_controller.rb
+++ b/app/controllers/account/appointments_controller.rb
@@ -34,6 +34,11 @@ module Account
       end
     end
 
+    def destroy
+      Appointment.find(params[:id]).destroy
+      redirect_to account_appointments_path, flash: {success: "Appointment cancelled."}
+    end
+
     private
 
     def appointment_params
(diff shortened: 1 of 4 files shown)


test/controllers/account/appointments_controller_test.rb
```diff
@@ -0,0 +1,29 @@
...
+      @appointment.save
+
+      sign_in @user
+    end
+
+    test "should cancel appointment without affecting holds" do
+      
+        delete account_appointment_path(@appointment)
+        assert_nil assigns(:appointment)
+        assert Hold.find_by(member_id: @member.id) != nil, "Member holds should not be deleted when an appointment is cancelled"
```
jim, line 25: I think the intention of this test could be a little more explicit if we assert against the total number of holds:

```suggestion
        assert_equal 1, member.holds.count, "Member holds should not be deleted when an appointment is cancelled"
```

app/controllers/account/appointments_controller.rb
```diff
@@ -34,6 +34,11 @@ def create
       end
     end
 
+    def destroy
+      Appointment.find(params[:id]).destroy
```
jim, line 38: I think this needs to be scoped to the appointments for the current member to prevent deleting another person's appointments:

```suggestion
      current_member.appointments.find(params[:id]).destroy
```

jim (changes requested):
Thanks for the contribution! It will be great to allow members to perform this for themselves.

The blocker to merging this is that we need to scope to the current member's appointments when deleting.

app/controllers/account/appointments_controller.rb
avinashupadhya99: Thanks, can't believe I missed the scoping part. Thanks for the suggestion. I'll make the necessary changes.

jim (approved):

jim:
Thank you for making those changes!

Issue #374, opened 2020-10-27
Fix footer display on Items page


rkim00:
Footer ("For questions and issues...") is displaying in middle of page on items page (e.g., https://circulate-staging.herokuapp.com/items/36 ).
![footer](https://user-images.githubusercontent.com/7916357/97314034-a974ae00-183d-11eb-817f-def79ad4994d.png)

Fix so that it displays at the bottom of the page.

efgalvao:
Can I try this one ?

rkim00:
You got it! Thanks!

PR #380, opened 2020-10-27
Fixed footer display on Items page.

efgalvao:
Changed footer specs to fix issue #374

# Why it is important

Fixes the issue #374.

# UI Change Screenshot

After
Screenshot_2020-10-27 Circulate

# Implementation notes

* Choose for a fixed footer.

### DIFF
--- a/app/javascript/stylesheets/application.scss
+++ b/app/javascript/stylesheets/application.scss
@@ -1085,8 +1085,6 @@ ul.hold-items {
   }
 }
 
-
-
 .tool-card .modal {
   .stats {
     margin-bottom: 1.2rem;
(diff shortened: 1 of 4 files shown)


jim (changes requested):
Thanks for the contribution! This implementation works well on the item page, but on the item index it causes the footer to appear on top of the listing. We want the footer to appear at the bottom of the page and not on top of the other page content.

I'd like to see if we can fix this without adding more CSS. I think the issue is related to the structure of the markup on these pages and not on the styling. Here's the structure of the item index, which is displaying properly:

Screenshot 2020-11-01 093710

And here it is for the item page, where the footer is showing up incorrectly:

Screenshot 2020-11-01 093556

efgalvao:
Screenshot_2020-11-01 Circulate
Screenshot_2020-11-01 Circulate(1)
I think now is the way you wanted.

jim (approved):

jim:
@efgalvao This is great! Thank you for making those changes.

PR #381, opened 2020-10-28
Display librarian appointment time next to date

### LINKED ISSUE #378: Librarian Appointment Page - display time next to date
Currently, on the View/Edit Appointments page, we only display the date of the appointment, not the time as well. 
appointment

We should also display the appointment time! Display time in the format "1pm–2pm". Note that "–" is an en-dash, not a hyphen (-).

time can be displayed next to or below the date, whatever seems to look better in the page.

laasem:
# What it does

Adds appointment start and end time in the following format: "1pm – 2pm" next to the date on the librarian appointment page.

# Why it is important

Fixes #378.

# UI Change Screenshot

**Before:**
before_time

**After:**
after_time

# Implementation notes

Since the `hour` time format uses `%l` which adds a blank space before the hour digit, I used spaces around the en dash, unlike the example in the issue which had no spaces.

### DIFF
--- a/app/views/admin/appointments/show.html.erb
+++ b/app/views/admin/appointments/show.html.erb
@@ -13,6 +13,7 @@
         <div class="column col-3">
           <h6>Appointment Date</h6>
           <h6><%= l current_appointment.starts_at, format: :date %></h6>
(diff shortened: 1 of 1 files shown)


jim (approved):
This is great! Thanks for the contribution.

PR #383, opened 2020-10-31
Fix renewal of A tools

### LINKED ISSUE #344: Fix renewal of A tools
Renewing of A tools no longer seems to be appearing as an option for member. Debug and fix this. Might be a merge problem.

See #173 and #151 for intended functionality.

So far seeing this in `verifiedmember@example.com` and `admin@chicagotoollibrary.org` on Staging. This might be a data issue with these two accounts, as partially documented in #341 ; I haven't been able to test in other accounts yet.

littleforest:
# What it does

closes #344 

* Added `member_renewable` field to the admin borrow policy form
* Added an A tool to the seeds file
* Updated the `member_renewable?` method for loans to be false if the loan has an `ended_at` date

# Why it is important

I suspect that this isn't working on staging because the `member_renewable` flag for the A policy is set to false.

# UI Change Screenshot

**Admin form before**:

Screen Shot 2020-10-30 at 4 56 22 PM

**Admin form after**:
Screen Shot 2020-10-30 at 4 56 52 PM

**Member page with Renew button**

Screen Shot 2020-10-30 at 5 01 14 PM
Screen Shot 2020-10-30 at 5 01 20 PM

# Implementation notes
* Check that the `member_renewable` flag for the A policy is set to true
* I tried to add the `member_renewable` field to the table on the borrow policies index page, but the alignment ended up wonky, and I couldn't see immediately how to fix it.

### DIFF
--- a/app/controllers/admin/borrow_policies_controller.rb
+++ b/app/controllers/admin/borrow_policies_controller.rb
@@ -24,7 +24,7 @@ module Admin
     end
 
     def borrow_policy_params
-      params.require(:borrow_policy).permit(:name, :duration, :fine, :fine_period, :uniquely_numbered, :code, :description, :default, :renewal_limit)
+      params.require(:borrow_policy).permit(:name, :duration, :fine, :fine_period, :uniquely_numbered, :code, :description, :default, :renewal_limit, :member_renewable)
     end
   end
 end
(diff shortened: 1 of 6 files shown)


jim (approved):
A solid improvement! Thanks for tackling this one.

PR #384, opened 2020-10-31
Fix flaky test

### LINKED ISSUE #382: Test failure with specific seed value
Running `/rails test --seed 31967` will result in the following test failure:

```
Failure:
LoanTest#test_is_member_renewable_if_borrow_policy_allows_member_renewal_is_within_original_loan_duration_and_has_not_exceeded_limit [/Users/jimb/code/circulate/test/models/loan_test.rb:285]:
Expected false to be truthy.
```

This tests doesn't fail when running `rails test test/models/loan_test.rb`, so it seems to be related to how the test data is being setup. I think it has something to do with the `Loan#within_borrow_policy_duration?`.

I'm not totally sure if we need this method, or what it is used for. It's possible that we should just remove it and its uses; otherwise, we should add some tests abound its behavior.

invacuo:
# What it does

Fixes the flaky test mentioned in #382 

# Why it is important
Green tests at all times 😄 

# Implementation notes
- The test was failing due to timezone differences.
- When creating the due date in the test, the code is using `X.days.from_now` which is a rails method and uses the timezone specified in the app
- When humanizing the dates, the code is using `Date.today` which is a ruby method and uses the system time (UTC)
- Changed everything to use the rails date/time methods

Closes #382

### DIFF
--- a/app/helpers/loans_helper.rb
+++ b/app/helpers/loans_helper.rb
@@ -13,9 +13,9 @@ module LoansHelper
   end
 
   def humanize_due_date(loan)
-    if loan.due_at.to_date == Date.today
+    if loan.due_at.to_date == Time.zone.today
       "today"
-    elsif loan.due_at.to_date == Date.tomorrow
+    elsif loan.due_at.to_date == Time.zone.tomorrow
       "tomorrow"
     else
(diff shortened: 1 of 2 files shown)


jim (approved):
Thanks for the contribution!

PR #385, opened 2020-11-02
379-fixed Items to be checked-out

italomatos:
# What it does

This is a fix to update item count after checking out item, like described in card: https://github.com/rubyforgood/circulate/issues/379


jim (approved):
This looks good. The test breaking appears to be unrelated.

PR #406, opened 2020-11-18
Adding the Edit Note and Delete Note functionality for members and it…

preethinarayan:
…ems in the admin view

# What it does

Added the 'Edit' and 'Delete' Notes functionality for members and items in the admin view.

# Why it is important

https://github.com/rubyforgood/circulate/issues/370

# UI Change Screenshot

Screenshot for edit item note
Screen Shot 2020-11-17 at 11 22 08 PM

Screenshot for edit member note
Screen Shot 2020-11-17 at 11 21 54 PM

# Implementation notes

- Edit note on the member page goes to a new page where you can either update or delete the note. I've removed the cancel since it wasn't rendering the admin member page correctly (maybe there is a way to do it and I don't know)
- Edit note on the item page opens the editor inline.
- Update and delete functionality both work and I've added tests for them
- Open to feedback regarding the UI flow for both the use cases

cc: @jim


jim (approved):
This looks good to me! I appreciate you taking the time to add tests covering the new behavior.

Long-term, it would be nice to align the UIs of the item and member notes. But I'm not sure how to do that yet, and I think what you have here is a step in the right direction.

I found a few semi-unrelated things to tweak in the notes code when reviewing your changes that I am going to fix in the process of merging this in: the `Admin::NotesController` should be inheriting from `Admin::BaseController`, and the edit link in the modal was opening in a new window due to some JS code in the Stimulus modal controller.

Issue #402, opened 2020-11-15
Upgrade to Ruby 2.7.2


jim:
When we deploy to Heroku, this message is displayed:

```
remote: ###### WARNING:
remote:
remote:        There is a more recent Ruby version available for you to use:
remote:
remote:        2.7.2
remote:
remote:        The latest version will include security and bug fixes. We always recommend
remote:        running the latest version of your minor release.
remote:
remote:        Please upgrade your Ruby version.
remote:
remote:        For all available Ruby versions see:
remote:          https://devcenter.heroku.com/articles/ruby-support#supported-runtimes
```

This upgrade probably just means bumping the Ruby version listed in the `Gemfile` and `Dockerfile` and making sure that the tests all pass.

tullur:
Hey, I'll upgrade Ruby version.

rkim00:
Thanks for the assist!

PR #407, opened 2020-11-18
Bump Ruby Version

tullur:
# What it does

Bumps Ruby version to 2.7.2

# Why it is important

Issue: #402


jim (approved):
This is great! Thank you.

PR #420, opened 2020-12-31
[Admin] Add Items in Maintenance report

SpencerLindemuth:
# What it does

Add report to admin portal that shows all tools currently in maintenance, and tab in reports dropdown to access it

# Why it is important

Closes issue #419 

# UI Change Screenshot

Screen Shot 2020-12-30 at 8 45 52 PM
Screen Shot 2020-12-30 at 8 46 02 PM

# Implementation notes

I think this is a pretty straightforward issue, with a barebones implementation. I can add additional functionality if need be.


test/controllers/admin/items_in_maintenance_controller_test.rb
```diff
@@ -0,0 +1,15 @@
...
+class Admin::ItemInMaintenanceControllerTest < ActionDispatch::IntegrationTest
+  include Devise::Test::IntegrationHelpers
+
+  setup do
+    @user = create(:admin_user)
+    sign_in @user
+  end
+
+  test "should get index" do
+    get admin_items_in_maintenance_index_url
```
jim, line 12: If we create an item before making the request, the view will render the table exercise that code as well. There are factories for most of the models in `test/factories`, including one for Item:

```suggestion
    create(:complete_item, status: Item.statuses[:maintenance])
    get admin_items_in_maintenance_index_url
```

app/controllers/admin/items_in_maintenance_controller.rb
```diff
@@ -0,0 +1,8 @@
+module Admin
+    class ItemsInMaintenanceController < BaseController
```
jim, line 2: We're starting to move the report controllers into the `Admin::Reports` namespace, and it would be great to put this one there as well.

jim (changes requested):
Thanks for working on this! The implementation looks good overall and there are few small changes I'd like to see before merging.

jim (approved):

jim:
@SpencerLindemuth This looks great!

Issue #263, opened 2020-10-03
Membership columns are named incorrectly


jim:
`memberships.started_on` and `memberships.ended_on` store timestamps but are named using `_on` which indicates that they should be dates. Rename these columns to `ended_at` and `started_at` using a database migration to align their names with the values they store.

jim:
This is dupe of #264

verma-pulkit:
Hey Jim,I would like to take this up. And try my best.

jim:
@verma-pulkit Sounds great. I'll close the other one then and assign this one to you.

dannyflatiron:
The migration would only change the column names, not the datatype?

dannyflatiron:
Hello, is this issue still available?

akp2603:
Hey @jim is this issue still available, would love to fix this.

rkim00:
@verma-pulkit are you still working on this issue? No worries if you still are, but if not, some other people have been asking to take it.

efgalvao:
Hi, I'm learning ruby and would like to try to fix this.

verma-pulkit:
@rkim00  I am sorry, I got stuck in college work. Sorry would not be able to work on this.

rkim00:
Thanks for the update @verma-pulkit .

@dannyflatiron @akp2603 @efgalvao would any of you like to take this issue?

akp2603:
I can help with this. You can go ahead and assign it to me.

rkim00:
Thanks! Assigned!

leahriffell:
Hi @akp2603! Are you still working on this? If not, I can pick it up.

akp2603:
@leahriffell sure, you can pick it up.

jim:
@leahriffell If you still want this one, it's all yours 😄

PR #421, opened 2020-12-31
[#263] Rename memberships started on and ended on columns

gregblake:
# What it does

This PR contains the following changes:

1. A database migration to rename two columns on the `memberships` table.
2. Application and test level changes to reflect the renamed attributes on the `Member` model.

Relevant quote from https://github.com/rubyforgood/circulate/issues/263:

> memberships.started_on and memberships.ended_on store timestamps but are named using _on which indicates that they should be dates. Rename these columns to ended_at and started_at using a database migration to align their names with the values they store.

# Why it is important

This PR closes https://github.com/rubyforgood/circulate/issues/263.

# UI Change Screenshot

There are no UI changes on this PR. But I thought a before/after comparison of the Membership records would be helpful, so I'm including this screenshot:

before_and_after_migration

# Implementation notes

* I didn't include previous database migrations when I renamed the attributes.
* The tests are green locally, with the exception of some system tests which are also failing on the [main branch](https://gist.github.com/gregblake/92b0370e378d28d8fb59d2e5d04cff4a).
* It would be wise for someone who is knowledgable about how librarians and members use the application to do a quick regression test to make sure I didn't miss anything related to the name changes.

### DIFF
--- a/app/controllers/admin/reports/memberships_controller.rb
+++ b/app/controllers/admin/reports/memberships_controller.rb
@@ -16,7 +16,7 @@ module Admin
           .left_joins(memberships: :adjustment)
           .select("email", "members.id", "status", "full_name", "preferred_name", "loans_count",
             "count(memberships.id) as memberships_count",
-            "max(memberships.ended_on) as expires_on",
+            "max(memberships.ended_at) as expires_on",
             "array_agg(adjustments.amount_cents ORDER BY adjustments.created_at ASC) as amounts")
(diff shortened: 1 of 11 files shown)


jim (approved):
This looks great! I will do some manual QA before merging.

gregblake:
Thanks @jim. Let me know if anything looks off, and I'll make additional changes.

jim:
Everything is looking good.

PR #439, opened 2021-01-27
add database credentials to dotenv example

italomatos:
# What it does

This PR has goal to add Database credentials to .env exemple


jim (approved):
Thanks for the fix!

PR #444, opened 2021-01-29
443-remove pronoun and custom_pronoun references and columns from members

italomatos:
# What it does

Remove pronoun and custom_pronoun from member table, and remove all reference about these columns.

Issue link: https://github.com/rubyforgood/circulate/issues/443

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -20,7 +20,6 @@ class Member < ApplicationRecord
   has_many :notes, as: :notable
 
   PRONOUNS = ["he/him", "she/her", "they/them"]
-  enum pronoun: [:"he/him", :"she/her", :"they/them"]
   enum id_kind: [:drivers_license, :state_id, :city_key, :student_id, :employee_id, :other_id_kind]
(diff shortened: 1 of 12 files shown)


app/models/member.rb
```diff
@@ -20,14 +20,13 @@ class Member < ApplicationRecord
   has_many :notes, as: :notable
 
   PRONOUNS = ["he/him", "she/her", "they/them"]
-  enum pronoun: [:"he/him", :"she/her", :"they/them", :custom_pronoun]
+  enum pronoun: [:"he/him", :"she/her", :"they/them"]
```
jim, line 23: I think we can remove this line entirely since we're removing the `pronoun` column.

jim (changes requested):
This looks good! There is one modified line that can be removed entirely. Thanks for the contribution.

app/models/member.rb
italomatos: Make sense 👍🏻

jim (approved):
This is great! Thanks.

PR #446, opened 2021-01-30
Fix default admin email in README

pjsier:
# What it does

I went through the setup instructions and it looks like the default admin email has changed in `seeds.rb`, but not in the README. The example.com version works, so I updated the docs here.

# Why it is important

Without it I wasn't able to log in after running the setup commands

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ The following third party services are used:
 
 ## Development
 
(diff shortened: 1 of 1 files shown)


jim (approved):
Thank you for fixing this!

PR #447, opened 2021-01-30
fix database environment variables in dotenv.example

italomatos:
# What it does

PR #445 change postgres credentials variables , and its necessary to update these on .env.example


jim (approved):
Good catch!

PR #449, opened 2021-02-01
Update membership Square order number

### LINKED ISSUE #441: Square order numbers should not just be the ID of the member.

Our Square order numbers just use the member number, which means that now that we're doing renewals, we have duplicates!

We need a new numbering scheme. The order number is really just a reference, so having it be human-readable is pretty important. They don't need to be sequential (we have other ways to count memberships).

**Describe alternatives you've considered**
I was thinking about something that used the date of purchase, so like `20210131-154`, where the first part is the date of purchase and part after the dash is the member number. I don't think we'll need to support more than one purchase per member per day.

**Implementation notes**

The order number is called the `reference_id` in the Square API. [We set that value here](https://github.com/rubyforgood/circulate/blob/main/app/lib/square_checkout.rb#L18).

Testing things connected to the Square API are a little bit involved to setup, but I think the tests [in this file](https://github.com/rubyforgood/circulate/blob/main/test/lib/square_checkout_test.rb) will provide enough to complete this story. I'm happy to verify in a Square sandbox account when we get to that point.

pjsier:
# What it does

Updates the Square reference ID for membership orders to include the membership creation date before the member ID.

# Why it is important

Closes #441

# Implementation notes

* I made `date` a parameter of `checkout_url` and passed `Date.today` from the controller because I'm assuming that will always correspond to the membership creation date. Let me know if any of that isn't what you're expecting!

### DIFF
--- a/app/controllers/renewal/payments_controller.rb
+++ b/app/controllers/renewal/payments_controller.rb
@@ -17,7 +17,7 @@ module Renewal
         return
       end
 
-      result = checkout.checkout_url(amount: @form.amount, email: @member.email, return_to: callback_renewal_payments_url, member_id: @member.id)
+      result = checkout.checkout_url(amount: @form.amount, email: @member.email, return_to: callback_renewal_payments_url, member_id: @member.id, date: Date.today)
       if result.success?
         redirect_to result.value
       else
--- a/app/controllers/renewal/payments_controller.rb
+++ b/app/controllers/renewal/payments_controller.rb
@@ -17,7 +17,7 @@ module Renewal
         return
       end
 
-      result = checkout.checkout_url(amount: @form.amount, email: @member.email, return_to: callback_renewal_payments_url, member_id: @member.id, date: Date.today)
+      result = checkout.checkout_url(amount: @form.amount, email: @member.email, return_to: callback_renewal_payments_url, member_id: @member.id, date: Date.current)
       if result.success?
         redirect_to result.value
       else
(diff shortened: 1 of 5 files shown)


app/controllers/renewal/payments_controller.rb
```diff
@@ -17,7 +17,7 @@ def create
         return
       end
 
-      result = checkout.checkout_url(amount: @form.amount, email: @member.email, return_to: callback_renewal_payments_url, member_id: @member.id)
+      result = checkout.checkout_url(amount: @form.amount, email: @member.email, return_to: callback_renewal_payments_url, member_id: @member.id, date: Date.today)
```
jim, line 20: This a fairly minor thing, but in Rails `Date.today` isn't timezone aware. I think we want to use `Date.current` instead. [This article has a list of does and don'ts](https://thoughtbot.com/blog/its-about-time-zones#do-use) and I usually use it as a cheat sheet because I can't remember the specifics usually.

test/lib/square_checkout_test.rb
```diff
@@ -111,6 +112,7 @@ class SquareCheckoutTest < ActiveSupport::TestCase
         email: member.email,
         return_to: "http://example.com/callback",
         member_id: member.id,
+        date: Date.new(2021, 1, 1),
```
jim, line 115: I think this should use `Time.zone.local` instead of `Date.new`.

jim (changes requested):
Outside of a timezone thing, this is great!

app/controllers/renewal/payments_controller.rb
pjsier: Ah whoops, thanks for catching that and the article!

jim (approved):

jim:
@pjsier Thanks for making those adjustments!

PR #450, opened 2021-02-04
Prevent item audit from being created if item categories do not change

### LINKED ISSUE #417: Updating an item mistakenly records that the categories have changed
**Describe the bug**
When a user goes to the edit item screen in the librarian view and clicks "Update", an edit is recorded in the item's history that a change was made to the item's categories- even if the categories weren't changed. 

**To Reproduce**
1. Go to an item's page in the librarian view
2. Click on "Edit"
3. Click on "Update Item"
4. Look at the "Edit History" tab on the item's page.

**Expected behavior**
The categories should only be recorded if they are changed.

**Screenshots**
image

SpencerLindemuth:
# What it does

This prevents an item history audit event from being created if an items categories were not updated

# Why it is important

Seeing too many edits is overwhelming and confusing for users. Closes #417 

# UI Change Screenshot

Screen Shot 2021-02-03 at 6 15 27 PM

Editing categories still audit as expected, blank updates do not create entry

# Implementation notes

Also fixed issue when dev data is created with no user for item creation, causing crash on item history screen


jim (changes requested):
Thanks for tracking this one down!

It would be nice to add a unit test to cover this behavior since it's kind of complex. There is a somewhat similar test [in item_test.rb](https://github.com/rubyforgood/circulate/blob/main/test/models/item_test.rb#L51-L59) that could be used as a starting point.

jim (approved):
Thanks for adding the test!

Issue #123, opened 2020-09-05
Member can request renewal for B & C tools


rkim00:
As a member, I want to be able to request to renew "B" & "C" category tools right from the app, so that I don't have to write an e-mail to do so.

On the My Loans page, next to B & C category tools that are checked-out, Members should get a "request renewal" link (or button) if
- the length of time between today and the due date is < the standard loan period for the item AND
- there are no pending holds for that item.

After clicking on this link:
- renewal request is stored in the database (model/table probably doesn't exist yet) as pending
- Link should change to plain text that reads, "Renewal requested. We'll e-mail you an update."

Renewal request should be associated with at least this information:
- date of request (probably time as well)
- member making the request
- loan requested to be renewed

See mock-ups:
https://drive.google.com/file/d/1sbf62tG2x7nUCHX5DN8_neipsGQqtdqs/view

See admin Member page view for related renewal code, e.g., https://ctl-staging.herokuapp.com/admin/members/399.

Also see #173 for related functionality.

Note that this story is related to a number of other stories (#307, #305, #306, #346) that should probably be launched at the same time. Therefore, may want to disable/hide the link in Production until the other stories are merged and tested.

For reference, A tools renewal: #151, #344

pjsier:
I'm interested in taking this!

jim:
This is done!

PR #454, opened 2021-02-06
Set up initial renewal requests

pjsier:
# What it does

This sets up the initial model and controller for members to request renewals for B and C items for #123. Adds `member_renewal_requestable?` to the `Loan` model to check whether a loan is eligible for a renewal request.

I think it could be more straightforward to add some of the other dependent issues onto this PR (or PRs against this branch) rather than only enable this on production, and I'm happy to keep working on the next few if that's alright. Let me know what you think and if I should make any changes, thanks!

# Why it is important

Addresses everything mentioned in #123 and starts the process for unblocking the related issues.

# UI Change Screenshot

request-renewal

# Implementation notes

* I'm not using pundit for the access checks based on the conversations on the last call
* I tried to use `any?` and sorting/filtering on loaded objects rather than `empty?` to minimize the amount of database calls if we're eager loading the renewal requests
* I didn't include the member on the `RenewalRequest` model because it didn't seem like there were any scenarios where the requester would be anyone other than the member attached to the loan, but if I missed something there I'm happy to add it
* I noticed that the current `member_renewable?` method isn't checking whether there are any holds. Is that intentional because there aren't many scenarios where they'd be needed?
* I added some `TODO` comments and method stubs for functionality like the mailer where it seemed easy enough to add a placeholder


app/models/loan.rb
```diff
@@ -84,6 +85,10 @@ def member_renewable?
     renewable? && within_borrow_policy_duration? && item.borrow_policy.member_renewable? && ended_at.nil?
   end
 
+  def member_renewal_requestable?
+    renewable? && within_borrow_policy_duration? && ended_at.nil? && !item.active_holds.any? && renewal_request.nil?
```
jim, line 89: One question about the last bit of this boolean expression- does `renewal_request.nil?` mean that a person can request another renewal if they have one that has already been approved? I think we want there to only be a single renewal request per loan that is either approved or rejected. I haven't thought a lot about this scenario, though, so let me know if there are angles that I'm missing here.

I am also feeling a need to refactor `renewal?` ,`member_renewable?`, and the new `member_renewal_requestable?` a bit, but that can be something we do in a future piece of work.

app/models/loan.rb
```diff
@@ -136,6 +141,12 @@ def checked_out?
     ended_at.blank?
   end
 
+  def renewal_request
```
jim, line 144: I think the name of this method could be a little more specific. `unapproved_renewal_request` or something that is a little more explicit would make the intent a bit clearer.

jim (changes requested):
Overall, this looks good! I tested it out and things work as expected.

I left some remarks below to hopefully add some context for some of the project's less obvious conventions.

I also realized that we'll need to have a way to hide this functionality until it's ready to ship to everyone (we don't want to let members request a renewal until the staff have a way to respond to it.)

We can use an environment variable as a simple feature flag. if the environment variable `FEATURE_RENEWAL_REQUESTS` is set to "on", we'll show the new UI you're building here, and if it's not, we'll hide the UI and the new routes.

If you add the environment variable to `.env.development`, you should be able to get that to work. I'm happy to help you sort that out, so LMK if you want extra eyes or hands and I can push that stuff up to this branch.

app/controllers/account/renewal_requests_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      flash_message = if @renewal_request.save
+        {success: "Renewal requested. We'll e-mail you an update."}
+      else
+        {error: "Something went wrong!"}
+      end
+
+      redirect_to account_home_path, flash: flash_message
+    end
+
+    def update
```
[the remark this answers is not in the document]
pjsier: Thanks for clarifying! Sure I can split this off

app/models/loan.rb
pjsier: Yeah, that would be the case, but I think a single renewal request might be more intuitive if that makes sense on your end. Agreed on probably clarifying the renewal methods as well, but for now I can clarify the intent on the other method you mentioned and also clean this up so that it only allows one renewal request

pjsier:
I made some updates here! Since the feature gating is temporary I just threw in some extra conditionals, but let me know if there's a cleaner approach.

In order for the tests to pass I also re-patched the flaky test from #455 in a separate commit based on my comment there, but I can pull that out to keep things more clean if that's preferable.

jim (approved):

jim:
@pjsier Thanks for those revisions. This looks great.

PR #455, opened 2021-02-07
Fix flaky test with freeze_time

pjsier:
# What it does

Cleans up the flaky test from #449 with the `freeze_time` helper

# Why it is important

It looks like the controller test fixed here has been failing on some PRs, so this should clean that up

### DIFF
--- a/test/controllers/signup/payments_controller_test.rb
+++ b/test/controllers/signup/payments_controller_test.rb
@@ -19,22 +19,24 @@ module Signup
       mock_result.expect :value, "https://squareup.com/checkout/12345"
 
       mock_checkout = Minitest::Mock.new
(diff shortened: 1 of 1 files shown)


jim (approved):
Thanks for proactively addressing this issue!

pjsier:
Ah I just saw this pop up again and I think the cause was different than what I initially thought. It looks like `Time.use_zone` is getting called in the application controller, but there isn't any place where the time zone is set in tests, so it's defaulting to UTC and fails at any time where the UTC date differs from the Chicago date.

The quick fix should be changing `freeze_time` to `Time.use_zone "America/Chicago"` in this same block, but it could also get set more broadly in the test configuration. Leaving this here first to make sure I'm on the right track

Issue #307, opened 2020-10-13
Librarian/staff can view renewal requests for B&C tools


rkim00:
This story probably comes after #123 .

As a Librarian, I want to be able to view renewal requests submitted by members for B&C tools (#123 ) , so that I can approve (#306 ) or deny (#305 ) the request.

Display renewal requests on a new page for Librarians. See mock-ups (https://drive.google.com/file/d/1sbf62tG2x7nUCHX5DN8_neipsGQqtdqs/view?usp=sharing , bottom of page).

Librarians should be able to toggle between all, pending ("requested" on mock-up), and closed renewal requests.

Not pictured in mock-up:
- name of tool should link to the Librarian detail on that tool (e.g., https://circulate-staging.herokuapp.com/admin/items/1167)
- name of member should link to Member detail on that Member (e.g., https://circulate-staging.herokuapp.com/admin/members/585)
- would be nice to have a column next to Name called "Number" that shows the inventory number of the item (e.g., B-1234). However, this isn't necessarily for the first version of this story.

Note that "holds" below the item Name should be the number of holds on that Item, with nothing displaying if there are no holds. This could also be a separate column if that makes implementation easier.

pjsier:
I'm interested in taking this one on as a follow up to my last PR!

jim:
Closed by #456

PR #456, opened 2021-02-10
Allow admins to view renewal requests

pjsier:
# What it does

Adds a page for admins to view renewal requests and apply filters based on status. Also fixes an env var typo in the user loans template and re-runs `standardrb` since it looks like my lefthook config wasn't working earlier.

# Why it is important

Implements the changes from the designs in #307

# UI Change Screenshot

renewal-requests-view

# Implementation notes

Everything was based on some existing admin views, so it shouldn't be too different, but let me know if my call to `includes` isn't loading all the models we need. I didn't see any duplicate queries locally, but want to be sure since I'm still getting a hold of the schema

### DIFF
--- a/app/controllers/admin/renewal_requests_controller.rb
+++ b/app/controllers/admin/renewal_requests_controller.rb
@@ -1,5 +1,14 @@
 module Admin
   class RenewalRequestsController < BaseController
+    include Pagy::Backend
+
+    def index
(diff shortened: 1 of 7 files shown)


jim (approved):
This is great! I ran through things locally and it looks like the eager loading on the requests page is working just fine.

PR #463, opened 2021-03-04
Rework account nav

jim:
# What it does

This PR simplifies the account navigation by moving it out of a menu and to the top of the screen. We've gotten some feedback about members having difficulty navigating the member portal and this is a first attempt to improve their experience.

image


rkim00:
Hey @jim , we can always change this later, but I wonder if the nav names in the screen shot might be a little confusing? Especially "Loans & Holds" vs. "Account," I find it a little confusing where to view the status of your checked-out items. I also think it might be confusing to have search on the right, which is a function for the inventory, but "View Inventory" all the way on the left, with personalized account navigation in the middle.

Let's maybe discuss at the next office hours—I imagine you'll have had feedback on the nav by then anyway, so we'll be operating from some evidence rather than my hunches.

PR #472, opened 2021-03-20
Send staff daily email of renewal requests

### LINKED ISSUE #346: Librarian/staff receives a daily e-mail summary of renewal requests
As a librarian, I want to receive a daily e-mail summary of renewal requests so that I can keep track of how many requests are coming in.

This can be sent at 8pm local time to team@chicagotoollibrary.org .
(We're going to have a local time zone parameter at some point per library... we don't have this now)

See sample text at:
https://docs.google.com/document/d/1z2vgVix4W6dQRLxg8Y9Zx1XFGOtYkECnsxzpNp70Bd4/edit

pjsier:
# What it does

Adds a task for sending a daily summary of renewal requests to library staff

# Why it is important

Closes #346

# UI Change Screenshot

renewal_requests_email

# Implementation notes

* The email goes to both staff and admins, but should this just go to staff?
* I reused `MemberMailer` instead of adding a separate mailer because all of the boilerplate seemed the same and it didn't seem worth splitting out for now
* I based the query for determining whether a request was made on the current day on the `LoanSummary` `active_on` scope, but it's not clear to me why that works with UTC instead of the local time
* I'm filtering on the user role database enum directly because I didn't see any case where we filter for multiple roles, not sure if a scope would make more sense

### DIFF
--- a/app/lib/activity_notifier.rb
+++ b/app/lib/activity_notifier.rb
@@ -37,7 +37,7 @@ class ActivityNotifier
 
   def send_staff_daily_renewal_requests
     daily_renewal_requests = RenewalRequest.requested.where.not(loan_id: nil).where("created_at >= ?", @now.beginning_of_day.utc).includes(loan: [:item, :member])
-    Member.joins(:user).where(users: {role: [:staff, :admin]}).each do |staff|
+    Member.joins(:user).where(users: {role: :admin}).each do |staff|
       MemberMailer.with(member: staff, renewal_requests: daily_renewal_requests).staff_daily_renewal_requests.deliver
     end
   end
--- a/app/lib/activity_notifier.rb
+++ b/app/lib/activity_notifier.rb
@@ -35,6 +35,13 @@ class ActivityNotifier
     end
   end
 
+  def send_staff_daily_renewal_requests
+    daily_renewal_requests = RenewalRequest.requested.where.not(loan_id: nil).where("created_at >= ?", @now.beginning_of_day.utc).includes(loan: [:item, :member])
+    Member.joins(:user).where(users: {role: [:staff, :admin]}).each do |staff|
+      MemberMailer.with(member: staff, renewal_requests: daily_renewal_requests).staff_daily_renewal_requests.deliver
+    end
+  end
+
   private
 
   def each_member(ids, &block)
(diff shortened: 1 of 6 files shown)


jim (approved):
This looks good!

* I'm going to check with Tessa about who to send the emails to.
* Your question about the timezone stuff with the Loans is something I need to dig into to wrap my head around. Please stand by 😄

jim:
@pjsier OK so the calls to `#utc` in `LoanSummary.active_on` don't actually do anything. They change the presentation of the time to be in UTC, but the underlying value isn't modified. And since everything in the database is in UTC anyway, all times get presented in that timezone when generating SQL:

```ruby
# This normally happens in an around_filter in ApplicationController
irb(main):007:0> Time.zone = "America/Chicago"
=> "America/Chicago"

irb(main):008:0> date = Time.current
=> Tue, 23 Mar 2021 20:18:41.143112300 CDT -05:00

# The same time value, just shown in a different timezone
irb(main):009:0> date.utc
=> 2021-03-24 01:18:41.1431123 UTC

# Still the same time back in central
irb(main):010:0> date.utc.in_time_zone("America/Chicago")
=> Tue, 23 Mar 2021 20:18:41.143112300 CDT -05:00
```

You can see how the values that end up getting sent to the database are the same whether we call `#utc` or not:

```ruby
irb(main):011:0> LoanSummary.where("loan_summaries.created_at BETWEEN ? AND ?", date.beginning_of_day.utc, date.end_of_day.utc).to_sql
=> "SELECT "loan_summaries".* FROM "loan_summaries" WHERE (loan_summaries.created_at BETWEEN '2021-03-23 05:00:00' AND '2021-03-24 04:59:59.999999')"

irb(main):011:0> LoanSummary.where("loan_summaries.created_at BETWEEN ? AND ?", date.beginning_of_day, date.end_of_day).to_sql
=> "SELECT "loan_summaries".* FROM "loan_summaries" WHERE (loan_summaries.created_at BETWEEN '2021-03-23 05:00:00' AND '2021-03-24 04:59:59.999999')"
```

In terms of who to send emails to, we want it to be only admins right now. Filtering on the enum directly is OK. We'll be building out some more complex role and permissions stuff eventually, but it's good to keep things simple for now.

pjsier:
Sounds good, and thanks for clarifying! I was getting tripped up on whether or not `beginning_of_day` in combination with UTC was pulling the UTC start time instead of in local time, but that makes sense

PR #478, opened 2021-03-20
Allow staff to modify appointment times

### LINKED ISSUE #237: Librarian Appointment Page - Edit Date and Time
As an librarian I should be able to edit the date and time of a scheduled appointment.

On pages like https://circulate-staging.herokuapp.com/admin/appointments/5 , should be able to edit the date and/or time.

Best way to do this is probably to add a small "edit" button/link next to the date. 

Easiest thing to do after is probably to take the user to a separate edit screen, with a pulldown to select a new date, and "confirm" and "cancel" buttons.

Alternately, we could do something like: clicking on the button would replace the date with a pull-down that fills in the existing date & time with other date & time selections, and that has "confirm" and "cancel" buttons underneath it.

In either case
- confirm would change the date & time and update the database
- cancel would take the user back to the original screen with no changes

See pages like https://circulate-staging.herokuapp.com/account/appointments/new for implementation of date & time pull-down.

Spun off from #207 . Related to Remove Item (#260 ) and Add Item (#259 ).

pjsier:
# What it does

Adds a form for staff and admins to modify appointment times and a link from the appointment detail page to get to the form. Also adds a helper method for converting an appointment's start and end into a `time_range_string` for the select input.

# Why it is important

Closes #237

# UI Change Screenshot

edit-appointment
edit-appointment-button

# Implementation notes

* Is there a reason for including `current_appointment` as a helper here instead of something like `before_action :set_appointment`?
* I copied `load_appointment_slots` from the other appointments controller, but that could probably be abstracted somewhere. That also might be better as a `before_action`

### DIFF
--- a/app/controllers/admin/appointments_controller.rb
+++ b/app/controllers/admin/appointments_controller.rb
@@ -8,6 +8,19 @@ module Admin
     def show
     end
 
+    def edit
+      load_appointment_slots
+    end
+
+    def update
+      if current_appointment.update(appointment_params)
+        redirect_to admin_appointments_path, flash: {success: "Appointment updated."}
+      else
+        load_appointment_slots
+        render :edit
+      end
+    end
+
     def destroy
       current_appointment.destroy
       redirect_to admin_appointments_path, flash: {success: "Appointment cancelled."}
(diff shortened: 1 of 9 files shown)


jim:
> Is there a reason for including current_appointment as a helper here instead of something like before_action :set_appointment?

Not really. We've had different groups of developers help out on the project, and some favored a helper-heavy style that embraced defining helpers within the controller and doing lazy loading when the helper was called. We've been moving away from that and towards an approach that loads data in `before_action` or just within the controller method. I think that style is easier for a range of developers to work with and is more similar to what they've seen on other projects.

You don't have to change the code here if you don't want to, this is a something that has been happening over time as we work through different parts of the app.

For `load_appointment_slots`, a good approach might be to move the formatting logic into a helper and just load the `Event` models in the controller. There isn't really enough logic to share in terms of loading the Events (just a couple scopes), and the helper can easily be used in multiple views as needed.

jim (approved):
There is some refactoring we can do in the future, but I think we can merge this in as-is.

pjsier:
Thanks for the review! Would you want a separate issue for refactoring this? I think it should be pretty straightforward based on what you're saying

PR #479, opened 2021-03-22
Show message if member can't make an appointment

stephanieminn:
# What it does

If the member doesn't have any items on hold or to return, show a message and disable the form fields on the create appointment page.

# Why it is important

Resolves this issue: https://github.com/rubyforgood/circulate/issues/466

# UI Change Screenshot

When you can't make an appointment...
image

When you can make an appointment! 🎉 
image

# Implementation notes
I suppose I could also hide the form fields instead of disabling, whatever feels best

### DIFF
--- a/app/views/account/appointments/_form.html.erb
+++ b/app/views/account/appointments/_form.html.erb
@@ -5,6 +5,14 @@
     <% end %>
   </ul>
 <% end %>
+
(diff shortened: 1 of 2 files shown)


jim (approved):
This is great!

I think disabling the form is OK. This way folks can get a sense of what it will look like even if they haven't put anything on hold yet.

PR #500, opened 2021-03-25
Members can edit appointments

stephanieminn:
# What it does

Members can now update appointments to add/remove items, change the appointment time, or edit the comment. 🔨 🧰 

# Why it is important

Resolves this issue: https://github.com/rubyforgood/circulate/issues/458

# UI Change Screenshot

Existing appointment:
image

Editing the appointment:
image

Updated appointment!
image

# Implementation notes

I tried my best to go through all the user flows, but if others could pull down and test, that would be great!


jim (approved):
The code looks great! I will do some manual testing before merging to verify all the user flows are working properly as you suggested.

If you are interested in adding a system test in a followup PR, it would be awesome to add that 😄

jim:
@stephanieminn Just FYI I found an issue with the admin appointment editing that I fixed. I ended up moving some logic into `Appointment#time_range_string=` so that it could be easily called from multiple controllers.

stephanieminn:
@jim nice, thank you!

PR #510, opened 2021-04-01
Add system test for updating appointments

stephanieminn:
# What it does

Adds a system test for updating appointments

# Why it is important

Follow up to this PR: https://github.com/rubyforgood/circulate/pull/500


test/system/appointments_test.rb
```diff
@@ -58,10 +58,10 @@ def select_first_available_date
 
     visit account_home_url
 
-    within row_containing(@held_item1.complete_number) { assert_text "Scheduled for pick-up" }
-    within row_containing(@held_item2.complete_number) { assert_text "Ready for pickup" }
-    within row_containing(@borrowed_item1.complete_number) { assert_text "Scheduled for drop-off" }
-    within row_containing(@borrowed_item2.complete_number) { assert_text "Checked-out" }
+    within(row_containing(@held_item1.complete_number)) { assert_text "Scheduled for pick-up" }
```
stephanieminn, line 61: these assertions weren't being executed with the in-line block syntax, but this change seems to work!

jim (approved):
Amazing! Thanks for adding this coverage.

test/system/appointments_test.rb
jim: Good catch.

PR #519, opened 2021-04-17
Require member appointments be before holds expire

### LINKED ISSUE #375: Member should be restricted from making an appointment to pick-up a hold outside of expiration date
Currently we allow members to make appointments to pick-up holds on dates that are outside the date we give them to pick-up the item.

We should restrict this. I think easiest is probably to error-check on the form submit for making an appointment (https://circulate-staging.herokuapp.com/account/appointments/new). If an item is selected whose hold will expire before the appointment date, display an error message next to the item that says "This item must be picked up by [date]. Please pick an appointment date on or before [date]."

Open to other UX suggestions or implementations as well. For instance, we may also want to display an error message at the top of the page, or put an error message next to the date pulldown.

See also #195, #343 .

pjsier:
# What it does

Adds a conditional validation to `Appointment` making sure that the appointment starts before the earliest hold expiration. 

# Why it is important

Closes #375

# Implementation notes

* I was initially looking at using a validation context, but it looks like those require calling `save` or `valid?` after using `assign_attributes` which forces related models to update before validation. Using an `attr_accessor` to determine whether or not a member is updating the record seemed to be the simplest way to only apply the validation to members and not admins

### DIFF
--- a/app/controllers/account/appointments_controller.rb
+++ b/app/controllers/account/appointments_controller.rb
@@ -63,7 +63,8 @@ module Account
         holds: Hold.where(id: appointment_params[:hold_ids], member: @member),
         loans: Loan.where(id: appointment_params[:loan_ids], member: @member),
         comment: appointment_params[:comment],
-        time_range_string: appointment_params[:time_range_string]
+        time_range_string: appointment_params[:time_range_string],
+        member_updating: true
       }
 
       @appointment.update(params)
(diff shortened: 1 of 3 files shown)


jim:
@pjsier I will give this a review this weekend. Sorry about the delay, and thanks for the contribution (as always) 😄

pjsier:
No rush! Thanks for taking a look

jim (approved):
I think this approach works. Conditional validations are tricky and we don't have a pattern in the app for this yet.

PR #520, opened 2021-04-18
Clear holds when an item is made inactive

### LINKED ISSUE #475: Items put in to maintenance or retired status should clear any holds.

pjsier:
# What it does

Adds an `after_update` callback to items so that whenever an item is changed to an inactive status all of its active holds are made inactive.

# Why it is important

Closes #475 

# Implementation notes

* The check for whether the status is changed to `maintenance` or `expired` is more verbose than I expected because `saved_change_to_attribute?` doesn't seem to support multiple values in the `to:` argument. And not sure if this is the expected behavior, but `status` in the callback is a string instead of an int so I'm converting it back in `Item.statuses[status]`

### DIFF
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -67,6 +67,8 @@ class Item < ApplicationRecord
   before_validation :assign_number, on: :create
   before_validation :strip_whitespace
 
+  after_update :clear_holds_if_inactive
+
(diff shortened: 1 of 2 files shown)


jim (approved):
Looks great!

PR #551, opened 2021-05-23
Allow librarians to create appointments

### LINKED ISSUE #358: Librarian can add appointment
As a Librarian, I want to be able to add an appointment for members so that I can schedule an appointment for members who are not able to use the Member side of the app to do so.

Currently there's no way for a Librarian to create an appointment for Members. The easiest way to do this is to associate an appointment date with a Member; the Librarian can then edit the appointment to add items for pick-up or drop-off.

Add an Appointment tab to the Librarian view Member page (e.g., https://circulate-staging.herokuapp.com/admin/members/589 ).

Selecting the tab will display the associated appointment dates & times for a particular member (if any), with a box below with a form to create an appointment.

Clicking on an appointment date and time will take user to the Librarian view of the appointment (e.g., https://circulate-staging.herokuapp.com/admin/appointments/23).

Create Appointment form will associate an appointment date & time with a member, and add that appointment date & time to the list.

Note that Librarian can then add and remove items from the view appointments page.

See mockup:
Librarian Member Appointments

pjsier:
# What it does

Allows librarians to create appointments in the member section of the admin page. I tried to base this on existing functionality so let me know if there's anything that should change!

# Why it is important

Closes #358 

# UI Change Screenshot

circulate-issue-member-appts

# Implementation notes

* I based the conditional validation on #519
* I moved the `load_appointment_slots` into a concern now that we're using it in more than one place, but let me know if there's a better spot
* Adding the new "Appointments" tab causes the line to wrap, so not sure if we want to address that

### DIFF
--- a/app/controllers/account/appointments_controller.rb
+++ b/app/controllers/account/appointments_controller.rb
@@ -1,5 +1,7 @@
 module Account
   class AppointmentsController < BaseController
+    include AppointmentSlots
+
     def index
       @appointments = current_user.member.appointments.upcoming.includes(:member, :holds, :loans)
     end
@@ -73,14 +75,6 @@ module Account
     def load_holds_and_loans
       @holds = Hold.active.includes(member: {appointments: :holds}).where(member: @member)
       @loans = @member.loans.includes(:item, member: {appointments: :loans}).checked_out
-    end
-
-    def load_appointment_slots
-      events = Event.appointment_slots.upcoming
-      @appointment_slots = events.group_by { |event| event.start.to_date }.map { |date, events|
-        times = events.map { |event| [helpers.format_appointment_times(event.start, event.finish), event.start..event.finish] }
-        [date.strftime("%A, %B %-d, %Y"), times]
-      }
     end
   end
 end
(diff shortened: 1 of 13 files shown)


jim (changes requested):
Thanks for the submission! I think this approach is good. Putting the slot loading in a concern seems like a good idea. I think I like it more there than in a view helper since it's so database-centric.

I think there are a few minor presentational things we can tweak before merging this in:

Screen Shot 2021-05-24 at 10 54 28 PM

1. Reduce the width of the select box so that the appointment time that is shown ends up on the same page
2. Remove the leading zeros from the time that is displayed in the list of appointments
3. (This wasn't in the story, but would be a small improvement) Change the "N upcoming appointments" link in the member sidebar to link to the appointments tab on the member page instead of the first appointment (which is kind of a hack since there wasn't a list of appointments on a per member basis before).

pjsier:
Thanks for the feedback! Those changes should be made. I updated the `appointment_date_and_time` helper directly because it seems like that change would apply in the other places we're using it, but I can make that more specific if we don't want to do that

jim (approved):

PR #642, opened 2021-09-01
Add scheduling appointment reminder text and update admin new item form text

stephanieminn:
# What it does

Just tackling some small tickets from the board.
- Add reminder text when scheduling an appointment
- Update admin new item form field "Other names" -> "Other names and keywords"

# Why it is important

Resolves issues https://github.com/rubyforgood/circulate/issues/605, https://github.com/rubyforgood/circulate/issues/606

# UI Change Screenshot

image

image

### DIFF
--- a/app/views/admin/items/_form.html.erb
+++ b/app/views/admin/items/_form.html.erb
@@ -4,7 +4,7 @@
   <%= form.autocomplete_text_field :name, path: admin_ui_names_path(format: :json), autofocus: true,
       hint: "Generic name for the tool; e.g. Impact driver, Orchard ladder, Flathead screwdriver" %>
 
-  <%= form.text_field :other_names,
+  <%= form.text_field :other_names_and_keywords,
       hint: "Search terms that should return this item; typically slang or other names for the item" %>
 
   <%= form.tag_select :category_ids, @categories %>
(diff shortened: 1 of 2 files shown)


app/views/admin/items/_form.html.erb
```diff
@@ -4,7 +4,7 @@
   <%= form.autocomplete_text_field :name, path: admin_ui_names_path(format: :json), autofocus: true,
       hint: "Generic name for the tool; e.g. Impact driver, Orchard ladder, Flathead screwdriver" %>
 
-  <%= form.text_field :other_names,
+  <%= form.text_field :other_names_and_keywords,
```
jim, line 7: I think we need to change just the label that is displayed like this:

```suggestion
  <%= form.text_field :other_names, label: "other_names_and_keywords",
```

Otherwise, the form builder will assume that the field is called `:other_names_and_keywords` and it will fail to save when the user submits the form because the name of the field, what's allowed in `item_params` in the controller, and the database column name need to all match.

jim (changes requested):
Just need a small adjustment to the form field name change.

app/views/admin/items/_form.html.erb
stephanieminn: oo didn't know that was all lumped together in rails forms, thanks!

stephanieminn: made the change and tested the form to make sure it still worked 🙂

jim (approved):
Looks great! Thanks for the submission.

PR #645, opened 2021-09-07
Use "in_container" driver when in docker setup

lamphanqg:
# What it does

Fix issue https://github.com/rubyforgood/circulate/issues/644

When using docker environment, the driver (`headless_chrome_in_container` or `chrome_in_contaier`) is set, but the driver being used is the one passed to `driven_by` method, which is `headless_chrome` or `chrome`. Because the driver in use does not use Chrome in the `selenium_chrome` container, and there is no Chrome installed in `web` container, Rails cannot find Chrome browser to run test.

I changed the code so that the correct driver is passed into `driven_by` method.

# Why it is important

To let developers with Docker environment run `docker-compose exec web rails test`.

# Implementation notes

Because we've already used `driven_by` method, I think we don't need to setup `Capybara.javascript_driver` and `Capybara.current_driver` so I removed those lines.

I have tested running `rails test` or `docker-exec web rails test` in following environments:

* Windows with/without docker
* Mac with/without docker

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -56,17 +56,16 @@ FactoryBot::SyntaxRunner.class_eval do
 end
 
 class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
-  driver = ENV["HEADLESS"] ? :headless_chrome : :chrome
-  driven_by :selenium, using: driver, screen_size: [1400, 1800]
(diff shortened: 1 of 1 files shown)


jim (approved):
Awesome! Thank you for fixing this and doing such extensive testing in various environments.

Issue #631, opened 2021-08-26
Update application to Rails 6.1 defaults


jim:
`config/application.rb` currently has the framework defaults set to those from Rails 6.0. Updating this was a step of the 6.1 upgrade that was missed.

We need to change this value to `6.1` and make any other changes to the application to keep its existing behavior. [Here is a list of what values will change](https://guides.rubyonrails.org/configuring.html#for-6-1-defaults-from-previous-versions-below-and).

pjsier:
Should be closed by #648, but feel free to reopen if I missed anything!

PR #648, opened 2021-09-11
refactor: update to 6.1 defaults

pjsier:
# What it does

Updates the configuration to start from the 6.1 defaults, but stays pretty conservative about going against defaults if they seem to change behavior.

# Why it is important

#631

# Implementation notes

* `legacy_connection_handling` is required to be true or the tests will fail. This could be possible for production, but to be cautious I'm applying it for all objects for now
* `form_with_generates_remote_forms` is required to be true otherwise multiple tests fail
* The rest are mostly assumptions about what could cause changes, in particular on ActiveStorage and not wanting to change anything with timezones

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -19,11 +19,24 @@ Bundler.require(*Rails.groups)
 module Circulate
   class Application < Rails::Application
     # Initialize configuration defaults for originally generated Rails version.
-    config.load_defaults 6.0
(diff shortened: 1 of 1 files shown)


jim (approved):
This looks good. We can tackle some of these bigger changes (like the active storage stuff) separately since it's probably going to be a bit of work.

PR #649, opened 2021-09-12
Add member pronunciation to librarian view

lamphanqg:
# What it does

* Add pronunciation column of string type to members table
* Show pronunciation only in librarian view's show and edit pages

# Why it is important

Resolve https://github.com/rubyforgood/circulate/issues/630

# UI Change Screenshot

* Librarian view
    * Show with pronunciation
with_pronunciation
    * Show without pronunciation    
no_pronunciation
    * Edit
librarian_view_form
* Member view (no change)
    * Show
member_view
    * Edit
member_view_form
    * Sign up form
sign_up_form

# Implementation notes

* I don't see any existed integration test for show and edit member page of librarian view, so I only add a model test. Is that OK? 

### DIFF
--- a/app/views/admin/members/_form.html.erb
+++ b/app/views/admin/members/_form.html.erb
@@ -1,5 +1,5 @@
 <%= form_with(model: [:admin, member], builder: SpectreFormBuilder) do |form| %>
-  <%= render partial: 'fields', locals: { form: form, show_password_inputs: true } %>
+  <%= render partial: 'fields', locals: { form: form, show_password_inputs: true, librarian_view: true } %>
 
   <%= form.text_area :bio %>
   <%= form.select :status, options_for_select(member_status_options, member.status), prompt: true %>
--- a/app/controllers/admin/members_controller.rb
+++ b/app/controllers/admin/members_controller.rb
@@ -65,7 +65,7 @@ module Admin
 
     def member_params
       params.require(:member).permit(
-        :full_name, :preferred_name, :email, :phone_number, :postal_code,
+        :full_name, :pronunciation, :preferred_name, :email, :phone_number, :postal_code,
         :desires, :reminders_via_email, :reminders_via_text, :receive_newsletter, :volunteer_interest,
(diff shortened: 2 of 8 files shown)


app/views/admin/members/_form.html.erb
```diff
@@ -1,5 +1,5 @@
 <%= form_with(model: [:admin, member], builder: SpectreFormBuilder) do |form| %>
-  <%= render partial: 'fields', locals: { form: form, show_password_inputs: true } %>
+  <%= render partial: 'fields', locals: { form: form, show_password_inputs: true, librarian_view: true } %>
```
jim, line 2: Maybe we could rename this param to `show_pronunciation_inputs` to mirror the `show_password_inputs` param?

jim (changes requested):
This looks good! Let's try to unify the param names that we're passing to the item form.

If you wanted to add a system test to cover member editing, would be nice to have to increase our test coverage. But that's a bonus and not at all required to merge this in. 😄

app/views/admin/members/_form.html.erb
lamphanqg: I use `librarian_view` because I guess there could be more fields which only show in librarian view but not in member view. In that case we can reuse this variable in the future.
What do you think? If the possibility of that is low, I will change the name like you suggested.

lamphanqg:
It will take more of your review effort, but if it's OK, I'd like to add some system tests for member editing. I guess I will have some time for that this weekend.

app/views/admin/members/_form.html.erb
jim: That's a good point. Let's leave the name as you have it then.

jim:
@lamphanqg I'm happy to review any system tests you add to this PR!

jim (approved):
This looks good! I'm going to merge in a big PR that adds multitenancy to the app shortly and want to get this in before that.

jim:
Also, I'm not sure why Circle doesn't report the build status for this PR.

PR #679, opened 2021-09-18
Fix system test for custom drivers

lamphanqg:
# What it does

This fixes the way we use registered drivers with `driven_by`, so that system test can run with those custom defined drivers.

# Why it is important

Solve https://github.com/rubyforgood/circulate/issues/660

# Implementation notes

About my approach here:
I checked the signature of `driven_by` and found that we need to pass the driver to the first argument, and the browser to `using:`. In our case here, when we call `Capybara.register_driver`, we define a driver, not a browser, so that `driver` should be passed as the first argument.
https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/system_test_case.rb#L154-L158

The default value for `using:` is `:chrome`, so I also checked if there is any problem when we don't pass `using:` into `driven_by`. It seems like when we pass a custom driver, value of `using:` is ignored, so it will not cause any issue.
https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/system_testing/driver.rb#L20-L26

For `headless_chrome` driver, I couldn't run it in headless mode, so I did some research. Seems like the driver API has been changed, so we must use `goo:chromeOptions` instead of `chromeOptions`, or change to use `Chrome::Options` class instead of `Remote::Capabilities.chrome`. Because we use `browser: :chrome` here, as opposed to `browser: :remote`, I decided to use `Chrome::Options` class.
https://bugs.chromium.org/p/chromium/issues/detail?id=971227#c17

I tested `bin/rails test:system` for headless and no-headless modes, with and without Docker.

By the way, there are some messages about deprecation and 422 error, but I guess they are not related to this fix.
```
$ bin/rails test:system
Run options: --seed 20613

# Running:

Capybara starting Puma...
* Version 5.3.1 , codename: Sweetnighter
* Min threads: 0, max threads: 4
* Listening on http://0.0.0.0:4000
DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated: admin/appointments/_appointment.html.erb (called from render_to_portal at /usr/src/app/app/controllers/concerns/portal_rendering.rb:7)
.................S...........JS console (severe): http://example.com:4000/admin/items/703/notes - Failed to load resource: the server responded with a status of 422 (Unprocessable Entity)
..SS.................................SS..........
[Minitest::CI] Generating test report in JUnit XML format...


Finished in 61.926151s, 1.2596 runs/s, 6.3786 assertions/s.
78 runs, 395 assertions, 0 failures, 0 errors, 5 skips

You have skipped tests. Run with --verbose for details.
```

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -23,7 +23,7 @@ end
 
 Capybara.register_driver(:headless_chrome) do |app|
   capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
-    chromeOptions: {args: %w[headless disable-gpu no-sandbox disable-dev-shm-usage]}
+    "goog:chromeOptions": {args: %w[headless disable-gpu no-sandbox disable-dev-shm-usage window-size=1400x1800]}
   )
 
   Capybara::Selenium::Driver.new(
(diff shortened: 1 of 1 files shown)


jim (approved):
I tested this and it also worked locally under WSL2 running Ubuntu on Windows with `HEADLESS=true`.

Thanks for digging in here and sorting out all this complexity! 😄

Issue #627, opened 2021-08-26
Fix WeGlot turning off after one click


rkim00:
WeGlot translations are turning off after one click, and language selection in lower-right is also turning off. Fix this so that we can determine whether or not this is something that is "good enough" translation for now.

PR #718, opened 2021-09-23
re-initialize weglot on turbolinks load

pjsier:
# What it does

Based on the conversation I'm re-initializing WeGlot on every page load. I haven't dug too deeply into the WeGlot docs, so I'm not sure if there's another function I should be calling here. From some local testing it looks like `turbolinks:load` is fired on initial page load as well so I'm always calling it as an event handler

# Why it is important

#627


jim (approved):
I looked at the docs and the only other thing I could find was a way to say that parts of the page are dynamic. Since it's the entire page, this seems like as good a solution as any.

jim (approved):

Issue #628, opened 2021-08-26
Delete item input controller so people can paste in item numbers.


jim:
On librarian view, on member pages, when lending or holding an item (e.g., https://circulate-staging.herokuapp.com/admin/members/1545#loan_12665), the box to input numbers only allows certain keys to be pressed.

This was intended to be helpful, but ultimately it has caused quite a few issues (for example, text can't be pasted into the field).

We should remove the special behavior from the app and allow any text to be entered into this input. The file at `https://github.com/rubyforgood/circulate/blob/main/app/javascript/controllers/item_number_input_controller.js` can be deleted, along with any markup that attached that controller to the form.

After this is done, we should do the following on submit:
* strip to a number
* look up the result
* return result if exists, else return existing no results message

Check these cases
* User can look items up by their complete number (A-165)
* User can look up Items using only the numeric portion of the item number (165)

![Screenshot 2021-08-31 190415](https://user-images.githubusercontent.com/3331/131591598-d67b6a7b-3764-47eb-a8cb-6efaa814d9db.png)

jim:
Closed by #732

PR #732, opened 2021-09-24
628 Field Controls for Item Number

RachelWyatt:
# What it does

- Removes the JavaScript field validation so a user can copy/paste an item number into the field controls
- Implements some Rails handling in an attempt to smooth out the user experience

# Why it is important

See Issue #628 

# UI Change Screenshot

Previous UI:

Screen Shot 2021-09-24 at 3 31 55 PM

See below for updated error behavior:

Screen Shot 2021-09-24 at 3 12 08 PM

# Implementation notes

Looking for feedback on the user flow from a UI perspective. Regarding invalid input, do we want to
a. Warn the user with an error message,
b. Physically not allow the user to enter characters other that 1-9 at a length of 1-4,
c. Magically strip away everything that isn't a number with no error message, or
d. Some other implementation I haven't thought of?

Right now it is a little of A & C. I lean away from B because I think it can be frustrating for a user.

- [ X ] I have the time and interest to make additional changes to this PR based on feedback.


jim (approved):
I think this is a good approach. It is very accepting of input from the user (which hopefully means they don't find it frustrating like they are with the current approach), while having a simple implementation that's easy to reason about.

PR #733, opened 2021-09-24
extend wait on two flaky tests Fixes #729

### LINKED ISSUE #729: Failing system tests on Main branch
**Describe the bug**
There are two failing system tests on main branch and a deprecation warning when running the system tests.

**To Reproduce**
run `rails test:system`

**Expected behavior**
Expecting all tests to pass with failures and with no deprecation warnings.

jimnanney:
# What it does

There were two tests that would fail occasionally. This PR extends the timeout on the assert-selector in those two tests to allow more time for the background job to complete.

# Implementation notes

### DIFF
--- a/test/system/user_signup_test.rb
+++ b/test/system/user_signup_test.rb
@@ -46,7 +46,7 @@ class UserSignupTest < ApplicationSystemTestCase
     perform_enqueued_jobs do
       click_on "Complete in Person"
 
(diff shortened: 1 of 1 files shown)


jim (approved):
Works for me! I know having waits in here isn't super satisfying, but they were already there and if extending them makes the tests more reliable, that's the right move.

PR #736, opened 2021-09-24
Add method to swallow js errors when they are intentional

jimnanney:
# What it does

This adds a method that clears the js console errors and prevents them from littering the test output when they are caused intentionally.

# Why it is important

Any noise in the test output has the potential to distract a developer from seeing actual errors versus the extraneous output. In cases where you will cause a console error intentionally, you should be able to prevent that from showing in the test output.

# Implementation notes

There are two ways you can use the `ignore_js_errors` method in your tests: inline and block form.

In the block form, surround the code you know will cause the errors.

```ruby
test "it gives an error when you submit without filling in the form" do
  ignore_js_errors(reason: "Creating without filling in the fields gives a 422") do
    visit edit_item_url(@item)
    click_on "Submit"
    expect(page).to have_text("field needs to be filled out")
  end
end
```

The same test can be written the inline form of `ignore_js_errors` by calling it at the end of the test.

```ruby
test "it gives an error when you submit without filling in the form" do
  visit edit_item_url(@item)
  click_on "Submit"
  expect(page).to have_text("field needs to be filled out")
  ignore_js_errors(reason: "Creating without filling in the fields gives a 422")
end
```

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -104,6 +104,12 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   private
 
+  def ignore_js_errors(reason: "I know what I am doing")
+    Rails.logger.info("Swallowed JS error because: #{reason}")
+    yield if block_given?
+    page.driver.browser.manage.logs.get(:browser)
+  end
+
   def sign_in_as_admin
     @user = FactoryBot.create(:user, role: "admin")
     login_as(@user, scope: :user)
--- a/test/system/admin/notes_test.rb
+++ b/test/system/admin/notes_test.rb
@@ -6,18 +6,20 @@ class NotesTest < ApplicationSystemTestCase
   end
 
   test "adding a note to an item" do
-    @item = create(:item)
-    visit admin_item_url(@item)
+    ignore_js_errors(reason: "This intentionally causes a 422 error") do
+      @item = create(:item)
+      visit admin_item_url(@item)
 
-    click_on "Add a Note"
-    click_on "Create Note"
+      click_on "Add a Note"
+      click_on "Create Note"
 
-    assert_content "can't be blank"
+      assert_content "can't be blank"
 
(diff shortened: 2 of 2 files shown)


jimnanney:
I tagged you Jim because this is an opinionated PR and not a necessary one.

test/system/admin/notes_test.rb
```diff
@@ -6,18 +6,20 @@ class NotesTest < ApplicationSystemTestCase
   end
 
   test "adding a note to an item" do
-    @item = create(:item)
-    visit admin_item_url(@item)
+    swallow_js_errors(reason: "This intentionally causes a 422 error") do
```
jim, line 9: This is a really nice API to handle cases where we expect JS “errors” like 4xx responses.

test/application_system_test_case.rb
```diff
@@ -104,6 +104,12 @@ def fail_on_js_error(error)
 
   private
 
+  def swallow_js_errors(reason: "I know what I am doing")
```
jim, line 107: I like this helper method. My only thought is that while “swallow” is a fairly colloquial way to refer to this kind of behavior, “ignore” is a little more clear and may be more likely to be understood by a wider range of developers.

```suggestion
  def ignore_js_errors(reason: "I know what I am doing")
```

jim (changes requested):
I think we can improve the naming, but overall this is a good solution to a problem that’s been hanging around for far too long.

test/system/admin/notes_test.rb
jimnanney: I updated the name as suggested. Good call, I like that better.

test/application_system_test_case.rb
jimnanney: I updated the commit and PR to have the better naming. Good call on that!

jimnanney:
Updated naming as suggested. Great call on using ignore instead of swallow.

jimnanney:
I also updated the git log to include the notes from the above pull request.

jim (approved):

PR #741, opened 2021-09-25
allow item description searchess

### LINKED ISSUE #683: Include description field when performing inventory search
Making the search bar return results that found the relevant words in the item's description would make it easier for people to find the items they are looking for even when they might be less obvious.

That would also make it less difficult for librarians to add items to the inventory in a way that would be likely to return for all the relevant search terms.

This is all part of a generalized effort to make the inventory more user friendly. 

**Implementation notes:**

We're using [pg_search](https://github.com/Casecommons/pg_search) to provide searching across items. Since `description` is implemented using [ActionText](https://edgeguides.rubyonrails.org/action_text_overview.html), we will need to look at [configuring pg_search to search through that association](https://github.com/Casecommons/pg_search#searching-through-associations).

h-m-m:
# What it does
This works by adding a before save hook that persists the plain text version of the rich text description to the database so it can be used with pg_search

# Why it is important

Resolves #683

# UI Change Screenshot

n/a

# Implementation notes

Other ways of implementing this that I passed on because they required more work
* keeping with the pg_search gem but swapping from the search scope to the pg_search multisearch feature (we may want to do eventually if we add search for anything other than items)
* trying to get pg_search to use the postgres extensions of `xpath('string(/div)', xmlparse(content action_text_rich_texts.body))` but I couldn't get the joins to work right without getting into the guts of pg_search

Possible future tasks
* add a rake task to make sure plain text descriptions are updated for records with description text created before this code's deploy


config/environments/development.rb
```diff
@@ -63,6 +63,7 @@
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
   config.hosts << ".circulate.local"
+  config.hosts << "chicago-circulate.pdev"
```
jim, line 66: Did you run into any issues with the subdomain stuff? It’s pretty fresh and I’m sure there are sharp edges still.

jim (commented):
This is a very pragmatic way to handle this issue. We will need a plan for backfilling the values in the new column, but that can be a follow up PR since there’s no harm to merging this first.

Issue #409, opened 2020-11-22
Send appointment confirmation email


jim:
**Is your feature request related to a problem? Please describe.**

Members like being able to see their appointment information in their email. Currently they don't receive any e-mails when they make an appointment.

**Describe the solution you'd like**

We should send a confirmation email to users when they schedule an appointment. It should look like the other emails we send (e.g., the loan summary template) and list the tools that will be picked up or dropped off.

Check with @jim if you've got any questions on this!

jim:
Closed by #742

PR #742, opened 2021-09-25
Add Appointment Confirmation Email

RachelWyatt:
- Send an email when appointment is created or updated in appointments_controller.rb
- Add appointment_confirmation email type in member_mailer.rb
- Rough draft of text for email
- Add email type to mailer preview

# What it does

First cut at implementation of an appointment confirmation email! Next steps as I imagine them...
- Input on what content email should have
- Styling for HTML emails
- Write tests
- Run tests and confirm email sends as intended

# Why it is important

See Issue #409

# UI Change Screenshot

Plain text of current email:

Screen Shot 2021-09-25 at 3 40 01 PM

# Implementation notes

There currently is only a plain text email, looking to get confirmation of content before I start any styling.

The email will list holds as well as loans if there are any holds on the appointment.

- [ X ] I am interested in feedback but don't need to make the changes myself.


rkim00:
Hi @RachelWyatt , thanks for your work! I'm not too concerned about text—we can easily change that later. But UTC is almost certainly going to be a confusing time display for our (for now) mostly Chicago-area audience, some of whom are not very technologically sophisticated.

Preferred time zone here is CT, switching with Daylight Savings Time. There *may* be a setting you can access that determines the local time on a per-library basis; if not, hard-coding CT is probably fine for now. @jim , do you know any more about time zones?

Also, strongly suggest we show day of the week as well, to reduce confusion.

rkim00:
Oh, also note that two different things can happen at an appointment: (1) picking up items on hold, and (2) returning items that are due, both of which are part of an appointment. So I'd indicate both of these separately. Something like:

You are scheduled to return these items—please make sure to bring!
- (item)
- (item)
You are scheduled to pick-up these items you have on hold:
- (item)
- (item)

jim:
Overall this is looking good! In my mind there are two things to change/add:

1. Regarding time display, we can probably us the same format that we use in [the account UI](https://github.com/rubyforgood/circulate/blob/main/app/views/account/appointments/index.html.erb), which should include day of the week in addition to date and then time.

2. I think we should add a sentence to the end that says:

    > You can make changes to this appointment by clicking here.

    Where “clicking here” links to that appointment’s page in the account UI.

RachelWyatt:
Hi @jim, my availability is pretty low right now, so if you can pass if off to someone else please feel free!

jim:
@pjsier Can you give this code a quick look? I think it's OK but would love another set of eyes before I merge it.

I ran standard so there are a few changes that are just from that adjusting syntax and such.

pjsier (approved):
This all looks good to me! The preview looks great

Issue #747, opened 2021-09-27
Setup GitHub Actions for CI


jim:
**Is your feature request related to a problem? Please describe.**

We occasionally have issues where builds from external contributors aren't run. I think it may be due to the contributors having their own Circle accounts, but it's unclear and as a maintainer it doesn't seem that I have the ability to fix this.

**Describe the solution you'd like**

I'd like to setup a GitHub Action to run the project's tests, and configure it to run against all PRs. It should run without any access to secrets, and therefore there should be no security risks in doing so.

**Describe alternatives you've considered**
* Figure out why CircleCI doesn't always run/report some builds (spent some time on this to no avail)
* Use a third CI service (doesn't have the benefit of consolidating our tooling since we're pretty invested in GitHub already)

pjsier:
I'm interested in taking this on!

PR #754, opened 2021-10-01
ci: add github actions for ci

pjsier:
# What it does

Adds a GitHub Actions workflow for running tests. We could also add linting in here if you want to double check in addition to the pre-commit hooks.

# Why it is important

#747 

# Implementation notes

* Not 100% sure this will work immediately, but the overall setup should be right.
* We don't necessarily need the artifacts piece at the end, but I was trying to match CircleCI as much as possible
* I didn't install imagemagick or other libraries here, so it'll be worth seeing which ones are potentially pre-installed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,93 @@
+name: CI
+on:
+  pull_request:
+    types: [opened, synchronize]
+  workflow_dispatch:
+
+env:
+  RAILS_ENV: test
+  CI: "true"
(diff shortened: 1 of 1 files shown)


jim (approved):
This is great. I agree we can expand the config once the basics are working.

I'm going to try to get these to start running on this branch so we can see the results before merging this into main.

pjsier:
@jim looks like the last update might have fixed it! Weird since we're still getting the same font error but installing the fonts seems to have helped

PR #755, opened 2021-10-03
fix: use attribute to find day in shift system test

jim:
# What it does

This fixes a system test that was very sensitive to the current day. It replaces naive text matching with a `data-test-date` attribute, which should insulate this test from timezone and other temporal shenanigans in the future.

# Why it is important

We have a broken test on the main branch.


pjsier (approved):
Makes sense!

PR #788, opened 2021-10-11
chore: Add run-scripts GitHub actions workflow job

### LINKED ISSUE #786: Run bin/setup and bin/reset in CI
We have two scripts that are important for contributors: `bin/setup`, which performs the initial setup of the project, and `bin/reset`, which resets the database to a known good state. 

Ideally, we can run these scripts in a CI job so that we find out as soon as possible if one of them starts failing to run.
We're migrating to GitHub Actions for CI, so we only need to update that configuration. The Circle CI config will be deleted in the nearish future.

**Additional context**
For the purposes of this issue, the scripts running without error is considered success. Whether they do all the right things can be considered to be separate work.

laasem:
# What it does

Runs the `bin/setup` and `bin/reset` scripts in the GitHub actions CI as a new job in the on pull request workflow.

# Why it is important

Closes #786.

# UI Change Screenshot

N/A

# Implementation notes

There is some duplication across steps, as well as across the `test` and `run-scripts` jobs. Would it be better to find a way to set these duplicate configurations at a higher level?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,7 +125,17 @@ jobs:
         bundler-cache: true
 
     - name: Run bin/setup script
+      env:
+        PGHOST: localhost
+        PGUSER: postgres
+        PGPASSWORD: postgres
+        PGPORT: ${{ job.services.postgres.ports[5432] }}
       run: ./bin/setup
 
     - name: Run bin/reset script
+      env:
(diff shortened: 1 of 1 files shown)


jim:
@pjsier Can you give this a review? It looks good to me, but you're more familiar with the Actions stuff.

PR #803, opened 2021-10-14
Admin Turn Off Payments

### LINKED ISSUE #796: Admin can disable Square/Payment for a library
As an admin for a library, I want to be able to turn off online payment for my library, so that I can disable a feature I might not need.

1. create UI on admin side to toggle the use of online payment (we use Square Checkout, but that could change in the future).
  - This is used in two flows: member signup and membership renewal
  - On both pages there is another option to continue, so when online payment is turned off, we'll just need to hide the relevant parts of the page and the other option will allow users to complete the workflow.
  - It's up to the implementer how to best accomplish this (disabling routes based on the current library or adding checks in controllers).

KFad11:
# What it does

Resolves #796

As an admin for a library, I want to be able to turn off online payment for my library, so that I can disable a feature I might not need.

- Added to the existing library feature management UI to toggle off payments.
- When payments are off, the payment steps & forms are skipped for both new and renewing memberships.

# Why it is important

https://github.com/rubyforgood/circulate/issues/796

# UI Change Screenshot

Screen Shot 2021-10-14 at 10 47 32 AM

Screen Shot 2021-10-14 at 10 48 33 AM

Screen Shot 2021-10-14 at 1 07 08 PM

- [X] I have the time and interest to make additional changes to this PR based on feedback.

### DIFF
--- a/app/controllers/renewal/acceptances_controller.rb
+++ b/app/controllers/renewal/acceptances_controller.rb
@@ -4,8 +4,10 @@ module Renewal
 
     def create
       @acceptance = @member.acceptances.new(terms: acceptance_params[:terms])
-      if @acceptance.save
+      if @acceptance.save && @current_library.allow_payments?
         redirect_to new_renewal_payment_url
+      elsif @acceptance.save
+        redirect_to renewal_confirmation_url
       else
         activate_step(:agreement)
         @document = Document.agreement


KFad11:
@jim - We went through both new member sign up and renewal flows with internal QA. Everything looks good on our end. Let us know if anything needs changing. Thanks!

PR #828, opened 2021-10-25
feat: create job for async neon crm updates

pjsier:
# What it does

Creates a new job for running the Neon CRM sync in the background

# Why it is important

https://github.com/rubyforgood/circulate/pull/789/files#r735003294

# Implementation notes

* It looks like `sucker_punch` is a pretty standard implementation of ActiveJob, but I might have missed something

### DIFF
--- a/test/models/member_test.rb
+++ b/test/models/member_test.rb
@@ -175,4 +175,15 @@ class MemberTest < ActiveSupport::TestCase
 
     assert_nil member.last_membership
   end
+
+  test "schedules a job for updating Neon membership when enabled" do
+    member = create(:member)
+
+    NeonMemberJob.stub :perform_async, true do
+      member.stub :can_update_neon_crm?, true do
+        member.update(city: "#{member.city} Test")
+        assert_send([NeonMemberJob, :perform_async])
+      end
+    end
+  end
 end
--- a/test/models/member_test.rb
+++ b/test/models/member_test.rb
@@ -187,6 +187,6 @@ class MemberTest < ActiveSupport::TestCase
       end
     end
 
-    assert mock
+    assert_mock mock
   end
 end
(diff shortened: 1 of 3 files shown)


test/models/member_test.rb
```diff
@@ -175,4 +175,15 @@ class MemberTest < ActiveSupport::TestCase
...
     assert_nil member.last_membership
   end
+
+  test "schedules a job for updating Neon membership when enabled" do
+    member = create(:member)
+
+    NeonMemberJob.stub :perform_async, true do
+      member.stub :can_update_neon_crm?, true do
+        member.update(city: "#{member.city} Test")
+        assert_send([NeonMemberJob, :perform_async])
```
jim, line 185: I think that `assert_send` asserts the result of calling the passed method, but in this case what we probably want is to assert that the `perform_async` method was already called by other code (in this case, the callback within Member).

To do that, we'd need a mock instead of a stub (at least in Minitest): https://semaphoreci.com/community/tutorials/mocking-in-ruby-with-minitest#h-mocking

jim (changes requested):
The implementation here looks good and works as designed. There's a change needed to have the test assert that the callback method is actually invoked when it's expected to be.

test/models/member_test.rb
pjsier: Makes sense! I think the update I just pushed should fix it

jim (approved):

test/models/member_test.rb
```diff
@@ -175,4 +175,18 @@ class MemberTest < ActiveSupport::TestCase
...
+    mock = Minitest::Mock.new
+    mock.expect :call, true, [Integer]
+
+    NeonMemberJob.stub :perform_async, mock do
+      member.stub :can_update_neon_crm?, true do
+        member.update(city: "#{member.city} Test")
+      end
+    end
+
+    assert mock
```
jim, line 190: I think this needs to be `assert_mock`?

```suggestion
    assert_mock mock
```

jim (dismissed):

PR #846, opened 2021-11-04
Remove Loan#within_borrow_policy_duration method

jim:
This method had a reference to `Time.current` that was causing tests to break when loans straddled a daylight savings time change and was additionally really difficult to reason about. There was a comment mentioning that it could be removed, so I went ahead and did that.

It's possible that this will make renewals more open than we wish, but we can lock things back down when we identify the situation where that is relevant.

### DIFF
--- a/app/models/loan.rb
+++ b/app/models/loan.rb
@@ -92,12 +92,12 @@ class Loan < ApplicationRecord
 
   # Can a member renew this loan themselves without approval?
   def member_renewable?
-    renewable? && within_borrow_policy_duration? && item.borrow_policy.member_renewable? && ended_at.nil?
+    renewable? && item.borrow_policy.member_renewable? && ended_at.nil?
   end
 
   # Can a member request this loan be renewed?
   def member_renewal_requestable?
-    renewable? && within_borrow_policy_duration? && ended_at.nil? && !any_active_holds? && !renewal_requests.any?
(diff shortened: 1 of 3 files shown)


app/models/loan.rb
```diff
@@ -92,26 +92,21 @@ def renewable?
...
   # Can a member renew this loan themselves without approval?
   def member_renewable?
-    renewable? && within_borrow_policy_duration? && item.borrow_policy.member_renewable? && ended_at.nil?
+    renewable? && item.borrow_policy.member_renewable? && ended_at.nil?
   end
 
   # Can a member request this loan be renewed?
   def member_renewal_requestable?
-    renewable? && within_borrow_policy_duration? && ended_at.nil? && !any_active_holds? && !renewal_requests.any?
+    renewable? && ended_at.nil? && !any_active_holds? && !renewal_requests.rejected.any?
```
pjsier, line 100: Is there any reason for adding the `rejected` scope here? Wouldn't this make it possible for a member to create multiple renewal requests for a single loan at the same time?

pjsier (approved):
I had one small note but otherwise this looks good, and so I'm good with it being merged either way!

app/models/loan.rb
jim: That's a good call. I'll take out the `rejected` call here.

jim: Oh, maybe it should be `requested` so that a member can't request a renewal for something that they already have?

PR #852, opened 2021-11-07
configure Pagy to return last page on overflow

### LINKED ISSUE #850: Paginated endpoints return HTTP 500 when page param exceeds number of pages of results
**Describe the bug**

Paginated endpoints will throw an exception when the page parameter exceeds the number of pages of items to display. This exception is actively [being triggered](https://sentry.io/organizations/chicago-tool-library/issues/2580855230/) by [an enthusiastic web crawler](https://www.semrush.com/bot/). The library's user receives an HTTP 500 server error.

**To Reproduce**
1. Fire up a development instance with the devdata loaded.
2. Log in.
3. Click on "Lending" then "Items".
4. Scroll to the bottom of the page.
5. Click on the last page number available. (Currently "10")
6. After the last page loads, in the browser's location bar, change the `page=##` parameter to a number higher than the one that appears—e.g. change "10" to "11"—and press Return on the keyboard.
7. Behold! An error!

**Expected behavior**

An error of "the server can't do that" sent to the user is unexpected. The unhandled exception alerts fired are less than ideal.

Some behaviors in the realm of reasonable expectations, some of which could be combined:
* The paginated page returns empty.
* The paginated page returns the last legitimate page of results.
* The user is shown a warning on a properly rendered page of results notifying them that the page that was requested is currently invalid.

robbkidd:
# What it does

By default, Pagy throws an exception when given a page parameter greater than the number of pages available in a result. This change adds in [Pagy's Overflow extra](https://ddnexus.github.io/pagy/extras/overflow) (included in the gem already) and configures overflow to return the last page of results when page param is too large (in other words, when page overflows).

# Why it is important

Closes #850, addressing an unhandled exception when a request is received with a page parameter that's too high.

# Implementation notes

This is probably the simplest fix for quieting the unhandled exception.

While the only "user" experiencing this error is a web crawler, returning the last page of results (say, page 10) when ever a page param is given that overflows (say, page 11 or 100, or 1,000,000) implies at the protocol level of HTTP that the last page of results is the valid collection of resources at the actually-erroneous URL with the overflowing page param. This probably isn't a huge issue because the library is predominantly focused on a web UI and does not provide API-only access to things. Noting this behavior here in case of future API considerations.

- ✔️ I have the time and interest to make additional changes to this PR based on feedback.
- ~I am interested in feedback but don't need to make the changes myself.~
- ~I don't have time or interest in making additional changes to this work.~
- ~Other or not sure (please describe):~

### DIFF
--- a/config/initializers/pagy.rb
+++ b/config/initializers/pagy.rb
@@ -1,2 +1,5 @@
 require "pagy/extras/bootstrap"
 require "pagy/extras/arel"
+require "pagy/extras/overflow"
+
+Pagy::DEFAULT[:overflow] = :last_page


jim (approved):
Looks good to me. Thanks for digging in and sorting this out. We won't miss the exception alerts 😄

PR #856, opened 2021-11-10
feat: add category import rake task

jim:
# What it does

We have a data team/superhuman who has gone through all of CTL's items and made sure they are assigned to a category. This PR adds a rake task that will allow me to import those categorizations into the database.


app/models/item.rb
```diff
@@ -45,7 +45,7 @@ def next_hold
     electric_battery: "electric (battery)"
   }
 
-  audited
+  audited except: :plain_text_description
```
jim, line 48: No need to track changes to this column: it's only used to make it easier to search for items.

lib/tasks/import.rake
```diff
@@ -36,4 +36,62 @@ namespace :import do
       end
     end
   end
+
+  desc "Updates items with categories"
+  task category_csv: :environment do
+
+    class Category
+      def refresh_category_nodes
+        # no need to do this after every change
```
jim, line 45: This is a hack to prevent the materialized view from being refreshed after every categorization change. Without this, the import would be a lot slower.

PR #901, opened 2021-12-06
fix: reference to wrong instance variable was breaking renewal emails

jim:
# What it does

A bug in one of the email templates is breaking the messages we send to members when their renewal requests are responded to by staff.

# Why it is important

We'd like our emails to reach our members 😄 

# Implementation notes

* The issue only came up when an item had an attached image, and our tests didn't create items with images. This has been addressed.

### DIFF
--- a/test/mailers/previews/member_mailer_preview.rb
+++ b/test/mailers/previews/member_mailer_preview.rb
@@ -79,20 +79,24 @@ class MemberMailerPreview < ActionMailer::Preview
   end
 
   def renewal_request_approved
-    tomorrow = Time.current.end_of_day + 1.day
-    member = Member.verified.first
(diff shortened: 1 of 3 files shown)


test/mailers/previews/member_mailer_preview.rb
```diff
@@ -79,20 +79,24 @@ def staff_daily_renewal_requests
   end
 
   def renewal_request_approved
-    tomorrow = Time.current.end_of_day + 1.day
-    member = Member.verified.first
-    loan = Loan.create!(item: Item.available.order("RANDOM()").first, member: member, library: member.library, due_at: tomorrow, uniquely_numbered: false)
-    renew_loan(loan)
-    renewal_request = RenewalRequest.create!(loan: loan, status: :approved)
-    MemberMailer.with(renewal_request: renewal_request).renewal_request_updated
+    ActsAsTenant.with_tenant(Library.first) do
```
jim, line 82: These previews were breaking because the loans couldn't be renewed without a current library bring set.

PR #902, opened 2021-12-06
add 'Overdue' item status label

### LINKED ISSUE #888: When an item is overdue change tag to a red "overdue" instead of yellow "checked out"
**What?**
When an item in the item inventory pages is overdue, we'd like it to be more obvious to folks browsing the inventory.

**Why?**
This will help members avoid putting items on hold that are overdue (and unlikely to return soon or possibly at all).

**Where?**
The pages in question are the [inventory page](https://app.chicagotoollibrary.org/items) and the [item detail page](https://app.chicagotoollibrary.org/items/2736).

The main logic involved in displaying this information lives in the [item_status_label](https://github.com/rubyforgood/circulate/blob/c6b194dcc51d3d3645a1a3b502c0519d1f8bb7fc/app/helpers/items_helper.rb#L166) method. We should be able to extend that method (or perhaps [`css_class_and_status_label`](https://github.com/rubyforgood/circulate/blob/c6b194dcc51d3d3645a1a3b502c0519d1f8bb7fc/app/helpers/items_helper.rb#L150) to accomplish this change.

It would also be great to delete the [`item_status_class`](https://github.com/rubyforgood/circulate/blob/c6b194dcc51d3d3645a1a3b502c0519d1f8bb7fc/app/helpers/items_helper.rb#L171) method since it is no longer used.

fpsvogel:
# What it does

Resolves #888 

Overdue items now appear with an "Overdue" status label, instead of "Checked Out".

Also, the method `ItemsHelper#item_status_class` is deleted since it is not used anywhere.

# Why it is important

This will help members avoid putting items on hold that are overdue (and unlikely to return soon or possibly at all).

# UI Change Screenshot

circulate-#888-before

circulate-#888-after

The same change appears at the item page's "Status" label.

### DIFF
--- a/app/helpers/items_helper.rb
+++ b/app/helpers/items_helper.rb
@@ -150,7 +150,11 @@ module ItemsHelper
   def css_class_and_status_label(item)
     if item.active?
       if item.checked_out_exclusive_loan
-        ["label-warning", "Checked Out"]
+        if item.overdue?
+          ["label-error", "Overdue"]
+        else
+          ["label-warning", "Checked Out"]
+        end
       elsif item.borrow_policy.uniquely_numbered? && item.active_holds.size > 0
         ["label-warning", "On Hold"]
       else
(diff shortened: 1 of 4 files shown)


jim (approved):
This looks great! Thanks for the contribution.

PR #919, opened 2021-12-20
Move appointment loans to the renewal when a loan is renewed.

jim:
# What it does

This fixes a bug where a loan that was renewed after being added to an appointment would be shown as being checked in.

# Why it is important

This bug could create confusion and cause staff to not check in items that should be checked in.

# Implementation notes

* I did this by updating the `loan_id` association on `AppointmentLoan`. The other option was to change `AppointmentLoan` to be associated with `LoanSummary` instead of `Loan`, but since that is a more intrusive change I went with this option.


pjsier (approved):
This makes sense!

PR #947, opened 2022-01-23
Replace search controller with items#index filters

### LINKED ISSUE #784: Inventory search should filter the main inventory page

When a user searches for an item on the public inventory pages, they are taken to a search results page like [this one](https://app.chicagotoollibrary.org/search?query=key).

1. When a user searches in the header search box for `your query`, they should see their search results in the form of a filtered inventory page with only matching items appearing in the list:
  image

2. If a user clicks on a category name, that category should be added as another filter:
  image

3. We can also remove the public search controller, views, etc.

**Describe alternatives you've considered**

**Additional context**

The images above are just mockups. A nicer UI for searching and filtering would be welcomed. There is a separate story to add another filtering option to the public inventory pages: #672

shkm:
# What it does

- Removed public-facing search controller/views
- Added filtering to the items#index

# Why it is important

Closes #784

# UI Change Screenshot

image

# Implementation notes

- I added .env.test.local to gitignore since I needed to modify the database URLs in a semi-docker-based (DB in docker, Rails not) test environment
- The param order is important due to the use of `link_to_unless` (and therefore `current_page?`) for the sort buttons
- Category counts on the left side do not reflect filtering, but this was the case before so I think it's fine for this PR. It's a bit tricky to do this right if we want to, particularly with regard to the query param.


app/controllers/items_controller.rb
```diff
@@ -2,17 +2,14 @@ class ItemsController < ApplicationController
...
 
   def index
-    item_scope = Item.listed_publicly.includes(:checked_out_exclusive_loan)
+    item_scope = Item.listed_publicly.includes(:borrow_policy, :active_holds, :checked_out_exclusive_loan, :categories)
 
-    if params[:category]
-      @category = CategoryNode.where(id: params[:category]).first
-      redirect_to(items_path, error: "Category not found") && return unless @category
+    item_scope = filter_by_category(item_scope) if filter_params[:category].present?
+    item_scope = filter_by_query(item_scope) if filter_params[:query].present?
```
jim, line 8: This reads nicely and does a good job of extracting the filtering logic into other methods.

test/controllers/admin/searches_controller_test.rb
```diff
@@ -1,7 +1,7 @@
 require "test_helper"
 
 module Admin
-  class ItemsControllerTest < ActionDispatch::IntegrationTest
+  class SearchesControllerTest < ActionDispatch::IntegrationTest
```
jim, line 4: Nice catch.

jim (approved):
This is great. The code looks good, and I tested the behavior locally and it's working.

One thing I noticed was that you can get into a dead end state by first searching and then clicking on a category with no results within it. Like you mentioned, the counts on the left don't update yet and that's nontrivial to implement. In the meantime, though, if we add a link to clear the search or category filter, it will at least give users a place to go next. I'm thinking something like this:

image

I'll add those links real quick and then merge this in.

PR #959, opened 2022-01-28
Add appointment counts to activity report

### LINKED ISSUE #752: Add appointment count to Activity Report
We'd like to add a new bit of information to the Activity Report: how many appointments were scheduled per month.

Here's the current report:

image

1. We'd like there two be a new column under `Activity` called `Appointments`.
2. The value for each column should be the total number of scheduled appointments during that month. 
3. Show the number of completed appointments in parenthesis in the same cell would be great.

So if there were 10 total appointments in a month and 8 of those were completed, the cell would show "10 (8)".

shkm:
# What it does

Adds appointment counts to the activity report

# Why it is important

Closes #752

# UI Change Screenshot

image

# Implementation notes

The spec said to add the completed count in brackets, but I felt this was more consistent. I've also added a test for this table.


jim (changes requested):
Thanks for the contribution. This code looks good!

I am seeing an error when I run the migrations on this PR, though:

```
== 20220128134936 UpdateMonthlyActivitiesToVersion2: migrating ================
-- update_view(:monthly_activities, {:version=>2, :revert_to_version=>1})
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

No such file or directory @ rb_sysopen - /home/jimb/code/circulate/db/views/monthly_activities_v02.sql
/home/jimb/code/circulate/db/migrate/20220128134936_update_monthly_activities_to_version_2.rb:3:in `change'

Caused by:
Errno::ENOENT: No such file or directory @ rb_sysopen - /home/jimb/code/circulate/db/views/monthly_activities_v02.sql
/home/jimb/code/circulate/db/migrate/20220128134936_update_monthly_activities_to_version_2.rb:3:in `change'
Tasks: TOP => db:migrate
(See full trace by running task with --trace)
```

I think `db/views/monthly_activities_v02.sql` probably just got missed in the git commit. If you can add that file, I think this will be ready to go!

shkm:
> Thanks for the contribution. This code looks good!
> 
> I am seeing an error when I run the migrations on this PR, though:
> 
> ```
> == 20220128134936 UpdateMonthlyActivitiesToVersion2: migrating ================
> -- update_view(:monthly_activities, {:version=>2, :revert_to_version=>1})
> rails aborted!
> StandardError: An error has occurred, this and all later migrations canceled:
> 
> No such file or directory @ rb_sysopen - /home/jimb/code/circulate/db/views/monthly_activities_v02.sql
> /home/jimb/code/circulate/db/migrate/20220128134936_update_monthly_activities_to_version_2.rb:3:in `change'
> 
> Caused by:
> Errno::ENOENT: No such file or directory @ rb_sysopen - /home/jimb/code/circulate/db/views/monthly_activities_v02.sql
> /home/jimb/code/circulate/db/migrate/20220128134936_update_monthly_activities_to_version_2.rb:3:in `change'
> Tasks: TOP => db:migrate
> (See full trace by running task with --trace)
> ```
> 
> I think `db/views/monthly_activities_v02.sql` probably just got missed in the git commit. If you can add that file, I think this will be ready to go!


Good call; turns out it was my global gitignore. I've force pushed it up now!

jim:
@shkm Thanks for adding that file. I was able to pull down the branch and the migration ran no problem against a clone of the production database.

But when I went to visit the report in the UI locally, the page never loaded. Running just the SQL statement against the production data takes minutes to complete. I know you don't have access to that, so it's likely that this is only something that appears with a certain number of rows.

I think it's possible that the original approach taken with the query is no longer serving us and we'll need to revisit how the report is querying for data at a more fundamental level. I'm also not sure it needs to be a view now that we know that it's only really used for one page.

shkm:
> @shkm Thanks for adding that file. I was able to pull down the branch and the migration ran no problem against a clone of the production database.
> 
> But when I went to visit the report in the UI locally, the page never loaded. Running just the SQL statement against the production data takes minutes to complete. I know you don't have access to that, so it's likely that this is only something that appears with a certain number of rows.
> 
> I think it's possible that the original approach taken with the query is no longer serving us and we'll need to revisit how the report is querying for data at a more fundamental level. I'm also not sure it needs to be a view now that we know that it's only really used for one page.

I tried to speed up the existing view but gave up in the end: it might be possible but I think we'd be going down a road of more obscure SQL. As an alternative, I've dropped that view and instead created separate views; we now have 3 calls to the DB but it ends up really fast.

config/puma.rb
```diff
@@ -25,7 +25,7 @@
 # Workers do not work on JRuby or Windows (both of which do not support
 # processes).
 #
-workers ENV.fetch("WEB_CONCURRENCY", 2) unless environment_name == "development"
+workers ENV.fetch("WEB_CONCURRENCY", 2) unless environment_name.in?(%w[test development])
```
shkm, line 28: I figured this wouldn't disrupt anyone's environment. Debuggers don't like concurrency.

jim: Good catch!

jim (approved):
Thanks for working through this. Three queries that execute fast are a lot bettern than one query that takes a looooong time 😄

PR #977, opened 2022-02-13
Hide Appointment Form if there's no hold nor return.

### LINKED ISSUE #940: Hide content on new appointment page when no appointment can be created
**Describe the bug**
Members can see an appointment form that doesn't make much sense

**To Reproduce**
1. Go to the new appointment page without any loans or holds

**Expected behavior**
It would be nice to clean this page up and hide the headings and other content since the member can't interact with them.

**Screenshots**
image

ChaelCodes:
# What it does

closes #940 

Instead of showing the appointment form, we show a blurb explaining they cannot make an appointment with no holds nor returns.

# Why it is important

This makes the appointment page less cluttered for folks who can't make an appointment because they don't have a hold nor an item to return.

# UI Change Screenshot

Before | After
--- | ---
blurb telling user they cannot make an appointment, and then a form to make an appointment. | form is gone on schedule appointment page

# Implementation notes

* I moved the form into an if/else instead of it living outside the if. I handled it inside _form, so if this form is used anywhere else, the change will also apply.

_My weekday schedule is a little full, but every weekend, I have time to contribute._

### DIFF
--- a/app/views/account/appointments/_form.html.erb
+++ b/app/views/account/appointments/_form.html.erb
@@ -8,112 +8,111 @@
     </ul>
   <% end %>
 
-  <% if !can_schedule_appointment %>
-    <div class="toast toast-warning">
-      <p>You can only schedule an appointment if you have a tool on hold or that can be returned.</p>
-      <%= link_to "View Inventory", items_path, class: "btn btn-lg btn-default" %>
-    </div>
-  <% end %>
+  <% if can_schedule_appointment %>
+    <%= form_with(model: [:account, @appointment]) do |form| %>
+      <div class="columns">
 
-  <%= form_with(model: [:account, @appointment]) do |form| %>
+        <% if loans.any? %>
+          <div class="column col-5 col-sm-12 col-mr-auto">
+            <h2 class="h3">Select Items to Return</h2>
 
-    <div class="columns">
+            <ul class="loan-list">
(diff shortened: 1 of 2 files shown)


test/system/appointments_test.rb
```diff
@@ -77,6 +77,20 @@ def fill_in_optional_field(text)
     within(list_item_containing(@held_item2.complete_number)) { assert_text "Ready for pickup" }
   end
 
+  test "attempts to schedule an appointment with no holds" do
```
jim, line 80: Thanks for adding a test for this.

app/views/account/appointments/_form.html.erb
```diff
@@ -8,112 +8,111 @@
     </ul>
   <% end %>
 
-  <% if !can_schedule_appointment %>
-    <div class="toast toast-warning">
-      <p>You can only schedule an appointment if you have a tool on hold or that can be returned.</p>
-      <%= link_to "View Inventory", items_path, class: "btn btn-lg btn-default" %>
-    </div>
-  <% end %>
+  <% if can_schedule_appointment %>
```
jim, line 11: Putting both cases into an if/else like this makes the flow easier to see.

jim (approved):
This looks good to me. Thanks for the contribution.

Apologies for the late review, we moved recently and things are just getting settled.

PR #1015, opened 2022-03-11
Allow admins to view previous holds

stephanieminn:
Thanks to @jutonz for pairing with me on this 🙂 

# What it does

Admins can toggle to see an item's active or inactive holds.

# Why it is important

Resolves issue https://github.com/rubyforgood/circulate/issues/980

# UI Change Screenshot

image

# Implementation notes

A few open questions:

1. We assumed "previous/historic" holds (as written in the [issue](https://github.com/rubyforgood/circulate/issues/980)) meant inactive, but it looks like there are more granular states. Was this the correct assumption or should the list filter on something else?
2. Related, we labeled the buttons "Active" and "Inactive" based on what I saw in the code. Will that work or there more user-friendly terms to use here? 
3. Inactive holds shows the same data in the table as active holds, but it looks like some of the columns might not be relevant, or that some additional information for inactive holds could be helpful! Let me know if there's a change I can make here.

### DIFF
--- a/app/controllers/admin/item_holds_controller.rb
+++ b/app/controllers/admin/item_holds_controller.rb
@@ -2,13 +2,25 @@ module Admin
   class ItemHoldsController < BaseController
     def index
       @item = Item.find(params[:item_id])
-      @holds = @item.holds.order(created_at: :asc)
+      holds_scope = @item.holds.order(created_at: :asc)
       @holds =
         if params[:inactive]
-          @holds.inactive
+          holds_scope.inactive
         else
-          @holds.active
+          holds_scope.active
         end
     end
+
+    def wait_time(hold)
+      if hold.active?
+        helpers.time_ago_in_words(hold.created_at)
+      elsif hold.ended_at.present?
+        helpers.distance_of_time_in_words(hold.created_at, hold.ended_at)
+      else
+        helpers.distance_of_time_in_words(hold.created_at, hold.expires_at)
+      end
+    end
+
+    helper_method :wait_time
   end
 end
--- a/app/controllers/admin/item_holds_controller.rb
+++ b/app/controllers/admin/item_holds_controller.rb
@@ -2,7 +2,13 @@ module Admin
   class ItemHoldsController < BaseController
     def index
       @item = Item.find(params[:item_id])
-      @holds = @item.active_holds.order(created_at: :asc)
+      @holds = @item.holds.order(created_at: :asc)
+      @holds =
+        if params[:inactive]
+          @holds.inactive
+        else
+          @holds.active
+        end
     end
   end
 end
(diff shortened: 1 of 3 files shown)


app/controllers/admin/item_holds_controller.rb
```diff
@@ -2,7 +2,13 @@ module Admin
...
     def index
       @item = Item.find(params[:item_id])
-      @holds = @item.active_holds.order(created_at: :asc)
+      @holds = @item.holds.order(created_at: :asc)
+      @holds =
+        if params[:inactive]
+          @holds.inactive
+        else
+          @holds.active
+        end
```
jim, line 11: This is totally fine, although I find the flow easier to follow when we can avoid re-assigning the value of `@hold` like this:

```suggestion
      holds_scope = @item.holds.order(created_at: :asc)
      @holds =
        if params[:inactive]
          holds_scope.inactive
        else
          holds_scope.active
        end
```

app/views/admin/item_holds/index.html.erb
```diff
@@ -1,4 +1,12 @@
 <%= render "admin/items/profile" do %>
+<div class="btn-group mt-2">
+  <%= link_to_unless params[:inactive].blank?, "Active", url_for, method: :get, class: "btn btn-sm" do %>
```
jim, line 3: Per your question, let's use "Current" as the label here instead of "Active".

app/views/admin/item_holds/index.html.erb
```diff
@@ -1,4 +1,12 @@
...
+    <button class="btn btn-sm active">Active</button>
+  <% end %>
+  <%= link_to_unless_current "Inactive", url_for(inactive: true), method: :get, class: "btn btn-sm" do %>
```
jim, line 6: And the text of this link can be "Ended". I think that will be more familiar to users.

test/system/admin/items_test.rb
```diff
@@ -161,4 +161,16 @@ class ItemsTest < ApplicationSystemTestCase
     assert_text "Item was successfully created"
     assert_equal 1, Item.all.map(&:name).count(@item.name)
   end
+
+  test "viewing an item's holds" do
```
jim, line 165: Nice! Thanks for adding a test.

jim (changes requested):
This is great! Thanks for working through that issue, especially since it was a bit light on details.

I'd say the link text changes are important, and the tweak to the controller code is a nice-to-have.

jim:
I forgot to respond to your third question:

> Inactive holds shows the same data in the table as active holds, but it looks like some of the columns might not be relevant, or that some additional information for inactive holds could be helpful! Let me know if there's a change I can make here.

This is a good point. I think what we'd want to know would be:

1. A link to the member who created the hold
2. When the hold was created
3. If the hold was picked up (this is indicated by `hold.loan` being set).
4. How long the person was waiting for the hold. This will be `ended_at - created_at` if `ended_at` is set (which is when a person picked up the hold) or `expired_at - created_at` for cases when the item wasn't picked up.

jim (approved):

jim:
@stephanieminn Thanks again for working through those changes. This is great!

PR #1028, opened 2022-04-08
Switch to the default async adapter for emails

jim:
# What it does

This PR replaces SuckerPunch with the built-in async adapter provided by Rails.

# Why it is important

Rails now provides a bulit-in async adapter for ActiveJob that provides nearly identical behavior to SuckerPunch. That library's docs even recommend that folks use the default `:async` adapter. So let's do that and remove a dependency.


jim:
@dependabot rebase

PR #1133, opened 2023-09-27
Fix "missing keyword: :enum_type" in migrations

phinze:
# What it does

I think this makes migrations work!

The `activerecord-postgres_enum` gem was updated across a major version bump in https://github.com/chicago-tool-library/circulate/pull/928

The 2.0.0 gem release notes include the change:

> BREAKING Renamed enum_name to enum_type because Rails 7 added basic support for enums

# Why it is important

I believe this is necessary for `db:migrate` to work.

# Implementation notes

The `schema.rb` changes look to be cosmetic reordering _except_ the dropping of this line:

```
enable_extension "pg_stat_statements"
```

It looks like this is never added via a migration. From some history spelunking I see it was:

 1. Added in ba23291c7594c429aec32226c54ed8c48514c905
 2. Removed explicitly in 6b2e1d2467079609a7324d015e8bed312d7fab86
 3. Added back in 8e35c8f46ef63621b051b95f5860f926b308abdb

Do we know if we want this included or not? If so it looks like we can [enable it in a migration](https://stackoverflow.com/questions/35180280/rebuild-database-enable-extension-pg-stat-statements-line-got-removed-in-schema) to get it to stick in the schema.

### DIFF
--- a/db/migrate/20190807020247_add_payment_info_to_adjustments.rb
+++ b/db/migrate/20190807020247_add_payment_info_to_adjustments.rb
@@ -3,7 +3,7 @@ class AddPaymentInfoToAdjustments < ActiveRecord::Migration[6.0]
     create_enum :adjustment_source, ["cash", "square"]
 
     change_table :adjustments do |t|
-      t.enum :payment_source, enum_name: :adjustment_source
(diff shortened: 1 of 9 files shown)


jim:
Ran this locally and `bin/reset` works.

PR #1135, opened 2023-09-30
Fix permissions issue and missing assets in dockerfile

archonic:
# What it does

This fixes a permissions issue I was experiencing on WSL2 + Docker. It also fixes missing assets by baking them into the image.

# Why it is important

Improves on boarding

# Implementation notes

Assuming docker-compose is being used in a dev environment, there's another service that should be running for webpack with the dev reload enabled. I left this out as we're moving to esbuild anyways. I can implement that service once that PR is merged.

* I have the time and interest to make additional changes to this PR based on feedback.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,8 @@ WORKDIR $RAILS_ROOT
 COPY --from=builder $RAILS_ROOT $RAILS_ROOT
 COPY --from=builder /usr/local/bundle/ /usr/local/bundle/
 
+RUN rails assets:precompile
+
 EXPOSE 3000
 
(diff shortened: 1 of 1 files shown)


archonic:
And now I've found [this gem](https://github.com/fly-apps/dockerfile-rails) which generates Dockerfiles and has all kinds of goodies like `--fullstaq` and `--yjit`. It looks like it will be included by default in Rails 7.1. I guess I'll make a new PR for 7.1 which moves to the Rails standard for the Dockerfile.

jim:
Moving to that sounds like a good plan to me. I'll merge this to make things better for folks in the short term. Thanks for making these changes.

PR #1136, opened 2023-09-30
Rubocop with Rails defaults

archonic:
# What it does

Not sure how you feel about rubocop but this introduces it and implements autofixes with the [Rails 7.0 defaults](https://github.com/rails/rails/blob/v7.0.8/.rubocop.yml) (without rubocop-packaging).

# Why it is important

Formalizes acceptable syntax without enforcing it

*  I have the time and interest to make additional changes to this PR based on feedback.


archonic:
Branched from the wrong branch -_-

PR #1137, opened 2023-09-30
Rubocop with Rails 7.0 defaults + autofixes

archonic:
# What it does

Not sure how you feel about rubocop but this introduces it and implements autofixes with the [Rails 7.0 defaults](https://github.com/rails/rails/blob/v7.0.8/.rubocop.yml) (without rubocop-packaging).

# Why it is important

Formalizes acceptable syntax without enforcing it

*  I have the time and interest to make additional changes to this PR based on feedback.


jim:
Thanks for taking the time to put this together! Code consistency is important.

We have been using [standard](https://github.com/standardrb/standard), which is a basically a set of config for Rubocop that doesn't have any knobs (which I admit is a pro or con depending on how you look at it 😄). It's closer to my personal preferences, too, but mostly I like not having anything to configure or talk about configuring.

Before we switch from standard all together, I think it would be worth upgrading it since we're pretty far behind and probably adding [standard-rails](https://github.com/standardrb/standard-rails) which we hadn't had setup previously. If that's interesting and you'd want to give it a shot, great! But if not, I no worries and I'll probably give it a try over the next week or so. Then I'll either close or merge this based on how it turns out.

jim:
Thanks again for getting the ball rolling on the Rails 7 upgrade! I think we're in a good place now with #1190 being merged, so I'll close this issue.

PR #1139, opened 2023-10-03
Revive PR#1067

jim:
This PR will merge in the work done by @stephanieminn  in #1067:

# What it does

- Allows members to remove holds that are associated with an appointment, and ensures that the hold removed from the appointment.
- If the hold was the only item in the appointment, the appointment is canceled.

# Why it is important

Resolves https://github.com/rubyforgood/circulate/issues/802 and https://github.com/rubyforgood/circulate/issues/666!

# UI Change Screenshot

### Alert when a user tries to remove a hold associated with an appointment:
image

# Implementation notes

I noticed that `Hold` was defined to have a `has_many` relationship with appointments, but experimenting with the app led me to believe that it shouldn't be possible for a hold to be added to any other appointment. As a result, I did a lil refactor to enforce that a hold `has_one` appointment relationship to make the subsequent work easier.

I also added a `dependent: :destroy` strategy for appointment holds to avoid orphaned records when a hold is destroyed. If these were the wrong assumptions, please let me know!

### DIFF
--- a/app/controllers/account/holds_controller.rb
+++ b/app/controllers/account/holds_controller.rb
@@ -19,7 +19,9 @@ module Account
     end
 
     def destroy
-      current_member.active_holds.find(params[:id]).destroy!
+      hold = current_member.active_holds.find(params[:id])
+      hold.cancel!
+
       redirect_to account_holds_path
     end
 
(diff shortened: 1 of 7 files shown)


phinze (approved):
This looks pretty straightforward!

Issue #1149, opened 2023-10-10
Undoing a loan successfully shows an error message


jim:
When a staff member clicks on "Undo loan" to undo a recently created loan, they will see an error message that says the loan couldn't be undone even when it was actually deleted successfully.

The error is likely that [this check in `undo_loan`](https://github.com/chicago-tool-library/circulate/blob/main/app/controllers/concerns/lending.rb#L53) returns `false`, and that value ends up being [returned to the controller](https://github.com/chicago-tool-library/circulate/blob/main/app/controllers/admin/loans_controller.rb#L49-L53) where it causes the error message to be erroneously displayed.

The fix here is:
1. Ensure that `undo_loan` returns `true` when the undo is successful.
2. Add a test to [lending_test.rb](https://github.com/chicago-tool-library/circulate/blob/main/test/controllers/concerns/lending_test.rb) that verifies the correct behavior.

jim:
This was fixed in #1151!

PR #1151, opened 2023-10-11
return true when undo_loan is successful

rendonluis:
# What it does

`undo_loan` returns true when loan is successfully destroyed.

# Why it is important

Fixes bug where error message is displayed even if loan was successfully undone described [here](https://github.com/chicago-tool-library/circulate/issues/1149)

# UI Change Screenshot

Old:
image
New:
image

### DIFF
--- a/app/controllers/concerns/lending.rb
+++ b/app/controllers/concerns/lending.rb
@@ -54,6 +54,7 @@ module Lending
         loan.item.increment_quantity
       end
     end
+    return true
   end
 
   def restore_loan(loan)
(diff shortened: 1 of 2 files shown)


jim (changes requested):
Hey, awesome! Thanks for making this change. I'm thinking it would be great to add a test to cover the "undo a non-consumable loan" case, which should be very similar to the one that you updated but with a different `borrow_policy`. We missed adding coverage for that case whenever this code was originally written.

The CI check for "standardrb" that's failing can likely be fixed using `bundle exec standardrb --fix`. I think it doesn't like the use of the `return` keyword when you can just use `true` on that line to indicate what value to return.

jim (approved):
This looks good! Thanks for making those changes 😄

PR #1155, opened 2023-10-12
Fix appointment merging bug

jim:
# What it does

This PR fixes an issue that we've seen in production where an ActiveJob instance fails to deserialize due to an Appointment not being found with a certain ID. I looked through the code, and I think the likely cause of the issue is that in some cases we merge multiple appointments that have been scheduled for the same time slot together, but we were still enqueuing a confirmation email for an Appointment that didn't exist after this merge was performed.

The errors we're seeing in production look like this:

```
Error while trying to deserialize arguments: Couldn't find Appointment with 'id'=11238
```

# Why it is important

Some members weren't getting confirmation emails, and it's nice to not have recurring errors due to bugs.

# Implementation notes

* I added a couple tests for appointment scheduling in the process of reproducing the bug.
* I refactored the code that handled merges a bit to consolidate that logic in one place.
* This change does mean we'll be sending out more emails for appointments than we previously were. We're now sending one for every appointment change, which might be noisy for members. But it seems good for them to always have the latest info in their email, so let's try this and see how it goes.


README.md
```diff
@@ -4,26 +4,26 @@
...
-    - [Running tests](#running-tests)
-    - [Code formatting and linting](#code-formatting-and-linting)
-    - [Setup pre-commit checks](#setup-pre-commit-checks)
-    - [Documentation](#documentation)
-    - [Who to log in as](#who-to-log-in-as)
-  - [Deployment](#deployment)
-    - [Buildpacks](#buildpacks)
-    - [Release Command](#release-command)
-    - [Daily Summary Emails](#daily-summary-emails)
-  - [Alternatives](#alternatives)
```
phinze, line 26: This looks like it might be diff noise from separate README sprucing?

phinze (approved):
One inline optional nit; overall this LGTM!

README.md
jim: Yeah, this is unintentional. I'll remove it.

PR #1156, opened 2023-10-13
Add Nix development setup

phinze:
I'm loathe to be the guy who comes in and messes with the development environment as the first thing he does, but here I am being that guy. 😅

I've been pretty into Nix for my development environments for the past couple of years for its ability to lock the full set of dev dependencies in a consistent way across machines and operating systems.

I'll use this setup myself either way, but it's a bit less overhead for me if I'm able to check in the configs vs dancing around local gitignores to keep it working. Plus it gives other people the opportunity to use it!

I tried to document everything nicely, keep it off to the side, and make it minimally invasive. But I totally understand if you'd rather keep this out of the repo! Just let me know. 🙇🏻 

----

Commit message:

> This setup works well for me, and should allow anybody familiar with Nix to get up and running with the project really quickly.
> 
> I documented this as an advanced development setup in the same vein as Docker, leaving the main README instructions as the happy path.
> 
> At any point this can be safely ripped out by reverting this commit or by manually doing the following:
> 
>  1. remove .envrc, flake.nix, flake.lock, and NIX.md
>  2. clear the the extra entry in .gitignore.
>  3. remove the extra configurable in application_system_test_case.rb

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,13 @@
+#
+# This file tells direnv, if both it and Nix are installed, to automatically
+# load the Nix dependencies defined in flake.nix.
+#
+# For details on the Nix development setup, see NIX.md.
+#
+if has nix; then
+  if ! has nix_direnv_version || ! nix_direnv_version 2.3.0; then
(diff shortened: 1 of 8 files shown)


jim (approved):
I think you did this in a good way, where it's optional if folks want to give it a try but it doesn't require any really changes from anyone otherwise. If it means that more folks are able to easily develop on the app and contribute to the project, that works for me 😄

phinze:
Hooray! Thank you 🙇🏻

PR #1157, opened 2023-10-14
Add note about adding library email to contacts to welcome messages

### LINKED ISSUE #1145: Include the suggestion to put "team@" as a trusted contact in emails

phinze:
# What it does

Resolves #1145 

# Why it is important

This could help with email deliverability!

# UI Change Screenshot

2023-10-14 at 12 49 05@2x

# Implementation notes

- Caught some duplication in the text version of the welcome email - added a covering test and fixed it.
- Some callers of the `instructions` partial use locals to pass `library` and others use instance variables. I think ideally moving to explicit locals is a bit cleaner, but I decided to continue to handle both in the partial for now rather than taking on the larger task of changing all the callers.
- Added basic unit tests for a couple other member emails which caught a deprecation warning on rendering a partial with a filename suffix.

### DIFF
--- a/app/views/member_mailer/membership_reminder.text.erb
+++ b/app/views/member_mailer/membership_reminder.text.erb
@@ -3,7 +3,7 @@ We'd love to see you!
 <%= HTMLToMarkdown.new.convert(render(partial: "member_mailer/reminder")) %>
 ---------------------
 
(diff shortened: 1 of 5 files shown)


jim (approved):
LGTM! Thanks for adding those missing tests.

PR #810, opened 2021-10-15
WIP: Upgrade app to hotwire

derricklannaman:
# What it does

- Start initial process to transition to Turbo/Hotwire
- Install hotwire and update Stimulus js
- Enable/use Redis and ActionCable
- Disable use of PortalRendering in NotesController
#### - Implements CRUD using hotwire
# Why it is important
This is considered the initial step for an eventual cut over to Hotwire 

# UI Change Screenshot

# Implementation note

https://user-images.githubusercontent.com/615262/137510085-a3359412-6d1f-470c-85d7-1f835ffee3c5.mov

Known Issues

Screen Shot 2021-10-25 at 7 03 59 PM

- turbo stream looking for `notes/_note` and throws missing partial error bc that file is not needed...
- but adding that file, even with nothing in it, results in a browser error
- neither breaks the application

Screen Shot 2021-10-25 at 5 12 13 PM

~~Adding a new note doesn't update in real-time~~
~~Add a Note" button disappears on click~~
~~Simple, not ready for primetime yet~~

# Your bandwidth for additional changes to this PR

_Please choose one of the following to help the project maintainers provide the appropriate level of support:_

- [x] I have the time and interest to make additional changes to this PR based on feedback.
- [ ] I am interested in feedback but don't need to make the changes myself.
- [ ] I don't have time or interest in making additional changes to this work.
- [ ] Other or not sure (please describe):

PR #1162, opened 2023-10-20
Make holds reorderable

jim:
# What it does

This PR resurrects some work I started a couple years back to make holds sortable.

# Why it is important

For a variety of reasons, staff sometimes need to adjust the order of folks in the holds line. Sometimes an item will get put in maintenance mode, and its hold list needs to be collapsed into another item. Sometimes we know a person can't make it to pick up an item and by bumping them back a spot in line, someone else gets something sooner and they keep their place.

# UI Change Screenshot

reordering holds

# Implementation notes

* If someone is first in line, this doesn't let someone else be moved ahead of them. This is because in some cases, we've sent an email to someone that says they can borrow an item. We might want to let admins do this anyway, but it's worth thinking through to avoid potentially bad experiences for members.
* The `acts_as_list` gem isn't great for large lists, but we generally have 5-10 people at most in line and are typically only moving people up or down a few spots.
* The way the reordered holds list is rendered has some duplication with the portal rendering code. That code will all be removed when we upgrade Turbolinks to Hotwire. There is [a PR already to do most of that upgrade](https://github.com/chicago-tool-library/circulate/pull/810) that I'll be working on reviving next.
* I moved two controllers (`app/controllers/admin/item_histories_controller.rb`and `app/controllers/admin/item_holds_controller.rb`) so that they were namespaced to follow the pattern used elsewhere in the admin (so they are now at `app/controllers/admin/items/histories_controller.rb` and `app/controllers/admin/items/holds_controller.rb`, respectively).

I tested the migration against a clone of the production database to make sure that it applies (and adds values for "position") cleanly.


db/schema.rb
```diff
@@ -10,76 +10,75 @@
...
+ActiveRecord::Schema.define(version: 2023_10_28_225001) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
   create_enum :adjustment_kind, [
     "fine",
     "membership",
     "donation",
-    "payment",
+    "payment"
```
jim, line 21: The churn in this file I think stems from standard not getting applied to `schema.rb`, or possibly config change between versions. When I run standard against the `schema.rb` in main, I see these changes there as well.

test/controllers/admin/items/holds/positions_controller_test.rb
```diff
@@ -0,0 +1,34 @@
...
+        setup do
+          @user = create(:admin_user)
+          create(:verified_member, user: @user)
+
+          sign_in @user
+        end
+
+        test "updates positions of holds" do
+          item = create(:item)
+          10.times.map { create(:hold, item: item) }
```
phinze, line 18: Totally optional micro-nit: [`Integer#times` takes a block directly](https://ruby-doc.org/3.2.2/Integer.html#method-i-times), so `map` is not needed here as you're not using the Array it builds and returns.

test/system/admin/holds_test.rb
```diff
@@ -152,4 +152,40 @@ def setup
...
+    @holds = 4.times.map { create(:hold, item: @item) }
+
+    visit admin_item_holds_path(@item)
+
+    @handles = all(".drag-handle")
+    assert_equal 4, @handles.size
+    hold_ids = all(".item-holds-table tr[data-id]").map { |row| row["data-id"] }
+
+    # Make the first hold the last
+    @handles.first.drag_to @handles.last
```
phinze, line 179: Ooo Capybara dragging and dropping! How exciting. 🥳

app/helpers/admin/holds_helper.rb
```diff
@@ -0,0 +1,13 @@
+module Admin
+  module HoldsHelper
+    def wait_time(hold)
```
phinze, line 3: FWIW I don't fully understand the meaning of this column in the table. I see it's the same as it was so NBD. 🙂

phinze (approved):
Looking great! A couple of inline questions and nits but nothing remotely blocking.

app/helpers/admin/holds_helper.rb
jim: I had to look at it for a while to figure out what it's doing. I'll add some comments to the branches for now, but I'm curious what staff would say about these values and whether they're useful or not.

phinze (approved):
Took a look at the more recent changes and this is still looking good to me!

PR #1164, opened 2023-10-24
Remove ImageKit

### LINKED ISSUE #1150: Remove need for image proxy
Back when we added image uploading, Rails didn't have a lot of options for generating image thumbnails. That's why we're using an image proxy in production (https://imagekit.io). It works well, but for the sake if having fewer third party dependencies and making it easier for multiple developers to recreate the production environment locally (and since our image uses are very simple), it would be nice to update how we handle item images:

* Keep uploads working as-is
* Configure Rails to track whether a thumbnail was generated already (may need a migration for this)
* Replace image tags with ones that point directly to the assets if they exist on S3
* If they don't, use the redirect method so that Rails will generate the thumbnail on the fly and then store it in S3
* Remove image kit code from the app

phinze:
# What it does

This changes the app to only rely on ActiveStorage and removes the third party dependency on ImageKit.

# Why it is important

Modern versions of Rails should be perfectly capable of serving thumbnails directly out of S3 without the help of a 3rd party service.

Resolves #1150 

# Implementation notes

* This turns ActiveStorage variant tracking on which removes the need for extra checks in S3 for variants. There's no need for an additional DB migration to turn this on as [we already have the tables from the Rails 6.1 upgrade](https://github.com/chicago-tool-library/circulate/blob/remove-imagekit/db/migrate/20210216165140_create_active_storage_variant_records.active_storage.rb).
* Since this change is messing with a dependency I'd recommend immediately testing image views, uploads, and editing after we deploy, which should let us revert if anything seems off.

### DIFF
--- a/.env
+++ b/.env
@@ -33,10 +33,6 @@ BUCKETEER_BUCKET_NAME=
 # Require basic authentication (useful for staging sites)
 HTTP_BASIC_USERS=
 
-# Imagekit.io proxy for resizing imges on the fly
-# Only set this if you have a working proxy setup!
-# IMAGEKIT_URL=https://ik.imagekit.io/proxyid/
-
 # Sendgrid is used to send email
 SENDGRID_USERNAME=
 SENDGRID_PASSWORD=
--- a/app/helpers/items_helper.rb
+++ b/app/helpers/items_helper.rb
@@ -118,32 +118,7 @@ module ItemsHelper
   end
 
   def item_image_url(image, options = {})
-    if ENV["IMAGEKIT_URL"].present?
-      transforms = if options[:resize_to_limit]
-        width, height = options[:resize_to_limit]
-        ["tr=w-#{width}", "h-#{height}", "c-at_max"]
-      else
-        []
-      end
-
-      if image.metadata.key? "rotation"
-        transforms << "rt-#{image.metadata["rotation"]}"
-      end
-
-      parts = [
-        ENV["IMAGEKIT_URL"],
-        image.blob.key
-      ]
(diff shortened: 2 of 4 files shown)


jim (approved):
This looks good to me. I'll merge this later this weekend and make sure that things are working OK in production.

phinze:
Cool will leave it to you then. Feel free to give me a ping when you do and I'm happy to help test 🥼

phinze:
Via Slack from @jim:

> I deployed the imagekit change, and while it seemed to work OK generally (images showed up, they could be uploaded and rotated in the admin without issue), the memory use of the app went up by like 50%. Presumably this is caused by the thumbnails being generated the first time they're viewed.

My research on this led me to a couple of findings:

 - This [experience report of ActiveStorage in production](https://discuss.rubyonrails.org/t/active-storage-in-production-lessons-learned-and-in-depth-look-at-how-it-works/83289/1) suggests that inline variant generation is a common issue.
 - In that report, one suggestion to mitigate memory usage is to [switch to libvips](https://discuss.rubyonrails.org/t/make-vips-the-recommended-default-variant-processor-for-active-storage/78368).
 - Another is to move variant transforms into background jobs, which is [a feature available as of Rails 7.1](https://github.com/rails/rails/pull/47473).

Given that both a vips switch and a Rails upgrade are nontrivial next steps, and given that ImageKit was not actively harming anything, I'm proposing we revert this change for now to give us a chance to follow through with Rails upgrades and we revisit with an implementation of Named Variants w/ preprocessing once we're at 7.1.

PR #1167, opened 2023-10-27
nix: add lefthook

phinze:
# What it does

Allows pre-commit linters to run in Nix development environments.

# Why it is important

Reduces likelihood of back-and-forth based on formatting issues.

# Implementation notes

Lefthook needs to be able to invoke `node` directly so explicitly add `nodejs` to dependencies as well. This is probably worthwhile for asset generation stuff anyways.


jim (approved):
Thanks for improving things for folks using nix.

PR #1168, opened 2023-10-27
Configure Solargraph with Standard Ruby

phinze:
# What it does

The current checked-in configuration runs Solargraph with a default Rubocop config which goes a bit wild with suggestions. This should make the LSP only suggest things that match the rules from Standard Ruby.

# Why it is important

Having the LSP config match the linter makes it a bit faster to write complaint code since you'll get suggestions in-line when using an LSP-supporting editor.

### DIFF
--- a/.solargraph.yml
+++ b/.solargraph.yml
@@ -8,8 +8,10 @@ exclude:
 - ".bundle/**/*"
 require: []
 domains: []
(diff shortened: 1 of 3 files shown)


jim (approved):
Awesome! This has needed some attention for a while now.

PR #1169, opened 2023-10-28
require yarn succeed in bin/setup

phinze:
# What it does

Changes the `bin/setup` script to fail if `yarn` does not work.

# Why it is important

This makes `bin/setup` halt with an error message saying to install Yarn before continuing, which should help folks getting started know they need to install that prerequisite.


phinze:
@chrissycooper I think this should prevent others from running into the situation you did!

jim (approved):
Good catch!

PR #1170, opened 2023-10-28
Available now item filtering

### LINKED ISSUE #1166: Improve item filtering
Desired functionality:
- option to hide checked out/overdue tools
- we'd like this to work on both the public inventory and the admin side. It's OK to do only one of those to get started (and we can spin up a second task to add it to the second side).

phinze:
Resolves #1166

# What it does

Allows item search results on both the public facing and admin tool inventories to be filtered by items "available now," meaning they have no holds or loans and aren't in maintenance.

# Why it is important

This is useful for people who want a tool they can pick up immediately.

# UI Change Screenshot

## Member side:

https://github.com/chicago-tool-library/circulate/assets/37534/d4e7231d-1b60-4167-9341-bef37d99adb1

## Admin side:

Screenshot 2023-10-28 at 9 25 27 AM

# Implementation notes

* I needed to convert the `Hold.active` scope to ActiveRecord so that `Item.active_holds` would be able to chain properly in the `Item.available_now` scope.
  * With the plain string version, the queries were failing with "ambiguous column" errors on `ended_at`.
  * The unit tests on that query were pretty good (as I learned when I mixed up the `>` and got a million failures 😂) so I'm reasonably confident this is a one-for-one swap out.
* Getting the toggle to trigger behavior was my first foray into Hotwire/Stimulus/Turbo land, so please feel free to nitpick my implementation there.
* (For later) I suspect we're not taking advantage of the full partial-page-swap ability of the Hotwire stack on the search pages, and I'd love to discuss that further as an opportunity to learn more.

### DIFF
--- a/app/controllers/admin/items_controller.rb
+++ b/app/controllers/admin/items_controller.rb
@@ -14,6 +14,10 @@ module Admin
         item_scope = @category.items
       end
 
+      if params[:available]
+        item_scope = item_scope.available_now
+      end
+
       item_scope = item_scope.includes(:categories, :borrow_policy, :active_holds).with_attached_image.order(index_order)
 
       @pagy, @items = pagy(item_scope)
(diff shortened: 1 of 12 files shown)


jim (approved):
This looks great! It's nice to see the refactoring of the scopes and such as a part of this new feature. Excellent test coverage as well 🚀

PR #1175, opened 2023-10-31
Ignore RecordNotFound errors in AppSignal

### LINKED ISSUE #1174: Gracefully handle item page 404s
We get a lot of exceptions because someone or something it attempting to view the public item page for an items that doesn't exist. For example, visiting https://app.chicagotoollibrary.org/items/2599 will trigger an exception that's collected by our error collection service.

The logged exception looks like this:

```
Couldn't find Item with 'id'=2599 [WHERE "items"."library_id" = $1 AND (status = 'active' OR status = 'maintenance')]
```

The default Rails behavior to catch this exception and show a generic 404 page is OK for internal pages, but since this one gets hit by anyone on the public internet, it would be nice to tweak the controller to avoid raising that exception since we can expect it to get triggered regularly.

phinze:
Resolves #1174

# What it does

Prevents AppSignal from treating RecordNotFound errors as reportable exceptions.

# Why it is important

It's important for us to only alert on real errors or we're more likely to suffer alert fatigue and miss a true problem.

# Implementation notes

* My first stab at this was reworking `ItemsController#show` to not raise an exception, but my research suggests that letting an exception generate the 404 is a normal setup for Rails. We already have [routes](https://github.com/chicago-tool-library/circulate/blob/ab0a397598cb1d454ea3b495b5d35eab7ad26210/config/routes.rb#L188-L190) and [`exception_app`](https://github.com/chicago-tool-library/circulate/blob/main/config/application.rb#L47-L48) set up to do this.
* I looked into other mechanisms for handling exceptions like [`rescue_from`](https://edgeapi.rubyonrails.org/classes/ActiveSupport/Rescuable/ClassMethods.html#method-i-rescue_from) and [`Rails.error.handle`](https://guides.rubyonrails.org/error_reporting.html), but I didn't see any obvious advantage to switching everything to either of these.
* The AppSignal gem can be configured using env vars or a config file. We currently set only `APPSIGNAL_PUSH_API_KEY` in Heroku, but since we're expanding our options here it seemed worth switching to file-based config so we can have comments. This config is based on the one generated by the `appsignal install` command.

### DIFF
--- a/config/appsignal.yml
+++ b/config/appsignal.yml
@@ -1,0 +1,24 @@
+default: &defaults
+  push_api_key: "<%= ENV["APPSIGNAL_PUSH_API_KEY"] %>"
+  name: "Circulate"
+
+  # Errors that should not be recorded by AppSignal
+  # For more information see our docs:
+  # https://docs.appsignal.com/ruby/configuration/ignore-errors.html
+  ignore_errors:
+    # We do error page rendering via Rails config.exceptions_app, which means
+    # that RecordNotFound exceptions coming from controller actions are ok.
+    - ActiveRecord::RecordNotFound
+
+  # See https://docs.appsignal.com/ruby/configuration/options.html for
+  # all configuration options.
+
+# Configuration per environment, leave out an environment or set active
+# to false to not push metrics for that environment.
+development:
+  <<: *defaults
+  active: false
+
+production:
+  <<: *defaults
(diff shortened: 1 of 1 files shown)


jim (approved):
Cool. Thanks for doing all the research on the best way forward here. I agree with the move to using a config file for this.

PR #1138, opened 2023-09-30
Rails 7 Upgrade

archonic:
# What it does

Upgrades to Rail 7 and closes #975.

# Why it is important

Upgrades are good :+1:

# Implementation notes

There was one deprecated use of `to_s` which we can now use `to_sf` for. A few other deprecations had no real impact according to tests. There was some host redirection issues as well which were fixed in one case by just allowing an external redirect to a payment completion page. Another fix for a host redirect was actually just a `www` issue. The fix assumes that the shortlinks controller is not meant to be used for external URLs.

# Your bandwidth for additional changes to this PR

* I have the time and interest to make additional changes to this PR based on feedback.

PR #1178, opened 2023-11-06
Rails 7 Upgrade

phinze:
# What it does

Upgrades the application to Rails 7.

# Why it is important

Getting on the latest version of Rails unlocks a few underlying features we can use to implement improvements to the application.

# Implementation notes

* This builds on and supersedes #1138  
* This holds off on switching to `vips` for image processing. We will probably do that as part of follow on work on ActiveStorage variants.
* The [upgrade guide](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-6-1-to-rails-7-0) suggests there are changes to `button_to`. No system tests are failing but we should audit the behavior to make sure everything still works.


phinze:
Interesting, those failures don't show up on my machine. I think it's a central time vs UTC time zone thing. 🔍

phinze:
Finally got repro! CI is using Postgres 11 and dev has us on 12, while Heroku looks like its on 13. Will get these all matched up which should fix tests.

jim (approved):
Changes look good!

I pulled down the branch and clicked around. Everything seems to be working.

For the `button_to` changes, I checked a few spots and everything works normally. I read the docs again and I think it only affects cases where you pass an `ActiveRecord` instance and it computes the path automatically; we seem to explicitly pass a path everywhere so I think we won't be affected.

jim:
I had two other thoughts about the `Read: #<ArgumentError: wrong number of arguments (given 1, expected 3)>` error being logged locally:

1. Maybe it's from an incompatibility between Rails 7 and the version of Rack we have? I noticed Rack's version didn't change in this PR.
2. There's a small gem [embedded in the app](https://github.com/chicago-tool-library/circulate/tree/main/gems/active_storage_log_suppressor) that was needed back in Rails 5 to quiet the logging of ActiveStorage requests. In development [it wraps calls into the app](https://github.com/chicago-tool-library/circulate/blob/c6d6c28b527fcc3f3def089ba3d53a313f752730/gems/active_storage_log_suppressor/lib/active_storage_log_suppressor/filters.rb#L35) and could possibly be swallowing an exception (or preventing it from breaking things in an expected way).

I'm happy to help hunt down the source of this if you need extra eyes on it 😄

jim:
I realized I could just delete the gem and it looks like the error went away. If that works for you as well, I say we just get rid of it.

phinze:
Thanks for the extra dive on this! Will remove that gem and get this merged and deployed now.

PR #1182, opened 2023-11-08
Fix error when logged out user views membership page

phinze:
# What it does

Fixes error seen here https://appsignal.com/chicago-tool-library/sites/60596f9214ad662e17191689/exceptions/incidents/144

# Why it is important

Prevents logged out users from seeing an error page if they pull up membership page somehow.

# Implementation Notes

I found `BaseController` in the `account` directory which requires auth for all actions. It seemed like the best fix was to inherit from that.


jim (approved):
Nice catch.

PR #1183, opened 2023-11-09
Remove sentry

### LINKED ISSUE #1160: Remove Sentry gem
We have both AppSignal and Sentry configured for the project. Since the Sentry gem is EOLed and needs to be upgraded anyway, let's just remove it for now and see how things go.

chrissycooper:
# What it does

This PR closes #1160 to Remove the outdated Sentry gem, as well as removes related code causing tests to fail after its removal. Thank you for your patience 

# Implementation notes

I took a very top-down approach to this ticket, first removing the sentry-raven gem, then running tests and inspecting the related files. I'm very new to the codebase, so I appreciate any extra pairs of eyes to my changes. 

I noticed my `yarn.lock` file had a *lot* of changes to it. Please take a look at that commit. I don't know if it had to do with some of the issues I had in my set up. I'd like to make sure it doesn't cause any problems in the future. Appreciate it!

- [💚] I have the time and interest to make additional changes to this PR based on feedback.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -4,7 +4,7 @@
 
 "@babel/runtime@^7.14.6":
   version "7.16.0"
-  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.16.0.tgz#e27b977f2e2088ba24748bf99b5e1dece64e4f0b"
+  resolved "https://registry.npmjs.org/@babel/runtime/-/runtime-7.16.0.tgz"
   integrity sha512-Nht8L0O8YCktmsDV6FqFue7vQLRx3Hb0B37lS5y0jDRqRxlBG4wIJHnf9/bgSE2UyipKFA01YtS+npRdTWBUyw==
   dependencies:
     regenerator-runtime "^0.13.4"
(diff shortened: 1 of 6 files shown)


chrissycooper:
I'll spend some time poking into why the CI failed tomorrow, if anyone ends up being around

yarn.lock
```diff
@@ -4,7 +4,7 @@
 
 "@babel/runtime@^7.14.6":
   version "7.16.0"
-  resolved "https://registry.yarnpkg.com/@babel/runtime/-/runtime-7.16.0.tgz#e27b977f2e2088ba24748bf99b5e1dece64e4f0b"
+  resolved "https://registry.npmjs.org/@babel/runtime/-/runtime-7.16.0.tgz"
```
jim, line 7: I'm not totally sure, but I am wondering if you used `npm` to install the node packages instead of `yarn`? There could be something else at play here as well.

In any case, since `yarn.lock` is for JS dependencies and `sentry-raven` is a Ruby dependency, we shouldn't need to make any changes to this file as a part of this PR. I think a `git checkout origin/main -- yarn.lock` will undo these changes (which you'll need to then commit).

jim (changes requested):
This is awesome! Thanks for being thorough and removing the other uses of the raven library in the codebase- I had forgotten about those entirely.

I took a peak at CI, and I think the main issue is that the `Gemfile.lock` needs to be updated along with the `Gemfile`. [This is the error I found](https://github.com/chicago-tool-library/circulate/actions/runs/6817905031/job/18542414186#step:4:43).

[These docs](https://bundler.io/guides/using_bundler_in_applications.html#recommended-workflow) have a decent list of steps to follow with Bundler; for the most part you'll need to run `bundle install` inf you haven't already, it'll make a change to `Gemfile.lock`, and then you'll need to commit that file and push it to your branch.

I will probably be around tomorrow, so feel free to ping me in Slack then or whenever if I can help. This is close and it'll be great to make a few tweaks and get it merged in 🎉

yarn.lock
chrissycooper: I think I did use npm before I got yarn installed, which is probably an origin of a lot of my hiccups. I'll make sure to remove the `yarn-lock` changes. Thanks, Jim! Super helpful, much appreciated.

chrissycooper:
Sweet, I updated the requested changes and removed a little more related code from a different file and the checks are passing! Let me know if there's anything else I can adjust

jim (approved):
This looks great! Thanks for working through those issues. CI is green, so I'll go ahead and merge this 😄

Issue #1184, opened 2023-11-10
Upgrade app from Turbolinks to Turbo


jim:
We need to revive the work started in #761, update the versions of turbo and stimulus, and convert our custom "portal rendering" stuff to use Turbo Frames. This will make it easier for folks to work in the app as it will be up-to-date with current documentation and common Rails practices.

PR #1186, opened 2023-11-10
Rails 7.1 Upgrade

phinze:
Fast on the heels of 7.0 here comes the latest and greatest Rails version!

# What it does

Gets us on Rails 7.1.

# Why it is important

Latest is greatest. And specifically this gets us access to background preprocessing of variants for finishing #1150 

# Implementation notes

* This gets us off the fork of `audited` we were using. The [change we switched for was upstreamed](https://github.com/collectiveidea/audited/pull/609) and we need the [more recent work to make it compatible with Rails 7.1](https://github.com/collectiveidea/audited/pull/686). Just a couple of code changes were needed and tests seem to be passing okay.
* Updates us to [Devise 4.9.3](https://github.com/heartcombo/devise/blob/main/CHANGELOG.md#493---2023-10-11) which added Rails 7.1 support.
  - Note Devise 4.9.0 [added Hotwire + Turbo support](https://github.com/heartcombo/devise/blob/main/CHANGELOG.md#490---2023-02-17), something we might want to take a look at in either #1184 or a follow on task. 
* Includes a bump to Postgres 15 for the Docker and CI setups to match production.
* The new Rails defaults in 7.1 raise in controllers when you mention an action that is not defined. This caught a couple of `before_action` typos.
* Running `rake app:update` yielded diffs on many of the config files. I took all the changes to `bin` scripts whole hog and merged the rest. A second set of eyes across those diffs would be helpful though! 👀 
* 💥 The upgrade caused an issue with threaded background work in tests whose fix is at the bottom of `config/environments/test.rb` with a note explaining.
* 💡 I had neglected to update the Yarn dependencies for Rails in the 7 upgrade, so catching those up here.
* 🚧 Asset compilation now runs on every execution of `bin/rails test` which made the SCSS deprecation warnings get pretty annoying. Pulling in a workaround from https://github.com/picturepan2/spectre/issues/694
* Zeitwork (the new code autoloader) did not like `lib/rails_solargraph.rb`, so following a [comment on the referenced gist](https://gist.github.com/castwide/28b349566a223dfb439a337aea29713e?permalink_comment_id=4346127#gistcomment-4346127) I switch us to the solargraph-rails gem.


app/models/item.rb
```diff
@@ -220,7 +220,7 @@ def audited_attributes
   end
 
   # called when item is updated
-  def audited_changes
+  def audited_changes(**args)
```
phinze, line 223: This started taking args in https://github.com/collectiveidea/audited/pull/657 so every Item interaction started throwing `ArgumentError`s. Since we call `super` with no args I believe they should automatically be passed along, so it's fine to just capture the args to make the method signature match the callers.

jim (approved):
Looks good! Good call on the devise turbo support, I'll rebase that PR on this one and see if I can get that sorted as a part of that work.

phinze:
@jim re-requesting review here since I made a lot more changes on this today ⚒️

bin/setup
```diff
@@ -1,38 +1,33 @@
...
-  #   FileUtils.cp 'config/database.yml.sample', 'config/database.yml'
+  # unless File.exist?("config/database.yml")
+  #   FileUtils.cp "config/database.yml.sample", "config/database.yml"
   # end
 
   puts "\n== Preparing database =="
-  system! 'bin/rails db:setup'
-
-  puts "\n== Loading dev data =="
-  system! 'bin/rails devdata:load'
```
jim, line 31: This line didn't make it over into the newer version. It theoretically saves first time devs a step when they start up the app for the first time, so probably worth keeping unless there's a reason we'd want `bin/setup` to leave the database totally empty.

config/environments/development.rb
```diff
@@ -41,22 +45,26 @@
...
-  # Debug mode disables concatenation and preprocessing of assets.
-  # This option may cause significant delays in view rendering with a large
-  # number of complex assets.
-  # config.assets.debug = true
+  # Highlight code that enqueued background job in logs.
+  config.active_job.verbose_enqueue_logs = true
 
   # Suppress logger output for asset requests.
-  # config.assets.quiet = true
+  config.assets.quiet = true
```
jim, line 64: Ah yes, there it is 😄

jim (approved):
I read through the full diff again, one line was dropped from `bin/setup`, but otherwise everything looks good. Thanks for the thorough notes in the PR description, it made understanding the changes a lot easier.

PR #1190, opened 2023-11-13
Add standard-rails

phinze:
It was mentioned in #1137 and it should be helpful to keep around to make sure we're following some standard style practices with Rails

# What it does

Adds extra Rails-centric rules to the Ruby linter.

# Why it is important

Keeps all contributors writing similar looking code.


jim (approved):
Changes look OK to me. I'm not sure why ERBLint started failing, do the Standard checks get run on ERB now as well?

phinze:
I think the new errors are because erblint uses rubocop under the hood and this updated the rubocop version. I cleaned up the errors and I think we're good to go!

PR #1197, opened 2023-11-20
Initial support for SMS notifications (feature flagged)

phinze:
# What it does

Adds SMS notification support with overdue notices as a first reminder.

# Why it is important

Hopefully SMS notifications will help make it more likely that members will remember their appointments and return overdue items.

# UI Change Screenshot

2024-01-22 at 16 21 13@2x

# Implementation notes

* Overall the design is meant to mirror ActionMailers. I started by using a gem called `textris` which takes this shape, but the gem itself had a lot of compatibility issues so I ended up deciding to roll our own simple version of the design.
* The unit tests stub out the Twilio client, but interacting with this to send real messages requires Twilio credentials. I'll document more about the Twilio integration in the README in an upcoming PR.
* SMS Notifications are behind a feature flag `FEATURE_SMS_REMINDERS` which is set to on in development but will be off in production until we are happy with testing in staging.
* For saving `Notification` records, I decided to use the same model and loosen the schema to be able to cover SMS and emails.
* Changed the user factories/seeds to default to having a phone number that will generate a successful SMS response when using Twilio test credentials.
* The UI changes are not behind the feature flag, but I feel like they're fine to leave in as we're ok to collect user preferences ahead of the feature being on. I can change this if folks feel differently though.
* There's a small nix devenv tweak piggybacking on here that fixes a warning when `convert` is called in the tests.


app/lib/twilio/result_wrapper.rb
```diff
@@ -0,0 +1,19 @@
+module Twilio
+  class ResultWrapper
```
jim, line 2: We have [a `Result` class](https://github.com/chicago-tool-library/circulate/blob/main/app/lib/result.rb) already that might be worth subclassing/keeping a similar API to (at least the `success?` and `failure?` methods).

Although we might not be using this class yet anyway? I will say that converting third party API responses to results has worked well in the codebase so far, so 👍🏻 on the general pattern if we end up needing it here.

db/migrate/20240122202020_relax_notifications_schema_for_sms.rb
```diff
@@ -0,0 +1,15 @@
+class RelaxNotificationsSchemaForSms < ActiveRecord::Migration[7.1]
```
jim, line 1: I agree with the approach you used here. We never finished implementing the enum stuff for notifications anyhow, and this makes it a lot more flexible.

app/texters/member_texter.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  def initialize(member)
+    @member = member
+  end
+
+  def overdue_notice(summaries)
+    return unless member.reminders_via_text?
+
+    message = <<~EOM
+      Chicago Tool Library Reminder: You have #{pluralize(summaries.length, "overdue item")}! Please schedule a return appointment at #{account_member_url}
```
jim, line 14: I think that `summaries` can contain items that are not overdue but merely checked out. You can use `LoanSummary#overdue_as_of` to determine if the summary is for an overdue item.

This worked for the emails because they render the entire state of your account, and the items that are overdue [get called out](https://github.com/chicago-tool-library/circulate/blob/main/app/views/member_mailer/_summaries.erb#L13-L17).

jim (approved):
Heck yeah, excited to see this landing. I appreciate the way you designed this, and the tests are great!

app/texters/member_texter.rb
phinze: Good call to double check! It looks like these summaries are already [filtered to be overdue only](https://github.com/chicago-tool-library/circulate/blob/sms-reminders/app/lib/activity_notifier.rb#L18) before being passed in. You can see the [overdue template](https://github.com/chicago-tool-library/circulate/blob/a3db478b5cb2ac7c1893c3cf18946852d55ebaa6/app/views/member_mailer/_overdue.erb) just displays all of them them without filtering.

app/lib/twilio/result_wrapper.rb
phinze: This was dead code from an earlier iteration anyways so just ended up removing it. Good to know about that general purpose Result class though.

app/lib/activity_notifier.rb
```diff
@@ -56,6 +60,13 @@ def each_member(ids, &block)
         .or(member.loan_summaries.includes(:renewal_requests).checked_out)
         .includes(item: :borrow_policy)
       yield member, summaries
+    rescue => e
```
jim, line 63: I think this is a totally reasonable way to handle failures at this point. Bonus points for sending the exception to AppSignal so we can track them.

jim (approved):

PR #1209, opened 2023-11-30
Add controller tests for renewals/payments

jim:
# What it does

Adds a controller test for `Renewals::PaymentsController`, based heavily on the one we have for `Signup::PaymentController`. This is a followup to #1192, which fixed a bug but did not include tests for timing reasons.

# Why it is important

This is a critical flow in the application, and we'd like to ensure that it is still working as intended 😄 

# Implementation notes

* There's a decent amount of duplication between `Renewals::PaymentsController` and `Signup::PaymentController`, and the same is true of their tests. It is looking like we may end up re-working the payment flows as a part of upcoming work, so I opted to not do anything too invasive at this time.

### DIFF
--- a/app/controllers/renewal/payments_controller.rb
+++ b/app/controllers/renewal/payments_controller.rb
@@ -63,7 +63,7 @@ module Renewal
       end
 
       Rails.logger.error(errors)
-      reset_session
+      session.delete(:attempts)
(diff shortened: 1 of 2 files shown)


jim:
I'm going to merge this without a review as it fixes a production issue for folks renewing.

PR #1211, opened 2023-11-30
Add member with appointment to devdata

jim:
# What it does

This PR renames the devdata members so that their emails are more representative (e.g. `member_with_holds@example.com` instead of `member1@example.com`). It also adds a new member to that dataset with an upcoming appointment.

# Why it is important

I'm hopeful this will make it easier to know what devdata member has which attributes, and that having one with an appointment will reduce how much work someone has to do if they are going to work on something related to appointments.

# Implementation notes

* I also tweaked the script to add pronouns and a member number to these members so that they were more similar to our real members.

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -66,7 +66,7 @@ seed_library(chicago_tool_library)
 denver_tool_library = Library.create!(
   name: "Denver Tool Library",
   hostname: "denver.circulate.local",
-  city: "DEnver",
+  city: "Denver",
   email: "team@denvertoollibrary.org",
   address: <<~ADDRESS.strip,
     Denver Tool Library
--- a/lib/tasks/devdata.rake
+++ b/lib/tasks/devdata.rake
@@ -65,39 +65,70 @@ namespace :devdata do
       Library.all.each_with_index do |library, index|
         postal_code = postal_codes[index]
         ActsAsTenant.with_tenant(library) do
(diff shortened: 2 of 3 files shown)


db/seeds.rb
```diff
@@ -66,7 +66,7 @@ def seed_library(library, email_suffix = "", postal_code = "60609")
 denver_tool_library = Library.create!(
   name: "Denver Tool Library",
   hostname: "denver.circulate.local",
-  city: "DEnver",
+  city: "Denver",
```
chrissycooper, line 69: Love the attention to detail! Let's go, PROOFREADER

lib/tasks/devdata.rake
```diff
@@ -65,39 +65,70 @@ namespace :devdata do
       Library.all.each_with_index do |library, index|
         postal_code = postal_codes[index]
         ActsAsTenant.with_tenant(library) do
-          create_member(holds: 2, waiting_holds: 2, loans: 2, postal_code: postal_code)
-          create_member(loans: 5, postal_code: postal_code)
-          create_member(postal_code: postal_code)
+          create_member("member_with_holds_and_loans", holds: 2, waiting_holds: 2, loans: 2, city: library.city, postal_code: postal_code)
```
chrissycooper, line 68: Thank you this is so much clearer!

chrissycooper (approved):
Looks great, thanks for upgrading this!

PR #1218, opened 2023-12-14
Fix Capybara log fetching in Chrome 120

phinze:
# What it does

System tests started failing at https://github.com/chicago-tool-library/circulate/actions/runs/7199427363/job/19648174580

This looks to be because selenium-webdriver started fetching Chrome 120 which ran us into this issue: https://github.com/SeleniumHQ/selenium/issues/13112

The recommended workaround there is to change the `--headless` argument in Chrome to `--headless=new` which seems to work locally.

# Why it is important

Build's gotta pass!

# Implementation notes

* The `flake.lock` changes are a result of running `nix flake update` which I did to update to Chrome 120 in the nix dev environment.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "flake-compat": {
       "flake": false,
       "locked": {
-        "lastModified": 1673956053,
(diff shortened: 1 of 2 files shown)


jim (approved):
Nice catch! Thanks for keeping an eye on this.

PR #1222, opened 2023-12-19
Fix appointment titles

jgill123:
# What it does

Fixes an issue where, after 6pm CST, the appointments page would show incorrect labels for the day. It would show 'Today' for tomorrow and 'Tomorrow' for the day after tomorrow. 

In application_controller.rb, we attempt to manually set the time zone in the around_action to use Chicago time. However, there are a few places where we were using time/day functions that don't take time zone into account. See this article [here](https://thoughtbot.com/blog/its-about-time-zones#a-summary-of-do39s-and-don39ts-with-time-zones) for more info on this. Switching to use suggested functions should resolve the issue and we will apply the correct labels. 

Note: wasn't able to reproduce the exact issue locally, possibly because my local machine is already in Chicago time, but the changes seem low risk enough. 

# Why it is important

#1215 

# UI Change Screenshot

NA

# Implementation notes

### DIFF
--- a/app/controllers/admin/appointments_controller.rb
+++ b/app/controllers/admin/appointments_controller.rb
@@ -65,9 +65,9 @@ module Admin
     end
 
     helper_method def current_day_label
-      if @current_day == Date.today
+      if @current_day == Date.current
         "Today"
-      elsif @current_day == Date.tomorrow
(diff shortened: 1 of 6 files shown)


jim (approved):
This looks good to me! Thanks for digging into this one.

PR #1224, opened 2023-12-21
Remove special case for numbering A tools. Fixes #1223.

### LINKED ISSUE #1223: Update Borrow Policy numbering logic
- The original numbering system needs to be updated before we run out of numbers in the A range.
- This involves removing some special case logic that's applied to A tools.
- Going forwards, A tools will select the next available number instead of finding on in the range between 1-999.

jim:
# What it does

We originally treated A tools (those that we don't individually number) specially, with the idea being that having their numbers be 3 digits would be another signal that they were A tools. This worked, but... we're gonna run out of numbers since B tools start at 1000.

So the solution is to remove that special case for A tools. Once this is merged, all new items will get the next available number in the system.

# Why it is important

I don't know what will happen if we hit 1000 for A tools, but let's not find out.

### DIFF
--- a/test/models/item_test.rb
+++ b/test/models/item_test.rb
@@ -11,6 +11,16 @@ class ItemTest < ActiveSupport::TestCase
     assert item.number
   end
 
+  test "assigns a number, with an existing item" do
+    create(:item, number: 999)
+
+    borrow_policy = create(:borrow_policy)
+    item = build(:item, number: nil, borrow_policy: borrow_policy)
+    item.save!
+
+    assert_equal 1000, item.number
+  end
+
   test "it has a due_on date" do
     loan = create(:loan, due_at: Date.tomorrow.end_of_day)
     loan.item.reload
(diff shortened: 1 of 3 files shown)


test/models/item_test.rb
```diff
@@ -11,6 +11,16 @@ class ItemTest < ActiveSupport::TestCase
     assert item.number
   end
 
+  test "assigns a number, with an existing item" do
+    create(:item, number: 999)
+
+    borrow_policy = create(:borrow_policy)
```
phinze, line 17: I think this test won't fail if we revert the change, we need to explicitly make the code B.

phinze (changes requested):
Code looks reasonable - just a couple of test issues. One comment inline and one failure in CI that looks real:

```

Error:
Admin::ItemsControllerTest#test_should_get_edit:
ActionView::Template::Error: wrong number of arguments (given 1, expected 0)
    app/models/item.rb:107:in `next_number'
    app/views/admin/items/_form.html.erb:97
    app/views/admin/items/_form.html.erb:1
    app/views/admin/items/edit.html.erb:5
```

phinze (approved):
👍

PR #1233, opened 2024-01-09
Improve devdata somewhat

jim:
# What it does

This improves a few things about how we do devdata based on a discussion at tonight's meetup:

# Why it is important

It's important that we reduce the number of sharp edges in the development environment so that we're as welcoming to new volunteers as we can be.

# UI Change Screenshot

Screenshot 2024-01-08 at 9 23 56 PM

# Implementation notes

* We now generate a lot of events, (two a day, three days a week, each week for the next year). This means that folks won't have to run `bin/reset` if it's been more than a week since they last ran it. I still think it's a good idea to run it fairly often, but what we have now is a pretty bad experience considering that contributors are likely to go more than a week while working on a feature.
* Hardcoding an encrypted password should mean that you don't get logged out when you run `bin/reset` anymore!
* I updated the Documents to contain human-friendly default values so that it's more obvious when they appear in the app. Before, for example, when signing up as a new user you'd just see an empty page because there wasn't any content set in the Documents.

### DIFF
--- a/db/devdata/action_text/rich_texts.yml
+++ b/db/devdata/action_text/rich_texts.yml
@@ -1,0 +1,21 @@
+- id: 3
+  name: body
+  body: <div>This is the waiver of indemnification. It is editable, and usually says that folks will only check out items they are confident they can use safely.</div>
+  record_type: Document
+  record_id: 3
+  created_at: 2019-07-19 14:25:59 UTC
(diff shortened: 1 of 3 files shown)


phinze (approved):
Nice! Lots of quality of life improvements in these tweaks 🥳

PR #1235, opened 2024-01-14
Fix "remove item" button in appointment

### LINKED ISSUE #989: Remove item and cancel hold buttons do the same thing
**Describe the bug**
While viewing a member's appointment as a librarian, if you click "remove item" it seems to both remove the item from the appointment AND cancel the hold. It shouldn't do that because there is also a "cancel hold" button for that. 

**To Reproduce**
1. Create a fake appointment with an item on hold for a member
2. Go to the member's appointment
3. Click remove item
4. See that the item is removed AND the hold is cancelled when you look at the member's profile

**Expected behavior**
When clicking remove item, it should just be removed from the appointment but remain as a hold for the member.  It could be reasonable to just not have one option or the other but for the moment if both buttons exist they should do what they say they should do.

**Screenshots**

**Desktop (please complete the following information):**
-Windows 10 Chrome and Windows 10 Firefox

**Additional context**
Comes up fairly infrequently and the consequences aren't usually dire so this is low priority but hopefully easy.

djcunningham0:
# What it does

Fixes #989. Previously, the "Remove Item" button removed the item from the appointment *and* canceled the member's hold on the item. Now the "Remove Item" button works as expected: it removes the item from the current appointment but does not remove the member's hold on the item.

# Why it is important

Fixes #989

# UI Change Screenshot

No UI changes.

# Implementation notes

* The `cancel_hold` parameter was being passed to the `remove_appointment_hold` method as a string rather than a boolean value, which caused the conditional statement to always evaluate as true. I fixed the issue by coercing to a boolean value, which I think should work whether the parameter is passed in as a string or a boolean. Let me know if it would be better to fix it upstream, i.e., by making sure we're passing a boolean into the function rather than a string.
* I did not create or update any unit tests. Let me know if we should add a unit test for removing an item (I would likely need some help with this).

### DIFF
--- a/app/controllers/admin/appointment_holds_controller.rb
+++ b/app/controllers/admin/appointment_holds_controller.rb
@@ -17,7 +17,7 @@ module Admin
     def remove_appointment_hold
       @remove_appointment_hold ||= appointment.appointment_holds.find_by(hold_id: params[:id]).destroy
 
-      if params[:cancel_hold]
+      if params[:cancel_hold].to_s.downcase == 'true'
         Hold.find(params[:id]).destroy
       end
     end
--- a/app/controllers/admin/appointment_holds_controller.rb
+++ b/app/controllers/admin/appointment_holds_controller.rb
@@ -17,7 +17,7 @@ module Admin
     def remove_appointment_hold
       @remove_appointment_hold ||= appointment.appointment_holds.find_by(hold_id: params[:id]).destroy
 
-      if params[:cancel_hold].to_s.downcase == 'true'
+      if params[:cancel_hold].to_s.downcase == "true"
         Hold.find(params[:id]).destroy
       end
     end
--- a/test/controllers/admin/appointment_holds_controller_test.rb
+++ b/test/controllers/admin/appointment_holds_controller_test.rb
@@ -1,0 +1,44 @@
+require "application_system_test_case"
+
+module Admin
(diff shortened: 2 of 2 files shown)


app/controllers/admin/appointment_holds_controller.rb
```diff
@@ -17,7 +17,7 @@ def destroy
     def remove_appointment_hold
       @remove_appointment_hold ||= appointment.appointment_holds.find_by(hold_id: params[:id]).destroy
 
-      if params[:cancel_hold]
+      if params[:cancel_hold].to_s.downcase == "true"
```
jim, line 20: I think this check is OK! The `downcase` is probably not necessary, but it's not going to hurt anything.

jim (commented):
Thanks for digging into this and finding the cause of the problem.

It would be nice to add a test to this. We could use a controller test (a lower level integration test that directly interacts with the controller). I don't see one in place for this controller, but we could use [this one](https://github.com/chicago-tool-library/circulate/blob/main/test/controllers/admin/appointments_controller_test.rb) as a template. I'm happy to talk about how to go about writing that test either in Slack or at the next meetup 😄.

djcunningham0:
@jim I added two tests: one with cancel_hold set to true and one with cancel_hold set to false. Let me know if there are any other tests that would be worth adding for the AppointmentHoldsController while we're looking at it.

jim (approved):
These are beautiful tests! Thanks for writing them.

PR #1244, opened 2024-01-20
Setup Appsignal to track front end errors.

jim:
# What it does

We're using [AppSignal](https://www.appsignal.com) to track server-side exceptions already. This PR sets things up so that we are notified if there are client-side errors in the app's JavaScript.

With the recent move to the latest Turbo, it seemed like a good time to get this in place in case there are any issues in the new front end code.

# Why it is important

We want to know if the JS code breaks in anyone's browser.

# Implementation notes

* This will only track errors that occur within Stimulus controllers. That is probably a good place to start.
* AppSignal recommends **not** catching all client-side errors (which is why their library doesn't by default) due to the amount of noise generated by browser extensions, ad blockers, etc.
* It wouldn't be a bad idea to consider moving our rebuild config into config files as we're passing a lot of options. I'll leave this for a future PR, though.
* I set the `APPSIGNAL_FRONTEND_KEY` environment variable on both production and staging.

### DIFF
--- a/app/javascript/controllers/application.js
+++ b/app/javascript/controllers/application.js
@@ -1,6 +1,9 @@
 import { Application } from "@hotwired/stimulus"
+import { appsignal } from "./../lib/appsignal"
+import { installErrorHandler } from "@appsignal/stimulus"
 
 const application = Application.start()
+installErrorHandler(appsignal, application)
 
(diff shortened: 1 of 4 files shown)


phinze (approved):
Good call!

Issue #1245, opened 2024-01-20
Password reset emails being marked as dangerous by Gmail and having link stripped


phinze:
**Describe the bug**
When sending a password reset to a Gmail address, the email is diverted to spam and is displayed with a warning and all links stripped out of it.

**To Reproduce**
1. Get a CTL user account with a gmail address.
2. Log out and follow the reset password flow

**Expected behavior**
Expect to receive an email in my inbox with a password reset link. Instead email shows up in spam with all links stripped.

**Screenshots**
![2024-01-20 at 16 47 28@2x](https://github.com/chicago-tool-library/circulate/assets/37534/982daf08-ef4b-4b2d-bd94-63ed78f8b060)

**Additional context**

* I checked the standard email deliver-ability stuff (SPF, DKIM, DMARC) for our domain and it all looks green
* Devise only sends HTML emails, which is currently my best guess as to why the messages are getting flagged
* **Workaround**: In my testing, clicking "Looks Safe" on the Gmail header seems to bump the message into the inbox, where the link is rendered normally.

PR #1246, opened 2024-01-20
Add text part to password reset emails

phinze:
# What it does

Hopefully this addresses the problem in #1245 - we'll need to test deliverability in production to know if it fixes it or not.

# Why it is important

We don't want any emails to be marked as spam - especially password reset emails!

# Implementation notes

* See https://github.com/heartcombo/devise/issues/5034 for some good background on this approach.


jim (approved):
Thanks for identifying and fixing this!

PR #1248, opened 2024-01-20
Group Lending: Add skeleton for reservable items, reservations, and pickups.

jim:
# What it does

This adds the first parts of group lending, with the intent that we'll continue to iterate on this code as we build out more features.

# Why it is important

Gotta start somewhere :smile:

# UI Change Screenshot

image

image

# Implementation notes

* I am trying out having breadcrumbs above the page titles on the show pages for models, like this:
    
    image
    
    We'll see if we like this and want to introduce it elsewhere.

* The new code is behind the `FEATURE_GROUP_LENDING` feature flag/environment variable. It's set to `on` locally, and we can enable it on staging once this is merged. This means that the new code won't be accessible in production for now, but the database migrations will still run there.


phinze (approved):
This is a great start! The only issue I noticed in playing with this is that it allows for same-day drop-off/pick-up. Not at all a blocker but just want to make sure it was on the TODO list.

Looks like playing around with the parens/brackets in the third argument of `tsrange` should help us to get the inclusive/exclusive bounds wrangled: https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-CONSTRUCT

IMHO that's something to chase down as a separate effort, so big ole ✅ on this!

jim:
@phinze Good call! I added an issue so we don't forget: https://github.com/orgs/chicago-tool-library/projects/1/views/4?pane=issue&itemId=51216955

PR #1255, opened 2024-01-21
Upgrade Square gem and migrate to supported APIs

### LINKED ISSUE #1254: Square Checkout v1 and Transaction APIs are deprecated
We use these APIs to implement online payments. We'll need to move to the newer recommended replacements.

jim:
# What it does

Fixes #1254 by upgrading the `square` gem and updating the app to use the newer Square Checkout APIs.

# Why it is important

The checkout v1 APIs [have been deprecated for a couple years](https://developer.squareup.com/reference/square/checkout-api/create-checkout), and while they still seem to be supported, we should get onto the newer APIs.

# Implementation notes

* The new checkout API sandbox doesn't provide the same experience, so our system tests can't run through the entire flows like they did before 😦. We can still fake our way through the process, but we can't type in payment info (and even the "preview" page doesn't accurately reflect the options passed so it's mostly useless).

# Example test

Here's one of the revised system tests going through the entire membership signup flow. You can see the part where the square sandbox returns a 404 for the checkout URL for a bit, and the test has to wait and reload the page:

https://github.com/chicago-tool-library/circulate/assets/3331/ee78e33e-d359-47f9-ae66-28ad8c9f582b


test/application_system_test_case.rb
```diff
@@ -162,6 +162,25 @@ def assert_date_displayed(datetime)
...
+    rescue Capybara::ElementNotFound
+      attempt = 1
+      if attempt > 2
+        raise
+      end
+      puts "sandbox page didn't load, trying again"
+      sleep 1
+      refresh
+    end
+  end
```
phinze, line 182: :sad_trombone:

test/lib/square_checkout_test.rb
```diff
@@ -72,29 +81,13 @@ class SquareCheckoutTest < ActiveSupport::TestCase
...
-                currency: "USD"
-              }
-            }]
-          }
-        },
-        note: "Circulate signup payment"
-      })
+    # No need to assert anything about what params are passed to
+    # this method since we do that in the successful case. There doesn't seem to
+    # be a way to assert that a mocked method is called with _any_ kwargs.
```
phinze, line 86: I just ran into this too w/ the new minitest kwargs support: you can't do _any_ kwargs but you can get almost there. The matchers run per key but they use `===` so if you know the keys you can specify keys w/ class names as values  to allow any instance of that class as a value. [Here's an example](https://github.com/chicago-tool-library/circulate/pull/1291/files#diff-c87f8805a0f4917ed3ab9cfc15eb237da4c2e3d2dcb8f658701687c8cbe64fe1R103) from a test I had to fix with the upgrade.

test/system/user_signup_test.rb
```diff
@@ -21,7 +24,7 @@ def complete_first_three_steps
     fill_in "Full name", with: "N. K. Jemisin"
     fill_in "Preferred name", with: "Nora"
     find("label", text: "she/her").click # Styled checkboxes can't be toggled using #check
-    fill_in "Email", with: "nkjemisin@test.com"
```
phinze, line 24: lol nice easter egg - 5th season was great! 📚

phinze (approved):
LGTM! You vanquished the Square dragons! 🐉

phinze:
(p.s. You successfully turned that demo video into a suspense/horror movie w/ how low the battery was 🙀🤣🔋🚨)

test/lib/square_checkout_test.rb
jim: Oh, nice. I'll give that a shot- I'd like to assert something but really want to avoid it being too specific.

Issue #1220, opened 2023-12-16
Appointment emails say "new appointment scheduled" when an appointment is modified


jim:
When someone edits an appointment, the email that we send makes it sound like a new appointment was scheduled. We should update the email we send to say "an appointment was updated" instead of "new appointment scheduled".

You can view the current email if you have the app running locally by going to http://localhost:3000/rails/mailers/member_mailer/appointment_confirmation.

jpeace-pro:
Could I claim this issue?

jim:
@jpeace-pro It's all yours!

jpeace-pro:
Resolved in [#1257](https://github.com/chicago-tool-library/circulate/pull/1257)

PR #1257, opened 2024-01-22
Create appointment updated email

jpeace-pro:
# What it does

- Added appointment_updated email template
- Changed single instance of appointment_confirmation to appointment_updated for existing appointment updates in appointment controller
- Updated integration test to expect appointment_updated for existing appointments

# Why it is important

[Issue 1220 :)](https://github.com/chicago-tool-library/circulate/issues/1220)
_"When someone edits an appointment, the email that we send makes it sound like a new appointment was scheduled. We should update the email we send to say "an appointment was updated" instead of "new appointment scheduled"."_

# UI Change Screenshot

# Implementation notes

### DIFF
--- a/app/controllers/account/appointments_controller.rb
+++ b/app/controllers/account/appointments_controller.rb
@@ -80,7 +80,7 @@ module Account
       if simultaneous_appointment
         simultaneous_appointment.merge!(@appointment)
         message = "Your existing appointment scheduled for #{helpers.appointment_date_and_time(simultaneous_appointment)} has been updated."
-        MemberMailer.with(member: @member, appointment: simultaneous_appointment).appointment_confirmation.deliver_later
+        MemberMailer.with(member: @member, appointment: simultaneous_appointment).appointment_updated.deliver_later
       else
         message = "Your appointment was scheduled for #{helpers.appointment_date_and_time(@appointment)}."
         MemberMailer.with(member: @member, appointment: @appointment).appointment_confirmation.deliver_later
--- a/app/controllers/account/appointments_controller.rb
+++ b/app/controllers/account/appointments_controller.rb
@@ -22,7 +22,7 @@ module Account
       @appointment = @member.appointments.new
 
       if @appointment.update(appointment_params)
-        merge_and_complete_appointment_update
+        merge_and_complete_appointment_update(false)
       else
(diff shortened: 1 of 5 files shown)


app/controllers/account/appointments_controller.rb
```diff
@@ -80,7 +80,7 @@ def merge_and_complete_appointment_update
       if simultaneous_appointment
         simultaneous_appointment.merge!(@appointment)
         message = "Your existing appointment scheduled for #{helpers.appointment_date_and_time(simultaneous_appointment)} has been updated."
-        MemberMailer.with(member: @member, appointment: simultaneous_appointment).appointment_confirmation.deliver_later
+        MemberMailer.with(member: @member, appointment: simultaneous_appointment).appointment_updated.deliver_later
       else
         message = "Your appointment was scheduled for #{helpers.appointment_date_and_time(@appointment)}."
         MemberMailer.with(member: @member, appointment: @appointment).appointment_confirmation.deliver_later
```
jim, line 86: The change above is correct! I think we need one more change to this line, though, to handle the case where someone is updating a single appointment but not by attempting to create a new one at the same time.

I think it's [this test](https://github.com/chicago-tool-library/circulate/pull/1257/files#diff-74dda1fc42a3fd017a9def972cb599f695f55488f20abdd8651e1fc52bc962eeR79-R89) that covers this additional case.

jim (commented):
This is looking good. I appreciate that you even put in the mailer preview 🎉

app/controllers/account/appointments_controller.rb
jpeace-pro: When you say change to this line, do you mean adding one more case like:
```
elif !simultaneous_appointment & !create_new
  MemberMailer.with....appointment_updated
```

jim: Yeah, that's exactly what I'm thinking. We could add an `updating` (or whatever) parameter to `merge_and_complete_appointment_update` and pass in `true` from the call in `#update` and `false` from the one in `#create`.

In the case where `updating` is false, we'd want to continue to send the `appointment_confirmation` email so that folks get that email when they create an appointment for the first time.

jim (approved):

PR #1260, opened 2024-01-27
CI: Update actions/checkout to use latest version

### LINKED ISSUE #1241: Address Github action warnings about node version
Our action runs have a bunch of warnings on them from using an old version of Node:

Image

I think the solution might be to upgrade the version of `checkout` that's being used.

phinze:
# What it does

Should address warnings about node in CI

Fixes #1241

# Why it is important

Warnings bad, updates good!


jim (approved):
Love to see a simple upgrade.

PR #1261, opened 2024-01-27
Get a fresh schema dump post Rails upgrades

phinze:
# What it does

This gets us a fresh dump of the schema by running `db:drop`, `db:create`, and `db:migrate`.

The most important diff here is the `precision: nil` additions, which you can read about at length here:
https://github.com/rails/rails/issues/44571

Because of this change, we can drop the special timestamp handling we had added in the interim upgrades.

# Why it is important

Keeping the `schema.rb` consistent helps us reduce odd diffs in future work that modifies the database.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -18,46 +18,46 @@ ActiveRecord::Schema[7.1].define(version: 2024_01_06_201843) do
     "fine",
     "membership",
     "donation",
-    "payment"
+    "payment",
   ], force: :cascade
 
   create_enum :adjustment_source, [
     "cash",
     "square",
-    "forgiveness"
+    "forgiveness",
   ], force: :cascade
 
   create_enum :hold_request_status, [
     "new",
     "completed",
-    "denied"
+    "denied",
   ], force: :cascade
 
   create_enum :item_attachment_kind, [
     "manual",
     "parts_list",
-    "other"
+    "other",
   ], force: :cascade
 
   create_enum :item_status, [
     "pending",
     "active",
     "maintenance",
-    "retired"
+    "retired",
   ], force: :cascade
 
   create_enum :notification_status, [
     "pending",
     "sent",
     "bounced",
-    "error"
+    "error",
   ], force: :cascade
 
   create_enum :pickup_status, [
     "building",
     "ready_for_pickup",
     "picked_up",
-    "cancelled"
+    "cancelled",
   ], force: :cascade
 
   create_enum :power_source, [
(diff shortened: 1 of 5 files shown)


jim (approved):
Awesome! It'll be nice to have this fixed up- I know the different precisions has caused some confusion.

Do we need to check the production database and verify that the precision of those columns matches that in the schema.db now?

phinze:
> Do we need to check the production database and verify that the precision of those columns matches that in the schema.db now?

Confirmed that all timestamp precisions matched by doing the following steps.

Run the following query in production and against the development database, save result as CSV:

```sql
SELECT
    pg_class.relname as table_name,
    pg_attribute.attname AS column_name,
    pg_catalog.format_type(pg_attribute.atttypid, pg_attribute.atttypmod) AS data_type
FROM
    pg_catalog.pg_attribute
INNER JOIN
    pg_catalog.pg_class ON pg_class.oid = pg_attribute.attrelid
INNER JOIN
    pg_catalog.pg_namespace ON pg_namespace.oid = pg_class.relnamespace
WHERE
    pg_attribute.attnum > 0
    AND NOT pg_attribute.attisdropped
    AND pg_namespace.nspname = 'public'
    AND pg_catalog.format_type(pg_attribute.atttypid, pg_attribute.atttypmod) like '%time%'
ORDER BY
    1,2;
```

Run diff against CSV:

```diff
--- development.csv     2024-01-29 10:25:46
+++ dataclips_svqbhbbyewyoktagcgbyyrahiqqa.csv  2024-01-29 10:24:05
@@ -25,8 +25,6 @@
 categories,updated_at,timestamp(6) without time zone
 categorizations,created_at,timestamp(6) without time zone
 categorizations,updated_at,timestamp(6) without time zone
-date_holds,created_at,timestamp(6) without time zone
-date_holds,updated_at,timestamp(6) without time zone
 documents,created_at,timestamp(6) without time zone
 documents,updated_at,timestamp(6) without time zone
 events,created_at,timestamp(6) without time zone
@@ -68,17 +66,9 @@
 notes,updated_at,timestamp(6) without time zone
 notifications,created_at,timestamp(6) without time zone
 notifications,updated_at,timestamp(6) without time zone
-pickups,created_at,timestamp(6) without time zone
-pickups,updated_at,timestamp(6) without time zone
+pg_stat_statements_info,stats_reset,timestamp with time zone
 renewal_requests,created_at,timestamp(6) without time zone
 renewal_requests,updated_at,timestamp(6) without time zone
-reservable_items,created_at,timestamp(6) without time zone
-reservable_items,updated_at,timestamp(6) without time zone
-reservations,created_at,timestamp(6) without time zone
-reservations,ended_at,timestamp(6) without time zone
-reservations,reviewed_at,timestamp(6) without time zone
-reservations,started_at,timestamp(6) without time zone
-reservations,updated_at,timestamp(6) without time zone
 short_links,created_at,timestamp(6) without time zone
 short_links,updated_at,timestamp(6) without time zone
 ticket_updates,created_at,timestamp(6) without time zone
```

This looks good as the only diff is coming from:
 * pg_stat_statements being on in production
 * group lending feature flag being on in development

We have sort of a weird collection of `timestamp(6)` and `timestamp` data types, but it doesn't really matter in practice and they match in dev and prod so I think we're good!

PR #1262, opened 2024-01-27
Improve readability of member appointment page

### LINKED ISSUE #1236: Improve readability of member appointment page
Here's the way appointments currently appear to members:

Image

There are some issues with this:
* The line height of the date is too small (and probably, the date itself is too large)
* The list of items are right-aligned. They should be left-aligned.

At some point this screen might have been intended to have two columns (like the screen where you select the items while creating or editing an appointment). That's OK, but it's also OK to just make it one list above another list.

Notes

To work on this, you'll likely want to run `bin/reset` and then login in as `member_with_appointment@example.com` to see an upcoming appointment. To get images to appear for the items in that appointment, you'll need to log in as `admin@example.com` and add them to those items. I like to have two browser windows open to do this kind of thing (one normal, on in incognito or private mode so that I can be logged in as two different users at the same time).

djcunningham0:
# What it does

Fixes #1236. Readability improvements to the member appointment page. It fixes the points called out in the issue, plus a few other small changes.

* Decrease the font size of the date and time
* Move images to the left of descriptions
* Remove horizontal lines between items
* Add vertical margin between the "Appointments" section and the "How do appointments work" section when viewed on small screens (e.g., phones). There was already a horizontal margin when viewed on a wide screen.

This PR also adds images to all items in the development version of the app. The same image is used for all items.

# Why it is important

#1236 

# UI Change Screenshot

Before:

image

After:

image

# Implementation notes

The main issue was a div tag that wasn't closed properly. That caused the date/time header to flow directly into the card body rather than starting a new line.

### DIFF
--- a/app/assets/stylesheets/account.scss
+++ b/app/assets/stylesheets/account.scss
@@ -6,6 +6,9 @@
     @media all and (min-width: 841px) {      
       margin-left: 2rem;
     }
+    @media all and (max-width: 840px) {
+      margin-top: 2rem;
+    }
     ul {
       list-style: outside;
     }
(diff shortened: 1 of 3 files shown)


jim (approved):
Awesome! This screen looks much better now.

Issue #1247, opened 2024-01-20
Internal server error when loaning out items that are in maintenance


phinze:
**Describe the bug**

When trying to hold or lend items in maintenance, the app generates an error:

```
undefined method `persisted?' for false:FalseClass

      if @loan.persisted?
              ^^^^^^^^^^^
```

Coming from `app/controllers/admin/loans_controller.rb:21 create`

**To Reproduce**
1. As an admin, attempt to loan an item that is in maintenance

**Expected behavior**
1. Items in maintenance should produce a validation error when attempting to hold or loan them, not an error page. 

**Additional context**
* Appsignal page for the error: https://appsignal.com/chicago-tool-library/sites/60596f9214ad662e17191689/exceptions/incidents/225

phinze:
I think to fix this we should make the [create_loan method](https://github.com/chicago-tool-library/circulate/blob/main/app/controllers/concerns/lending.rb#L7-L14) always return the `Loan` object even if it fails to save. That way the LoansController can properly check `persisted?` and pass along the error messages to the UI.

jpeace-pro:
Resolved in [PR #1263](https://github.com/chicago-tool-library/circulate/pull/1263)!

PR #1263, opened 2024-01-28
Issue 1247 create_loan bug

jpeace-pro:
# What it does

- Change return of create_loan method to always return loan so loans_controller persisted? logic handles items appropriately
- Create test for creating loans for items in maintenance scenario

# Why it is important

[Issue 1247](https://github.com/chicago-tool-library/circulate/issues/1247) addresses issue where undefined values were being passed and causing internal server errors.

# UI Change Screenshot

Close your eyes and imagine a HTTP 500 slowly fading away~

# Implementation notes

* _Anything notable about the technical approach you took._
No

* _Any open questions you have about the PR or areas where you'd like specific feedback._
I have not yet reproduced the error logged in as an admin or verified_member!! I'm new to the app, so I'm not sure how to create a loan on an item in maintenance because I don't see the button to do it for example on item B-1341 Airbrush kit. Need some help testing this change will work.

### DIFF
--- a/app/controllers/concerns/lending.rb
+++ b/app/controllers/concerns/lending.rb
@@ -4,14 +4,10 @@ module Lending
   def create_loan(item, member, now: Time.current)
     loan = Loan.lend(item, to: member, now: now)
     loan.transaction do
-      if loan.save
-        if item.borrow_policy.consumable? && return_loan(loan, now: now)
-          item.decrement_quantity
-        end
-        loan
-      else
-        false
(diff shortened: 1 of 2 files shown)


phinze (approved):
The change and the test look perfect! 🥳 I checked this out to try yanking the code change and the test produces exactly the error message we observed. ✅ 

I can help walk you through how to reproduce the issue in the app at our next meeting.

PR #1276, opened 2024-01-30
Fix flaky RemoveHoldsTest cases

### LINKED ISSUE #1275: Flaky system tests in RemoveHoldsTests
**Describe the bug**

If you review the recent trends in our system tests there are a couple of tests that sporadically fail.

One ([example](https://github.com/chicago-tool-library/circulate/actions/runs/7716605411/job/21033796291#step:12:63)):

```
Failure:
RemoveHoldsTest#test_removes_hold_and_cancels_associated_appointment_if_it's_the_only_item [/home/runner/work/circulate/circulate/test/system/remove_holds_test.rb:49]:
expected to find text "You have no scheduled appointments" in "Skip to content\nHome View Inventory Loans Holds Appointments Membership Logout\nHolds\nView Hold History\nYou have no items on hold\nHow do holds work?\nPlacing a hold means you are putting yourself in line to borrow an item. Please do not do this unless you plan on borrowing the item.\nIf you are first in line you have one week to pick up an item. If you do not pick up the item, you will lose your position in line and another person may borrow it.\nIf another person is borrowing the tool when you place your hold you will receive an email when it is your turn to borrow the item. You then have one week to pick up the item.\nIf you no longer need to borrow an item you can delete your hold to remove yourself from the line.\nFor questions and issues, please email team@chicagotoollibrary.org."
```

In testing locally I also regularly saw this other failure:

```
RemoveHoldsTest#test_removes_hold_from_associated_appointment_if_appointment_has_more_than_one_item:
Capybara::ElementNotFound: Unable to find css "li.appointment"
    test/system/remove_holds_test.rb:33:in `block in <class:RemoveHoldsTest>'`
```

**To Reproduce**

Run the test file in a loop until it fails:

```bash
while ./bin/rails test test/system/remove_holds_test.rb; do echo OK; date; done
```

**Expected behavior**
Tests always pass

**Observed behavior**
Tests sometimes fail

phinze:
Closes #1275

# What it does

The tests seem to be failing because the `click_on "Appointments"` action is not working after the `accept_confirm`. This changes the `click_on` to be more conservative and adds an assertion to make the error message clear when that's the step that's failing. It also adds a short sleep between the `accept_confirm` and the `click_on` which in my testing seemed to address the flakiness of these two tests.

# Why it is important

Flaky tests bad. Tests should reliably pass!

# Implementation notes

I know that adding sleeps to address flaky Capybara tests is generally discouraged, but I feel like I exhausted all our other options in this investigation (barring something more drastic like moving away from javascript popups - I did find [one article suggesting they should be avoided](https://medium.com/@bsemaj/6-tips-to-get-your-flaky-capybara-tests-passing-a97817af1aec) because they're a source of flaky capybara behavior).

### DIFF
--- a/test/system/remove_holds_test.rb
+++ b/test/system/remove_holds_test.rb
@@ -28,7 +28,14 @@ class RemoveHoldsTest < ApplicationSystemTestCase
 
     visit account_holds_path
     accept_confirm { click_on "Remove Hold" }
-    click_on "Appointments"
+
+    # This sleep seems to quell a flaky failure with Capybara clicking on the
+    # Appointments link after the accept_confirm above
+    sleep 1
+    within("header.navbar") do
+      click_on "Appointments"
+    end
+    assert_current_path account_appointments_path
 
     appointment_card = find("li.appointment")
     within(appointment_card) do
@@ -44,7 +51,14 @@ class RemoveHoldsTest < ApplicationSystemTestCase
 
     visit account_holds_path
     accept_confirm { click_on "Remove Hold" }
-    click_on "Appointments"
+
+    # This sleep seems to quell a flaky failure with Capybara clicking on the
+    # Appointments link after the accept_confirm above
+    sleep 1
+    within("header.navbar") do
+      click_on "Appointments"
+    end
+    assert_current_path account_appointments_path
 
     assert_text "You have no scheduled appointments"
   end


jim (approved):
I'm of the same mind about adding sleeps, but if it works, it works.

PR #1288, opened 2024-02-06
Remove unused smooth-scroll-into-view package

phinze:
# What it does

Removes an unused dependency, as discussed in https://github.com/chicago-tool-library/circulate/pull/1272

# Why it is important

Less code makes maintenance easier!


jim (approved):
The fewer the merrier!

PR #1291, opened 2024-02-06
Update minitest and fix tests

phinze:
# What it does

Updates minitest and changes tests to pass in the newer version

# Why it is important

Staying on the latest version of our dependencies is good

# Implementation notes

Minitest's interface changed slightly when they added first class support for kwargs, see https://github.com/minitest/minitest/issues/912

Closes #1269

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -85,7 +85,7 @@ end
 
 group :test do
   # Adds support for Capybara system testing and selenium driver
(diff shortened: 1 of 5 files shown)


jim (approved):
Nice! Thanks for addressing those changes in the tests so we could stay up to date.

PR #1309, opened 2024-02-07
Add an improved test tagging system.

jim:
# What it does

In the process of upgrading the Square library, I found myself working with our test tagging a bunch and found it rather lacking. I looked for something off-the-shelf, but [the closest thing I could find](https://github.com/ordinaryzelig/minispec-metadata) only works with minitest-spec (which we aren't using).

The [README changes](https://github.com/chicago-tool-library/circulate/pull/1309/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R181) describe how this works fairly well, but here's a high level summary:

* Tests are tagged by passing a list of symbols after the test name, like `test "the app works", :remote do .... end`.
* We don't run tests tagged with `remote` by default.
* I based the syntax for tags off [how Rspec does it](https://rspec.info/features/3-12/rspec-core/command-line/tag/), albeit in a very simplified form.
    * You can use `-t` or `--tag` to scope the tests that are run to a specific tag, like `bin/rails test -t remote` will only run remote tests.
    * You can also negate tags to not run them using `~`. If we didn't negate remote tests by default, you could do the same thing with `bin/rails test -t ~remote`.
    * You can clear the settings for a tag using `:`. So to run all tests (and ignore the default negation of remote-tagged tests), you'd run `bin/rails test -t :remote`.
    * To run, tests have to match at least one required tag but none of the negated tags.

# Why it is important

As we're adding more remote tests and different types of tests, being able to run a specific subset will be very useful.

I started to mess with putting remote tests in a different directory, but then I had some lib tests and some system tests, and then I needed a way to share code between the remote and normal tests for the same class now that they were split apart. By leveraging tags, we can co-locate the test code when needed but still have control over what runs when.

### DIFF
--- a/test/minitest/tags_plugin.rb
+++ b/test/minitest/tags_plugin.rb
@@ -1,0 +1,118 @@
+module Minitest
+  # This is based roughly on the way tag filtering is done by Rspec.
+  #
+  # Tags can be either required or negated.
+  #
+  # If any tags are required, tests much match at least one of the provided
+  # tags. Required tags have no prefix.
+  #
+  # If any tags are negated, tests matching any negated tag will not run. Tags
+  # can be negated with ` tilde (`~`) prefix.
+  #
+  # Tests tagged with `remote` are negated by default.
+  #
+  # To clear a tag (make it neither required or negated), a colon (`:`) can
+  # be used as a tag prefix.
+  class TagFilter
+    def initialize
+      @required_tags = []
+      @negated_tags = []
+    end
+
+    def require_tag(name)
+      name_symbol = name.to_sym
+      @negated_tags.delete_if { |k| k == name_symbol }
+      @required_tags << name_symbol
+    end
+
+    def negate_tag(name)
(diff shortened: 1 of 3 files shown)


test/minitest/tags_plugin.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  # Override test definition to allow for extra arguments. These become the tags.
+  def test(subject, *tags, &block)
+    super(subject, &block).tap do |method|
+      method_tags[method.name] = tags
+    end
+  end
+end
+
+ActiveSupport::TestCase.send :extend, TestCasePatch
```
phinze, line 118: I don't think [`Object#extend`](https://ruby-doc.org/3.2.2/Object.html#method-i-extend) is private, so we don't need to `send` here, we can just call extend directly.

jim: Nice catch!

PR #1310, opened 2024-02-07
Update selectize

phinze:
# What it does

Updates selectize package.

It looks like selectize moved from https://www.npmjs.com/package/selectize to https://www.npmjs.com/package/@selectize/selectize which is why we were stuck on an old version with a security warning.

# Why it is important

This should clear out our remaining security warnings.


phinze:
Tested tag selector and still works:

2024-02-07 at 10 52 33@2x

jim (approved):

PR #1324, opened 2024-02-09
Update spy

phinze:
# What it does

Looks like they fixed a bug where the class was getting passed as an argument to a stub block, so I had to update my janky test.

Also loosened the rescue in ActivityNotifier so it doesn't swallow things like ArgumentError which it was doing here.

# Why it is important

Updates!


jim (approved):
Another one bites the dust!

PR #1327, opened 2024-02-10
Add ItemPools and update reservations accordingly.

jim:
# What it does

Adds the concept of an item pool and item pile (😎) to the app. This allows groups members to reserve a quantity of an item pool ("5 cordless drills"), and also to have item pools that don't have uniquely numbered items (e.g. work gloves).

# Why it is important

This is the next stage of building out group lending. Once this is merged, the main data models to represent items and reservations will be in place. Next up is how we handle pickups and actually checking out items.

# Implementation notes

* Some of the database querying code isn't well factored yet.


app/models/item_pool.rb
```diff
@@ -0,0 +1,71 @@
...
+    if totals_by_day.empty?
+      total_items
+    else
+      total_items - totals_by_day.values.max
+    end
+  end
+
+  private
+
+  # Return a Hash of date => total reservations for the given range.
```
jim, line 45: There is some duplication in this file still. This method I feel pretty good about, the other ones that wrap it I'm not as sure about. I figure that stuff will get ironed out as we start to build out real UIs for the new flows.

app/lib/spectre_form_builder.rb
```diff
@@ -137,6 +138,24 @@ def date_field(method, options = {})
     end
   end
 
+  # Render a set of fields for every object in an association, along with a
+  # button to create new instances of the join model.
+  #
+  # Based on the "Stimulus" example from this absolute legend:
+  # https://stackoverflow.com/a/71715794
```
phinze, line 145: Wow that's one of the most thorough SO posts I've seen in a while. 🤩

db/migrate/20240214190849_add_reservable_items_count_to_item_pools.rb
```diff
@@ -0,0 +1,5 @@
+class AddReservableItemsCountToItemPools < ActiveRecord::Migration[7.1]
+  def change
+    add_column :item_pools, :reservable_items_count, :integer, default: 0, null: false
```
phinze, line 3: ❓How is this field maintained? I see one reference to it in `ItemPool#max_reservable_quantity` but I don't see anything manipulating the field.

(Pretty sure this is working and it's just an issue with my understanding, so consider this non-blocking.)

test/models/item_pool_test.rb
```diff
@@ -0,0 +1,115 @@
...
+    reservation = create(:reservation, started_at: @starts, ended_at: @ends)
+    create(:date_hold, reservation: reservation, item_pool: item_pool, quantity: 1)
+
+    assert_equal 2, reservable_item.item_pool.max_available_between(@starts, @ends)
+  end
+
+  test "handles multiple reservations that don't overlap" do
+    # reservation 1  |-------|
+    # reservation 2            |-------|
+    # query               |-------|
```
phinze, line 39: These are a great idea and help a lot with understanding - thanks for including them!

phinze (approved):
This is coming along really nicely! 👏🏻 🥳

app/lib/spectre_form_builder.rb
jim: It must have taken soooo long to write up!

db/migrate/20240214190849_add_reservable_items_count_to_item_pools.rb
jim: It's Rails Magic™! By specifying [`counter_cache: true` in the association declaration within `ReservableItem`](https://github.com/chicago-tool-library/circulate/pull/1327/files#diff-a99705a6be1ae926d7cdf2f6713144c1ae48013c65d1f029c97d50d9538ee2c2R3), Rails will [automatically update this count for us](https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache).

As soon as we need to filter which associated items are counted (say, maybe we only want to count items with a certain status) we will have to implement it ourselves. But for now, this will do.

PR #1337, opened 2024-02-13
SMS Hold Available Notifications

phinze:
# What it does

Adds text message notifications when a hold becomes available.

# Why it is important

Hopefully this will help users who might not check email often remember to pick up their holds.

# UI Change Screenshot

(coming soon)

# Implementation notes

* I couldn't find any rake task testing patterns in the app so I wrapped a simple test around the `holds:start_waiting_holds` test using tips from this article: https://josh-works.medium.com/testing-rake-tasks-in-rails-6573f7185a0a 
* This starts to consolidate feature flag lookups into a centralized class. I'll follow up with a pass to update the other places we have feature flags.
* In #1336 I have SMS segment length as a constant, I'll update the magic number in the test to refer to that once that lands

### DIFF
--- a/app/lib/feature_flags.rb
+++ b/app/lib/feature_flags.rb
@@ -1,0 +1,5 @@
+class FeatureFlags
+  def self.sms_reminders_enabled?
+    ENV["FEATURE_SMS_REMINDERS"] == "on"
+  end
+end
--- a/app/lib/activity_notifier.rb
+++ b/app/lib/activity_notifier.rb
@@ -17,7 +17,7 @@ class ActivityNotifier
     each_member(members_with_overdue_items) do |member, summaries|
       summaries = summaries.overdue_as_of(@now.tomorrow.beginning_of_day)
       MemberMailer.with(member: member, summaries: summaries, now: @now).overdue_notice.deliver
-      if sms_reminders_enabled?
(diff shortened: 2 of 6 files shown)


app/lib/feature_flags.rb
```diff
@@ -0,0 +1,5 @@
+class FeatureFlags
```
jim, line 1: Big fan of adding this so we're not accessing env vars all over the place (and so there's a place to see what the feature flags currently are).

jim (approved):
Code looks great!

> Open question: This currently sends one text message per available hold... could this be spammy for users that place a lot of holds? Do lots of holds often become available at once?

I think it's fairly unlikely that it will be a huge problem, but it's a good thing to consider. We could look at the notifications table and see how often people are getting the email notification more than once in a given day to see how often this situation might come up.

phinze:
> I think it's fairly unlikely that it will be a huge problem, but it's a good thing to consider. We could look at the notifications table and see how often people are getting the email notification more than once in a given day to see how often this situation might come up.

Took a look at this and it looks like it's somewhat of an issue, but probably worth tackling holistically instead of specifically for text messages. Filed as https://github.com/chicago-tool-library/circulate/issues/1349

PR #1346, opened 2024-02-18
Upgrade standard to 1.34

jim:
# What it does

Upgrades standard and fixes a few issues from a new default setting. It promotes the use of [anonymous block arguments forwarding](https://www.writesoftwarewell.com/anonymous-block-forwarding-in-ruby/), which... 🤷🏻 

Standard only found some of the uses of `&block` in the codebase, which isn't great. It looks like there are some issues with the changes in Ruby 3.3 that may be trickling down here based on [Naming/BlockForwarding with anonymous block parameter (Ruby 3.3.0) ](https://github.com/rubocop/rubocop/issues/12571)

We might want to wait on this one to let some of these issues get shaken out.


phinze:
+1 to hold off on this one - i think we can tell dependabot to skip a version

phinze:
Ignored in #1341 and we can let the next update swing around 👍🏻

PR #1350, opened 2024-02-20
Update spectre to point to fork

phinze:
# What it does

The Spectre CSS Framework had stagnated under its original maintainer and some community members forked it to an org, which looks better maintained. https://github.com/spectre-org

# Why it is important

They fixed the sass incompatibilities, so we can drop the sass-migrator workaround.


jim (approved):
Nice! Hopefully this means we don't have to eventually migrate to something else 🤞.

PR #1366, opened 2024-02-22
Add membership_type to Membership.

jim:
# What it does

We derive whether a membership is the first for a person or a renewal in at least one database view, but having this stored on the model directly will make it a lot easier to query for this data and to do simple data export operations without loading a member's entire history.

# Why it is important

We got a request from staff to include this in data exports, and it seemed like it would be easy enough to start storing the value in a way that is much simpler to access.

# Implementation notes

* The second migration that updates value is naive, but for the amount of data we have (thousands of members, each with less than ten memberships) it should be OK. The migrations ran again production data in about 7 seconds on my machine despite not being optimized:

    ```console
    $ time bin/rails db:migrate
    == 20240222032301 AddMembershipTypeToMembership: migrating ====================
    -- create_enum(:membership_type, ["initial", "renewal"])
       -> 0.0018s
    -- add_column(:memberships, :membership_type, :enum, {:null=>false, :enum_type=>"membership_type", :default=>"initial"})
       -> 0.0016s
    == 20240222032301 AddMembershipTypeToMembership: migrated (0.0035s) ===========
    
    == 20240222034916 AddTypesToExistingMemberships: migrating ====================
    == 20240222034916 AddTypesToExistingMemberships: migrated (4.8780s) ===========
    
    ________________________________________________________
    Executed in    7.32 secs    fish           external
       usr time    4.49 secs  148.00 micros    4.49 secs
       sys time    0.64 secs  345.00 micros    0.64 secs
    ```


db/migrate/20240222034916_add_types_to_existing_memberships.rb
```diff
@@ -0,0 +1,25 @@
+class AddTypesToExistingMemberships < ActiveRecord::Migration[7.1]
+  class MigrationMember < ActiveRecord::Base
+    self.table_name = "members"
+    has_many :memberships, class_name: "MigrationMembership", foreign_key: "member_id"
+  end
```
phinze, line 5: I like this pattern of defining a minimal model for the purposes of migration right inline. Helps protect against code collisions down the line. 👏🏻

phinze (approved):
This looks straightforward and seems like an easy win for reporting 👍🏻

Issue #1312, opened 2024-02-07
As a user, I am getting routed to the wrong page after renewals


jpeace-pro:
**Describe the bug**
On renewing loaned items, if an item is allowed to be renewed without librarian approval, I get routed to a different page instead of back to the loaned items. If I want to renew another item after renewing one of these types of items, I have to reroute myself back to the loaned items page.

**To Reproduce**
1. Go to 'Loans'
2. Click on 'Renew Loan' on an item that does not require librarian approval
3. See routing error

**Expected behavior**
I am rerouted to the account/loans page. Not the account page.

**Screenshots**
This is where I'm being routed:
![image](https://github.com/chicago-tool-library/circulate/assets/25113003/76f7f4bd-a18f-4269-bfe4-ca83a8466339)
This is where I want to be routed:
![image](https://github.com/chicago-tool-library/circulate/assets/25113003/d2313c2d-e9c8-41dd-a89f-386b2b11b650)

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser: Chrome
 - Version: 121.0.6167.141

**Smartphone (please complete the following information):**

**Additional context**

jim:
Good call, it would be great to improve this interaction.

PR #1385, opened 2024-03-04
Issue 1312 Change loan renewal request routing

jpeace-pro:
# What it does

Addresses [issue 1312!](https://github.com/chicago-tool-library/circulate/issues/1312)

# Why it is important

Makes renewing multiple items more seamless!
Users were being routed back to their account home page instead of back to the renewal page for items that could be renewed without librarian consent (as opposed to renewal requested items that were routing users back to the loans page).
It gives the two types of renewals the same behavior. 

Whether or not they should be ideologically separate is a whole other discussion... :P

# UI Change Screenshot

# Implementation notes

- [x] I don't have time or interest in making additional changes to this work.

### DIFF
--- a/app/controllers/account/renewals_controller.rb
+++ b/app/controllers/account/renewals_controller.rb
@@ -7,7 +7,7 @@ module Account
       authorize @loan, :renew?
 
       renew_loan(@loan)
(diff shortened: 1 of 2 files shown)


jim (approved):
Thanks for finding and then fixing this!

PR #1388, opened 2024-03-11
Add items to pickups for group reservations

jim:
# What it does

Extends the "pickup" page to allow for specific items to be added as they are pulled from the shelf and added to the order.

# UI Change Screenshot

# Implementation notes

* `ReservationLoan` objects belong to either a `ReservableItem` for uniquely numbered items (item pools) or an `ItemPool` for non-uniquely numbered items (item pile). I can be convinced that there is a better way to model this.
* The use of `autofocus` to keep the item ID input accepting keyboard input is a temporary solution. We'll likely want something slightly more involved eventually. It's a bad practice from an accessibility standpoint; however, the main use case of the page is someone who has a barcode scanner and is going to scan **a lot** of items. Our current system sometimes loses focus and folks will scan for a while without realizing that it's not doing anything. So having some kind of focus control here is necessary.
* I am convinced we should rename `Pickup` to `Fulfillment`, but I don't want to do that in this PR as there are already a lot of changes.


app/controllers/admin/pickups/reservation_loans_controller.rb
```diff
@@ -0,0 +1,83 @@
+module Admin
+  module Pickups
+    class ReservationLoansController < BaseController
+      before_action :set_reservation_loan, only: :destroy
+
+      # There are two wolves inside this method. If :date_hold_id is passed, it means
```
phinze, line 6: 🌔🐺🐺

phinze (approved):
Like you suggested in your PR description, I'm also squinting a bit at the shape of the modeling. I'm not coming up with any better alternatives though so I think this is a good next step and we can see how it all settles in.

PR #1393, opened 2024-03-11
Update dotenv and use it directly.

jim:
# What it does

Updates us to dotenv 3, and moves to using that gem directly instead of `dotenv-rails` which is [no longer necessary](https://github.com/bkeepers/dotenv/pull/468).

This effectively the same as #1378, but moves to using `dotenv` as a direct dependency.


phinze (approved):
Nice LGTM

PR #1394, opened 2024-03-11
nix: update dependencies for postgres client

phinze:
# What it does

In the nix development environment, this matches the postgres client version with heroku's version (15.6), which only generates warnings during normal usage but when doing pg:pull and pg:push operations, pg_dump will complain about version mismatches.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -21,11 +21,11 @@
         "systems": "systems"
       },
       "locked": {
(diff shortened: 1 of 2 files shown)


phinze:
(Gonna self-merge this since I'm the only one using the nix setup at the moment.)

PR #1406, opened 2024-03-15
Tell Dependabot to do updates in weekly batches

phinze:
# What it does

Reconfigure dependabot to run weekly instead of daily, and have it batch updates into a few groups.

# Why it is important

Dependabot has proven to be moderately noisy when running daily, so I think it's worth trying to see if batching feels easier to manage.

# Implementation notes

* There are a lot of tools for deciding what gets batched together. As I start I tried splitting out rails and linter updates from others, but it'll be easy to tweak these as we learn more.
* Notably [dependabot cannot batch updates across package ecosystems yet](https://github.com/dependabot/dependabot-core/issues/8126), so rails upgrades will still likely need to be done manually so we sync the npm and ruby packages.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,8 +8,36 @@ updates:
   - package-ecosystem: "bundler"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
+    groups:
+      deps:
+        patterns:
+          - "*"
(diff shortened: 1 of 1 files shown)


jim (approved):
This is great; let's see if it's better for us. The daily updates create a lot of PRs.

PR #1407, opened 2024-03-15
Send welcome SMS to users who opt into text reminders

phinze:
# What it does

Members get an initial welcome message from the system when they have opted into text reminders for the first time and when they update their phone number.

# Why it is important

This should help inform users about what number future reminders will come from and also how to opt-out if they didn't mean to get them.

# UI Change Screenshot

2024-03-15 at 14 45 44@2x

# Implementation notes

* I noticed all our SMS body strings were ending in a newline, so I fixed that and added tests to verify we don't do it in the future. I believe Twilio was stripping these out but I figure it's safer for us to do so ourselves.

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -129,6 +129,9 @@ class Member < ApplicationRecord
 
   def send_welcome_text
     MemberTexter.new(self).welcome_info
+  rescue RuntimeError => e
+    Rails.logger.error("Error notifying member #{id}: #{e}")
+    Appsignal.send_error(e)
   end
 
   def update_user_email
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -56,6 +56,9 @@ class Member < ApplicationRecord
   before_validation :set_default_address_fields
   before_validation :downcase_email
 
+  after_save :send_welcome_text, if: -> {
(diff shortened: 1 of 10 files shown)


app/models/member.rb
```diff
@@ -56,6 +56,9 @@ class Member < ApplicationRecord
   before_validation :set_default_address_fields
   before_validation :downcase_email
 
+  after_save :send_welcome_text, if: -> {
+    reminders_via_text? && (saved_change_to_phone_number? || saved_change_to_reminders_via_text?)
```
jim, line 60: Will `saved_change_to_reminders_via_text?` return true if someone has text messages turned on but then disables them?

jim (approved):
Looks good to me. Solid testing improvements. I left one question about when we send the welcome message, but I don't consider it blocking.

app/models/member.rb
phinze: Yep the first clause should ensure that uncheck and save does not text. There's a clause in the exercising test here: https://github.com/chicago-tool-library/circulate/pull/1407/files#diff-43b9c2473f1b9fb4712413041dbe1968d063150039a6d4b7b38810e5bcf880b6R224-R227

jim: Awesome. Ship it!

Issue #1259, opened 2024-01-27
Member number has globally unique index but numbers are assigned unique per library tenant


phinze:
**Describe the bug**

The `Member#assign_number` method is automatically scoped to the current library by `acts_as_tenant`, but there's a globally unique database index on that column. This means that a number can be assigned that's unique for the current library but has a duplicate in the database, resulting in an index violation and a server error. This is not likely to be something that we hit in production, but it's an issue with the multi-tenancy code that we should fix.

**To Reproduce**
1. In a development server, log in as `admin@example.com`
2. Go to `new_member@example.com` in the admin
3. Click `Verify Info` and complete the verification screen

**Expected behavior**
Verification completes, member number is assigned.

**Observed behavior**
Unique index violation.

**Screenshots**
![2024-01-27 at 14 19 16@2x](https://github.com/chicago-tool-library/circulate/assets/37534/f91425ce-fc8b-4f2c-975a-e976c4e02298)

**Additional context**

 * In [this commit](https://github.com/chicago-tool-library/circulate/commit/e7ae1792237af871992090a0e324a167e29b66ae) member numbers were scoped to be per-library-tenant.
 * However, there's still a [globally unique index on member number column](https://github.com/chicago-tool-library/circulate/blob/f68289725689732c4a608f03241c125ac551b3af/db/schema.rb#L443).
 * This should be changed to be a unique index on the combination of library_id and number.

jim:
@jpeace-pro I inadvertently fixed this while addressing a different issue in #1415!

Issue #1345, opened 2024-02-18
New members should be listed as "pending" and not "ended"


jim:
When I create a new member account and pay for a membership through square, my member appears in the admin as "ended":

![Image](https://github.com/chicago-tool-library/circulate/assets/3331/d26c0b06-98f2-4865-a8a6-36677279bd4a)

This should really say "pending", since it's a new member who hasn't done anything yet.

jim:
Fixed by #1415

PR #1415, opened 2024-03-19
Clean up member listing status column

jim:
# What it does

Two things, somewhat related:

* Revises how the membership status is displayed in the member list. As of right now it shows some nonsensical things for new members in certain states (see #1345).
* Fixes the member number uniqueness constraint to be scoped to libraries. It's not currently, which means that some of the members created by dev data are broken and can't be verified or have memberships started. This isn't an issue in production, but we may as well fix it.

# UI Change Screenshot

This is the right three columns for the same set of members on the admin member listing.

Before:
Screenshot 2024-03-18 at 11 25 21 PM

After:
Screenshot 2024-03-18 at 11 25 14 PM

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.1].define(version: 2024_03_03_164619) do
+ActiveRecord::Schema[7.1].define(version: 2024_03_14_004211) do
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -452,9 +452,8 @@
     t.text "pronouns", default: [], array: true
     t.integer "library_id"
     t.string "pronunciation"
+    t.index ["library_id", "number"], name: "index_members_on_library_id_and_number", unique: true
     t.index ["library_id"], name: "index_members_on_library_id"
-    t.index ["number", "library_id"], name: "index_members_on_number_and_library_id"
-    t.index ["number"], name: "index_members_on_number", unique: true
```
jim, line 457: We always scope queries on `library_id`, so this index is redundant.

phinze (approved):
Nice - Good improvements, and I believe the index change should fix #1259

PR #1418, opened 2024-03-21
Fix error guarding in text message code

phinze:
# What it does

Turns out rescuing from RuntimeError is _not_ what you want to do, since TwilioErrors all descent from StandardError but not RuntimeError. Whoops!

# Why it is important

This should prevent any future errors from Twilio from stopping the notification train.

# Implementation notes

* Note that standardrb wants us to omit the error class entirely when rescuing from StandardError, as the bare `rescue` is equivalent to that.

### DIFF
--- a/test/lib/activity_notifier_test.rb
+++ b/test/lib/activity_notifier_test.rb
@@ -213,7 +213,7 @@ class ActivityNotifierTest < ActiveSupport::TestCase
       3.times { create(:loan, member: create(:verified_member, reminders_via_text: true), due_at: Time.current.end_of_day, created_at: 1.week.ago) }
 
       # Make MemberTexter.new explode one time then work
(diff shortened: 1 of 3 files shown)


jim (approved):

test/lib/activity_notifier_test.rb
```diff
@@ -213,7 +213,7 @@ def black_hole.respond_to_missing?(*args)
       3.times { create(:loan, member: create(:verified_member, reminders_via_text: true), due_at: Time.current.end_of_day, created_at: 1.week.ago) }
 
       # Make MemberTexter.new explode one time then work
-      error = RuntimeError.new("oh no")
+      error = Twilio::REST::TwilioError.new
```
jim, line 216: Using the actual error here is a good call.

Issue #1185, opened 2023-11-10
Use nodejs buildpack to address deploy warning


phinze:
Heroku builds include the following warnings:

```
###### WARNING:
       Installing a default version (20.9.0) of Node.js.
       This version is not pinned and can change over time, causing unexpected failures.
       
       Heroku recommends placing the `heroku/nodejs` buildpack in front of
       `heroku/ruby` to install a specific version of node:
       
       https://devcenter.heroku.com/articles/ruby-support#node-js-support
###### WARNING:
       Installing a default version (1.22.19) of Yarn
       This version is not pinned and can change over time, causing unexpected failures.
       
       Heroku recommends placing the `heroku/nodejs` buildpack in front of the `heroku/ruby`
       buildpack as it offers more comprehensive Node.js support, including the ability to
       customise the Node.js version:
       
       https://devcenter.heroku.com/articles/ruby-support#node-js-support
```

We should follow their recommendations!

jim:
Fixed by #1432.

PR #1432, opened 2024-04-01
Specify node/yarn version where heroku buildpack will find them

phinze:
# What it does

As a part of #1185 we need to tell the nodejs buildpack which version of node to deploy. To do this we follow [the heroku docs](https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version) and add to package.json.

Added a few unit tests to help us remember to keep versions consistent in the various config files.

# Why it is important

Per #1185 this gets us in line with Heroku's recommendations and quells a big warning in the logs every time we deploy.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,7 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
-      - run: yarn install --frozen-lockfile
+      - run: yarn install --frozen-lockfile --ignore-engines
 
       - run: bin/rails assets:precompile
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version-file: "package.json"
 
       - uses: actions/cache@v4
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -61,7 +61,7 @@ jobs:
 
       - uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version-file: "package.json"
```
jim, line 64: Nice, I love to see this config being consolidated.

test/lib/version_consistency_test.rb
```diff
@@ -0,0 +1,32 @@
...
+    tool_versions = parse_tool_versions
+    package_json = JSON.parse(File.read(Rails.root.join("package.json")))
+
+    assert_equal tool_versions["yarn"].split(".").take(2), package_json["engines"]["yarn"].split(".").take(2), "major and minor yarn version match"
+    assert_equal tool_versions["nodejs"].split(".").take(1), package_json["engines"]["node"].split(".").take(1), "major node version matches"
+  end
+
+  # Heroku buildpack reads from Gemfile
+  # asdf reads from .tool-versions
+  # nix reads from .ruby-version
```
jim, line 23: This is why we can't have nice things 🤣

jim (approved):
Thank you for sorting out this version mess!

PR #1433, opened 2024-04-02
Update production data scrubbing tasks to be more thorough.

jim:
# What it does

Fixes some issues with the production data scrubbing:

* There were some changes needed to accommodate changes to how ActiveStorage handles variants.
* More instances of personal information are cleared out or deleted.
* Some rough edges in the import process have been addressed so that importing production data is less likely to leak PG extensions, etc. into your local environment.

### DIFF
--- a/app/models/loan.rb
+++ b/app/models/loan.rb
@@ -2,7 +2,7 @@ class Loan < ApplicationRecord
   belongs_to :item, optional: true
   belongs_to :member
   has_one :adjustment, as: :adjustable
-  has_many :renewals, class_name: "Loan", foreign_key: "initial_loan_id"
+  has_many :renewals, class_name: "Loan", foreign_key: "initial_loan_id", dependent: :destroy
   has_many :renewal_requests, dependent: :destroy
   belongs_to :initial_loan, class_name: "Loan", optional: true
   has_one :hold, dependent: :nullify
--- a/lib/tasks/production.rake
+++ b/lib/tasks/production.rake
@@ -90,7 +90,7 @@ def scrub_data
   SQL
 
   # There are some members in the database without associated users that the above query doesn't update.
-  Member.left_joins(:user).where(user: { id: nil }).each { |m| m.destroy }
+  Member.left_joins(:user).where(user: {id: nil}).each { |m| m.destroy }
 
   User.connection.execute(<<~SQL)
     UPDATE users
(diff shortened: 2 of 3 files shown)


app/models/loan.rb
```diff
@@ -2,7 +2,7 @@ class Loan < ApplicationRecord
   belongs_to :item, optional: true
   belongs_to :member
   has_one :adjustment, as: :adjustable
-  has_many :renewals, class_name: "Loan", foreign_key: "initial_loan_id"
+  has_many :renewals, class_name: "Loan", foreign_key: "initial_loan_id", dependent: :destroy
```
jim, line 5: Without this, some members with renewals can't be deleted using `Member#destroy`.

lib/tasks/production.rake
```diff
@@ -14,60 +14,98 @@ def run_and_print(command)
...
     recipient_name = 'Recipient ' || id,
     code = 'ABCD' || id
   SQL
+
+  Member.connection.execute(<<~SQL)
+    UPDATE members
+    SET
+      full_name = 'Firstname Lastname',
+      preferred_name = coalesce('Member '||members.number, 'User '||users.id),
+      email=coalesce('member'||members.number, 'user'||users.id) || '@example.com',
```
jim, line 81: We're now using the member's number instead of their ID, which should make troubleshooting slightly nicer.

app/models/loan.rb
phinze: I think it would be nice to have a covering unit test here. I'll push a commit with one! 🛠️

phinze: 16e8792cf8c1cfd4df2edd381e4797cba93dbf67

lib/tasks/production.rake
```diff
@@ -14,60 +14,98 @@ def run_and_print(command)
   end
 end
 
+# This is useful when trying to reproduce a production issue, but it's best to
+# avoid doing it live.
 desc "Load production data and modify for local usage"
 task load_production_data: [:pull_production_database, :environment] do
   raise "no way buddy" if Rails.env.production?
+  delete_attachments
+  exorcise_production_remnants
```
phinze, line 23: lol good verb choice 👻

phinze (approved):
Looking good! It will be useful to be able to get scrubbed dumps for dev usage

jim:
Thanks for adding that test!

PR #1434, opened 2024-04-03
Fix CSRF error when reordering holds

### LINKED ISSUE #1427: Hold list order modifications don't take effect
**Describe the bug**
In the admin view, when I adjust the order of the holds it will show that it's changed but if the page is refreshed or viewed anew from somewhere else, the holds are shown as reverted to their original sequence.

**To Reproduce**
1. Go to 'C-1849' in the admin view.
2. Click on 'Keith' in 4th position and move them to 2nd position on the holds list.
3. Refresh the page or look at the same page from another window/tab.
4. See error with Keith reverting back to 4th position.

**Expected behavior**
The hold should update and actually reflect the desired change.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Google Chrome

**Additional context**
This obviously isn't super urgent because we have not used the feature to date.  Also, I have a question about what happens if someone is moved into first position in this manner. Will it send out a notification email?  Obviously the original 1st in line will have received an email which we won't be able to retract.

phinze:
# What it does

Fixes #1427

# Implementation notes

There was a small bug in getting the right token content from the meta tag that tests didn't catch because they have CSRF protection off. Turning on CSRF protection in system tests catches this bug and should hopefully prevent future ones too.

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -89,6 +89,8 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   setup do
     ActsAsTenant.test_tenant = libraries(:chicago_tool_library)
+    # System tests submit real forms so we can exercise CSRF protection
+    ActionController::Base.allow_forgery_protection = true
   end
 
   include Warden::Test::Helpers
@@ -102,6 +104,7 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   teardown do
     ActsAsTenant.test_tenant = nil
+    ActionController::Base.allow_forgery_protection = false
 
     errors = page.driver.browser.logs.get(:browser)
     fail = false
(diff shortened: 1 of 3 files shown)


test/application_system_test_case.rb
```diff
@@ -89,6 +89,8 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
 
   setup do
     ActsAsTenant.test_tenant = libraries(:chicago_tool_library)
+    # System tests submit real forms so we can exercise CSRF protection
+    ActionController::Base.allow_forgery_protection = true
```
jim, line 93: Nice. Hopefully this will help us catch things like this in the future.

jim (approved):
Cool. We've been seeing `InvalidAuthenticityToken` errors a bit, I wonder how many of them will go away with this fix.

Issue #555, opened 2021-05-28
Add a reason to the member page when a loan can't be renewed or have a renewal requested


jim:
> Sorry, this item isn't eligible for renewal at the moment because it's on hold by another member.

PR #1441, opened 2024-04-04
Add message when renewal is not available for an item

### LINKED ISSUE #1440: Add language about renewal being unavailable when renewal limit is reached

In conversation with @CTLvivek he mentioned that staff are hesitant to use the Renewal Limit feature of borrow policies because there is no language on the member loans page when renewal is unavailable.

When renewal is not allowed, the item should show some language that makes that clear.

phinze:
# What it does

Adds language on the member loans page to indicate explicitly when an item is not renewable.

# Why it is important

Resolves #1440 

# UI Change Screenshot

2024-04-04 at 12 08 26@2x

### DIFF
--- a/app/controllers/admin/bulk_renewals_controller.rb
+++ b/app/controllers/admin/bulk_renewals_controller.rb
@@ -7,7 +7,7 @@ module Admin
 
     def update
       @member.loans.checked_out.each do |loan|
-        renew_loan(loan) if loan.renewable?
+        renew_loan(loan) if loan.within_renewal_limit?
       end
 
       redirect_to admin_member_url(@member.id), status: :see_other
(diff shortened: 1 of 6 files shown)


phinze:
@CTLvivek when you get the chance can you review the language and confirm this addresses the concern we talked about?

phinze:
I just found #555 with more ideas for language on this page for various scenarios so I'm adding a few conditionals here.

When another hold exists:

2024-04-04 at 12 24 27@2x

When renewal limit is reached:

2024-04-04 at 12 27 14@2x

jim (approved):
It brings me joy seeing these old issues get addressed! This code looks good to me.

CTLvivek:
I added my thoughts on the thread in the #circulate Slack.  Appreciate all of your work!

phinze:
Ok incorporated the copy feedback and filed #1455 as a separate issue about blocking renewal for long-overdue items. Shipping this! 🚢

PR #1442, opened 2024-04-04
Add full name to memberships report

phinze:
# What it does

Adds full names to membership report

# Why it is important

@tessavierk needs a membership report with full member names, and the membership report only shows preferred / first names


phinze:
Self-merging this so we can get the report generated. If I've broken something, I will fix it 😄 🛠️

jim:
👍🏻

PR #1443, opened 2024-04-05
Fix broken teamup redirect

phinze:
# What it does

Fixes the `UnsafeRedirectError` we [saw yesterday](https://appsignal.com/chicago-tool-library/sites/60596f9214ad662e17191689/exceptions/incidents/3113/samples/timestamp/2024-04-05T02:38:32Z).

# Why it is important

Errors bad!


jim (approved):
Nice, thanks for fixing this!

PR #1444, opened 2024-04-06
Refactor Google Calendar authentication

### LINKED ISSUE #1437: Permission error when syncing calendars
We're starting to see the following error when trying to call `sync:calendars` in production:

```
Could not sync events: {
  "error": {
    "code": 403,
    "message": "Method doesn't allow unregistered callers (callers without established identity). Please use API Key or other form of API consumer identity to call this API.",
    "errors": [
      {
        "message": "Method doesn't allow unregistered callers (callers without established identity). Please use API Key or other form of API consumer identity to call this API.",
        "domain": "global",
        "reason": "forbidden"
      }
    ],
    "status": "PERMISSION_DENIED"
  }
}
```

Looks like we need to figure out how to generate a different kind of API key for this task.

@jim mentioned he would look into this since he's got permissions to poke around in our Google workspace account.

Since we generate events for the next 90 days we have plenty of grace time to fix this, so long as we don't change open hours in the meantime.

jim:
# What it does

* Use [googleauth](https://github.com/googleapis/google-auth-library-ruby) library to handle token generation (we previously were doing this ourselves)
* Allows the use of a GCP service account to authenticate (we'll use this in production)
* Adds documentation that explains how to setup authentication in development for anyone wanting to work on this code.
* This should fix #1437.

# Why it is important

* Google made a change to how authentication was working in production, and as a part of sorting that out, I wanted to move toward using a service account instead of OAuth. This turned out to be tricky to setup for local development (there are a lot of things to turn on/grant in a brand new GCP project, which is what most folks will start with), so for that use case I've preserved the existing OAuth option.

# Implementation notes

* We have a volunteer shift scheduling system that is currently not used by CTL. There's a decent amount of code for that functionality that we could consider removing, which would let us really simplify the Google Calendar parts of the codebase.
* The use of `.profile` to write the contents of an environmental variable to a file feels a little gross, but it seems to be a common way to work around the fact that Heroku's 12-factor philosophy and Google's default way of handling credentials are at odds.

# Deployment Plan

* I've already setup the environment variables in production. I will create some new calendars for staging and configure a service account there as well so that we can verify things are working before going to prod.


phinze (approved):
What a saga! Thanks for wading through all this. 🌊

PR #1445, opened 2024-04-06
Wait for logs to write to appsignal before exiting rake tasks.

jim:
# What it does

I noticed that logs for tasks run through the Heroku scheduler aren't showing up in AppSignal. [Their docs mention needing to add some code to ensure that logs are send before the process exits](https://docs.appsignal.com/ruby/integrations/rake.html#rake-tasks-and-containers), so I added that to the calendar tasks. If it works, we can add it to the other ones that send emails and the like.

# Why it is important

It's nice to keep an eye on if these tasks are running and how many events they are updating.

### DIFF
--- a/lib/tasks/shared.rake
+++ b/lib/tasks/shared.rake
@@ -1,0 +1,6 @@
+# Allow time for AppSignal to ship logs before the process exits
+# https://docs.appsignal.com/ruby/integrations/rake.html#rake-tasks-and-containers
+def wait_for_logs_to_flush
+  Appsignal.stop "rails"
+  sleep 5
+end
--- a/lib/tasks/shared.rake
+++ b/lib/tasks/shared.rake
@@ -1,6 +1,7 @@
-# Allow time for AppSignal to ship logs before the process exits
+# Close process allowing Appsignal to flush any exceptions that were rescued
+# and manually sent along, per recommendations here:
 # https://docs.appsignal.com/ruby/integrations/rake.html#rake-tasks-and-containers
-def wait_for_logs_to_flush
-  Appsignal.stop "rails"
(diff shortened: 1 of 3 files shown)


lib/tasks/shared.rake
```diff
@@ -0,0 +1,6 @@
+# Allow time for AppSignal to ship logs before the process exits
+# https://docs.appsignal.com/ruby/integrations/rake.html#rake-tasks-and-containers
+def wait_for_logs_to_flush
+  Appsignal.stop "rails"
+  sleep 5
+end
```
phinze, line 6: I have a similar helper in email.rake - do you think you could consolidate that copy into this one?

phinze:
Y'know even the [tasks that already have code for this](https://github.com/chicago-tool-library/circulate/blob/1f4fb6959353b4778e8d301c1476fe11170740f4/lib/tasks/email.rake#L54) don't seem to have logs showing up. 🤔 

I wonder if we should try ripping out lograge and just doing [vanilla `broadcast_to`](https://docs.appsignal.com/logging/platforms/integrations/ruby.html#rails-logger) to see if that might help?

lib/tasks/shared.rake
jim: Oh totally, I hadn't noticed we already had something in place for this.

phinze (approved):

jim:
@phinze Good call. I think it probably makes sense to try going stock Rails logging and see if the task logging starts working. No point in merging this since we know it's not gonna do anything.

phinze:
I actually think we should also merge this due to what they're saying in their docs about async stuff. I'll push the log changes as a separate PR regardless.

PR #1447, opened 2024-04-07
Send late hold available texts the next morning

### LINKED ISSUE #1446: Hold texts should not go out at unreasonably late hours
Most of our text messages are sent during jobs that run at specific times, and by controlling the times those jobs run we can ensure that we're only sending texts at reasonable times.

Hold available messages, however, are triggered by a job that runs once every 10 minutes. I had originally assumed that holds would only need to be notified when an item was returned, so they'd effectively only happen during tool library hours. I hadn't accounted for the fact that holds have an automatic expiration time of 7 days, and the expiration timestamp is set for midnight on that 7th day.

That means for holds that expire we're notifying users at somewhere between 12-12:10am, which is generally not when you want to get a text message from an automated system!

We should defer these text messages to the next day at a reasonable hour.

phinze:
# What it does

Resolves #1446 by using [Twilio's message scheduling feature](https://www.twilio.com/docs/messaging/features/message-scheduling) to defer late text messages to the next morning.

Specifically, if a hold available message would go out after 8pm, we schedule it for 9am the following morning.

# Why it is important

See #1446 for details.

# Implementation notes

* I tested this manually with Twilio credentials to verify the scheduling parameters seem to be working how we expect.
* I decided to only add this behavior to hold available messages for a couple reasons:
  1. To limit the complexity of this extra logic to a single message type
  2. Scheduled messages are an [extra 1.5 cents per message](https://www.twilio.com/en-us/sms/pricing/us#SMS-features) with the first 1000 per month being free. I think since we control the timing of the other messages with job time, we are more likely to keep ourselves under 1k/mo scheduled messages by only doing it for expired holds.
* Alternatively, we could implement message scheduling in our jobs system via [ActiveJob's `wait_until`](https://guides.rubyonrails.org/active_job_basics.html#enqueue-the-job), but since we're using in-process background jobs it seemed dangerous to float scheduled texting jobs for hours on end. Visibility would be low and we'd lose all waiting jobs on process reboot. In the future, we could explore the tradeoffs changing background job systems and handling the scheduling ourselves.

### DIFF
--- a/app/texters/base_texter.rb
+++ b/app/texters/base_texter.rb
@@ -25,8 +25,29 @@ class BaseTexter
     @client ||= BaseTexter.client || Twilio::REST::Client.new
   end
 
-  def text(to:, body:)
+  def text(to:, body:, send_at: nil)
     Rails.logger.debug("Sending SMS to: #{to}, body: #{body}")
-    client.messages.create(from:, to:, body:, messaging_service_sid:)
+    schedule_args = {}
+    if send_at.present?
+      schedule_args = {send_at: send_at, schedule_type: "fixed"}
+    end
+    client.messages.create(
+      messaging_service_sid:,
+      from:,
+      to:,
+      body:,
+      **schedule_args
+    )
+  end
+
+  REASONABLE_HOURS = (9...20)
+  # Sends text immediately if between 9am-8pm local time, otherwise schedules
+  # for tomorrow 9am.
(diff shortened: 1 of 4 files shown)


jim (approved):
Cool. I didn't even realize that Twilio had a feature we could use here.

Code, tests, and logic looks good to me.

> In the future, we could explore the tradeoffs changing background job systems and handling the scheduling ourselves.

I agree. It would be nice to have that setup, but this'll work for now.

PR #1181, opened 2023-11-08
Replace Turbolinks with Turbo (and upgrade Stimulus)

jim:
# What it does

* Replaces Turbolinks in the app with the newer version of the library, Turbo.
* Upgrades Stimulus to the latest version
* Replaces our PortalRendering code with equivalent Turbo-based code.

Fixes #1184.

# Why it is important

Turbolinks is EOLed, and we need to upgrade to Turbo to get back on a supported library. And by using Turbo instead of  custom code, we have less to maintain and it's easier for contributors to understand how things work as there are many resources out there for learning Turbo.

# Implementation notes

* This is based on the main parts of #810, without the use of real-time Turbo Streams as that requires Redis. We can add that later if needed, but I don't want to pull in that dependency as a part of this.

# Your bandwidth for additional changes to this PR

_Please choose one of the following to help the project maintainers provide the appropriate level of support:_

- [x] I have the time and interest to make additional changes to this PR based on feedback.
- [ ] I am interested in feedback but don't need to make the changes myself.
- [ ] I don't have time or interest in making additional changes to this work.
- [ ] Other or not sure (please describe):

PR #1450, opened 2024-04-08
Move appointments between pending/completed lists

### LINKED ISSUE #1438: When completing appointment, it does not pop down to the completed list
**Describe the bug**

According to @CTLvivek, the appointments list complete button used to pop the row down to the completed list right away, and now it only dims the row, but does not move it until reload.

**To Reproduce**
1. As an admin, go to the appointments list for a day with appointments
2. Click "complete"

**Expected behavior**
Appointment row pops down to completed appointments list lower on the page.

**Observed behavior**
Appointment row only dims, does not pop down until reload.

**Additional context**

I can't find any code that would cause the row to travel to the other list... I'm wondering if the previous observed behavior was due to the pre-turbo/tubolinks full page reload? Question to investigate: how long ago was the appointments list non-turbolinks-enahanced?

jim:
# What it does

Fixes #1438 by moving appointments to the appropriate list when the are completed or restored. This feature was originally implemented in dbf9983 but removed in #1181 as a part of the upgrade from Turbolinks to Turbo.

# Why it is important

Staff have given us feedback that they miss having the app automatically move completed appointments to the list at the bottom of the page without needing a page refresh.

# Implementation notes

* The new implementation works on both the new and old UI for the appointment list.
* `FEATURE_NEW_APPOINTMENTS_PAGE` now defaults to `false` in the test environment (so we can test that the current default is working in tests). We could also write tests for both, but that can be a followup.
* I also fixed a bug where completion didn't work properly when `new=true` using a query param.


app/models/appointment.rb
```diff
@@ -12,8 +12,15 @@ class Appointment < ApplicationRecord
...
-
   scope :simultaneous, ->(appointment) { where(starts_at: appointment.starts_at, ends_at: appointment.ends_at).where.not(id: appointment.id) }
+  scope :same_day_as, ->(appointment) {
+    where("starts_at > ? AND ends_at < ?", appointment.starts_at.beginning_of_day, appointment.ends_at.end_of_day)
+      .chronologically
+  }
+  has_many :same_day_and_member, ->(appointment) { same_day_as(appointment) },
+    class_name: "Appointment",
+    foreign_key: "member_id",
+    primary_key: "member_id"
```
jim, line 23: This association allows us to eager-load this data in `AppointmentsController#index` and avoid a bunch of N+1s.

app/helpers/appointments_helper.rb
```diff
@@ -7,4 +7,49 @@ def format_appointment_times(start, stop)
...
+  # 12-eee-12
+  #
+  def appointment_sort_key(appointment)
+    first_start_time_for_member = appointment.same_day_and_member.first.starts_at.hour
+
+    start_time = appointment.starts_at.hour
+    member_name = preferred_or_default_name(appointment.member)
+    created_at = appointment.created_at.to_i
+    "#{first_start_time_for_member}-#{member_name}-#{start_time}-#{created_at}"
+  end
```
jim, line 54: I considered putting method directly on `Appointment`, but ended up with it here since it's mostly about UI.

app/controllers/admin/appointments_controller.rb
```diff
@@ -8,15 +8,16 @@ def index
...
         .where(starts_at: @current_day.beginning_of_day..@current_day.end_of_day)
         .chronologically
-        .includes(:member, loans: {item: :borrow_policy}, holds: {item: :borrow_policy})
+        .includes(:member, :same_day_and_member, loans: {item: :borrow_policy}, holds: {item: :borrow_policy})
 
       @pending_appointments = []
       @completed_appointments = []
-      sort_by_member_and_time(@appointments).each_with_index do |appointment, index|
+
+      @appointments.sort_by { |a| helpers.appointment_sort_key(a) }.each do |appointment|
```
jim, line 16: This is a little funky, but I figured it made sense to use the same sorting technique here that we do on the front end.

mdecelle (commented):
I tried running this locally and I got some odd behavior in the UI? It seemed to not respect the "new" vs. "old" views and mixed them up when the appointment was completed. Not 100% sure why. 
image

phinze:
> I tried running this locally and I got some odd behavior in the UI? It seemed to not respect the "new" vs. "old" views and mixed them up when the appointment was completed. Not 100% sure why.

I figured this out! It would work fine if you specified the feature flag in `.env`, but if you were trying to just use the query param to change layouts it wouldn't properly pass through.

I just pushed a commit which fixes that issue. Next I'll look into that CI test failing... looks like it repros on my machine. 🔍

phinze (approved):
Ok tests passing now too. 🚀 Dropping my ✅ on this!

app/helpers/appointments_helper.rb
```diff
@@ -7,4 +7,49 @@ def format_appointment_times(start, stop)
...
+  # The sort key consists of the following values joined together:
+  #
+  # - start time of a member's first appointment of the day
+  # - the member's preferred name
+  # - the start time of the appointment
+  # - the creation date of the appointment
+  #
+  # The actual sort keys include the appointment creation time to ensure a stable sort order
+  # in the case of members with the same name and time, but for simpicity it is not shown in
+  # the examples below.
```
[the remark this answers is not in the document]
jim: That's a good call! We have some code [that prevents appointments from being created at the same time as an existing one](https://github.com/chicago-tool-library/circulate/blob/fbf56892151b5cad31e5e00d4859fd98b7eeeac1/app/controllers/account/appointments_controller.rb#L25). It should also handle updating an appointment to be the same time as another one.

jim:
@mdecelle and @phinze Thanks for the reviews, suggestions, and fixes!

jim:
I checked with the librarians and they said the group-by-member part was no longer needed. I love to delete complicated code!

PR #1452, opened 2024-04-09
Rework logging config to try to get rake logs in appsignal

phinze:
# What it does

Reconfigure logging to attempt to get non-web tasks to show up in AppSignal

# Why it is important

We want logs for background work!

# Implementation notes

* I went down the rabbit hole reading about this, and I think the best thing for us to try next is to remove the [Ruby level Appsignal logging integration](https://docs.appsignal.com/logging/platforms/integrations/ruby.html) and instead use a [Heroku Log Drain](https://docs.appsignal.com/logging/platforms/heroku.html).
* I've already configured the Heroku Log Drain in production, so this change is meant to simplify the config to just let lograge give us logs on stdout.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -100,27 +100,6 @@ Rails.application.configure do
   # require 'syslog/logger'
   # config.logger = ActiveSupport::TaggedLogging.new(Syslog::Logger.new 'app-name')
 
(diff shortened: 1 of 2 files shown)


phinze:
Going to self-merge this since it's either going to work in Heroku or not work. ⚖️

Issue #1240, opened 2024-01-15
Improve item search


jim:
Questions:

- How do we want to improve search? What doesn't work that we want to change?

PR #1454, opened 2024-04-11
Add Ahoy for tracking and begin recording events for tool search

phinze:
# What it does

This installs the [`ahoy_matey`](https://github.com/ankane/ahoy) gem for first-party analytics and begins using it to track when members do a search, view an item, and place a hold.

# Why it is important

As we aim to improve search in the system (#1240), it will be helpful to see what users are actually searching for and whether they are finding it.

# Implementation notes

* 🤫 By using a first-party analytics system (i.e. one that stores data in the same database we already use and control) we're not storing or exposing any more user information than we already have via our member information and logs.
* 🏋🏻 It's worth noting that this will put more storage pressure on the database, because it's recording rows for stuff we expect users to do a lot. So we'll want to keep an eye on database growth.
* ♻️ Along those lines, will want to add [a job that prunes old data](https://github.com/ankane/ahoy?tab=readme-ov-file#data-retention) but I figure that can come in a follow-on PR.
* 📈 We can add as many more events as we like over time, but I figure starting relatively simple is the way to go.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -102,3 +102,5 @@ gem "jsbundling-rails", "~> 1.3"
 gem "cssbundling-rails", "~> 1.4"
 
 gem "twilio-ruby", "~> 7.0"
+
+gem "ahoy_matey", "~> 5.1"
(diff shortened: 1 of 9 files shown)


jim (approved):
"ahoy_matey" 🤣🚢

PR #1456, opened 2024-04-11
Revise pull request template

phinze:
# What it does

Makes some tweaks to the pull request template. Specifically:

 * Removing the "bandwidth for additional changes" section
 * Changing the instructions into markdown comments so they're invisible if people don't delete them

# Why it is important

I believe the bandwidth section provides more noise than value. Most of our contributors nowadays are known to us. Even when people cut it down to just the line relevant to them, it triggers GitHub's "task tracking" feature on PR list views. I believe if we have any questions about somebody's interest for follow up work, we can just ask them case-by-case.


jim (approved):
I agree with all of this. The previous format was really designed for a different sort of team makeup.

PR #1460, opened 2024-04-17
Group Lending UI cleanup / attribute expansion

jim:
# What it does

This PR does two things:
* Starts to build out the `ItemPool` and `ReservableItem` models by adding more database fields and associations. This includes updating the categorization system to work with both items and item pools.
* Cleans up some of the top-level admin screens for item pools, taking advantage of the new attributes.

# UI Change Screenshot

Here's the still very WIP item pool index and show pages:

Screenshot 2024-04-17 at 8 45 55 AM

Screenshot 2024-04-17 at 8 46 08 AM

# Implementation notes

* I am experimenting with using `ActiveSupport::Concern` to share behavior between `Item`/`ItemPool`/`ReservableItem`. I think I like it so far, but it remains to be seen how well it works as we build more parts of group lending out.
* In terms of which of the existing `Item` fields end up on `ItemPool` or `ReservableItem` (or both), that is something that we'll need to be informed by talking to staff and also looking at the actual data we need to import. So there will definitely be changes there- this is just to get enough on those models that we have enough data to build out simple screens.
* I noticed that there are likely some places where we're not calling `CategoryNode.refresh` to rebuild the materialized `category_nodes` view that we should be. Fixing this is left as future work.


phinze (approved):
This is exciting progress!

The usage of the Concern seems like a good move at the moment, and I think the similarities/differences between Item, ItemPool, and ReservableItem are something we should pay attention to closely as this continues to evolve.

db/migrate/20240416013510_add_attributes_to_reservable_items.rb
```diff
@@ -0,0 +1,15 @@
...
+    add_column :reservable_items, :brand, :string
+    add_column :reservable_items, :model, :string
+    add_column :reservable_items, :serial, :string
+    add_column :reservable_items, :number, :integer
+    add_column :reservable_items, :purchase_price_cents, :integer
+    reversible do |direction|
+      direction.up { execute "UPDATE reservable_items SET number=id" }
+    end
+    change_column_null :reservable_items, :number, false
+    add_index :reservable_items, ["number", "library_id"], unique: true
```
phinze, line 13: ⚠️ Looks like power_source landed on item_pools but not reservable_items, but it moved to the Concern shared by all three models.

phinze:
Nonblocking conversation to tee up: how do we want to handle reservable items that only have one instance? The myturn "pools are just duplicate names" model allows them to ignore this question because there's no separate entity for the pool. Currently we require all ReservableItems belong to a pool, so for a single item would we want to have a pool w/ a single item? I can see pros and cons to different approaches, hence probably a discussion.

db/migrate/20240416013510_add_attributes_to_reservable_items.rb
jim: I was thinking we wouldn't have items in an item pool with more than one power_source, although maybe that's incorrect? Currently `ItemAttributes` gets mixed into `ItemPool` and not `ReservableItem`, but due to the naming that's likely confusing. Definitely a WIP.

I wonder if it's worth doing a naming review at some point and seeing if we want to adjust what we're calling things before we write too much code using them?

phinze: Ah ok I missed that ItemAttributes wasn't mixed into ReservableItem, that clears up my concern! I'm not super concerned with doing a naming review earlier vs later, but I do think it would be good to do one before we go live.

PR #1461, opened 2024-04-17
Add Pickup checkin/checkout

jim:
# What it does

Extends our group lending prototype to allow for pickups to be checked out and back in to represent a reservation being picked from the library and then brought back sometime later on.

# UI Change Screenshot

Screenshot 2024-04-17 at 9 00 50 AM
Screenshot 2024-04-17 at 9 01 21 AM


phinze (approved):
Nice this all looks good

PR #1464, opened 2024-04-22
1424 incorporate feedback from appointment UI changes

mdecelle:
# What it does

Creates a separate mobile and desktop layouts for librarian appointments.

# Why it is important

This enables a more user friendly design for mobile users while preserving existing usability for desktop users.

# UI Change Screenshot

Desktop View
image

Mobile View
image

# Implementation notes

Both views are always in the DOM. Spectre CSS is utilized to only display the proper view based on the screen size.

Note: Sorting of the appointments based on completion status has not been done for the mobile layout.


app/javascript/lib/appointments.js
```diff
@@ -1,6 +1,7 @@
 // Given an appointment container, moves it into the appropriate list (either
 // pending or completed).
 export function arrangeAppointment(id) {
+  console.log(id)
```
phinze, line 4: Looks like an extra debug log in here - I'll push a commit to remove it

phinze (approved):
Looking good! I'll get this out to prod so we can check it out on real appointments pages.

PR #1465, opened 2024-04-22
Rename DateHold to ReservationHold

jim:
# What it does

Renames `DateHold` to `ReservationHold` throughout the codebase and database. As I've been working on the group lending code, it's become clear to me that there is something useful about having the models prefixed with `Reservation` or `Reservable`.


phinze (approved):
Looks reasonable!

PR #1466, opened 2024-04-23
Nest reservable items underneath item pools

jim:
# What it does

Moves the `Admin::ReservableItemsController` into the `Admin::ItemPools` namespace. At this point I think this makes the most sense, as we're mostly going to be working with reservable items in the context of an item pool.


phinze (approved):
Makes sense to me!

PR #1468, opened 2024-04-23
Add ReservationMailer

jim:
# What it does

Adds a new mailer for reservation-related emails. This currently sends when a reservation is created, we'll want to adjust this to when it's submitted for approval once we get that flow in place.


app/views/reservation_mailer/reservation_requested.mjml
```diff
@@ -0,0 +1,10 @@
+<mj-section>
+  <mj-column padding-top="20px">
+    <mj-image src="<%= image_url "logo.jpg" %>" width="100px" />
+  </mj-column>
+</mj-section>
+<mj-section>
+  <mj-column>
+    <mj-text font-size="36px" line-height="28px" font-weight="bold" align="center"><%= @subject %></mj-text>
+  </mj-column>
+</mj-section>
```
phinze, line 10: Nonblocking micro-nit: I've noticed our editors fighting over EOF newlines on non-ruby files. I did a quick web search and it seems like the rough consensus is to leave newlines at end of every file ([example](https://evanhahn.com/newline-necessary-at-the-end-of-javascript-files/)). Do you think you could configure your editor to put them in? Here's instructions for VSCode: https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files

phinze (approved):
LGTM! One nonblocking nit inline.

app/views/reservation_mailer/reservation_requested.mjml
jim: Good call, and thanks for the nudge on this. I set this up for all filetypes (it was only set to true for Ruby files). I might look at adding it to the precommit hooks as well (although I don't know if anyone else has those enabled).

Issue #1250, opened 2024-01-21
Add search to Notifications report


jim:
We have a report that lists all notifications sent to users that you can see at http://localhost:3000. It's helpful to look in here when folks say they didn't receive and email to see if we actually sent one or not. Unfortunately, there is no search, so you have to page through the listing, and as we grow, this becomes more and more unwieldy.

Add a search field so that a staff member can enter an email address and see only notifications sent to that email.
Then, add a link to the member page that links to the search results page, so someone can see the notifications sent to that member in one click.

We might need to add an index to the `notifications` table so that we can efficiently search the `email` column.

jim:
No longer needed thanks to #1469.

PR #1469, opened 2024-04-27
Add link to filtered notifications report

jim:
# What it does

Adds a link to the members page that goes to the notifications report, filtered by that member. It turns out this functionality has been there for a while, but there was just not a convenient link that exposed it.

This obviates the need for #1250.

# Why it is important

It's really useful to see what communications we've sent to a given member.

# UI Change Screenshot

image

# Implementation notes

* There's already an index in the database on `notifications.member_id`.

### DIFF
--- a/app/views/admin/members/_profile.html.erb
+++ b/app/views/admin/members/_profile.html.erb
@@ -2,6 +2,7 @@
   <%= index_header preferred_or_default_name(@member) do %>
     <%= link_to "Edit Member", edit_admin_member_path(@member), class: "btn" %>
     <%= link_to "Add Note to Member", new_admin_member_note_path(@member), class: "btn", data: {turbo_frame: "new-note"} %>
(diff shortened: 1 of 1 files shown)


phinze (approved):
Looks like an easy win to me! Should we pull the link out of the overflow menu or leave it in both places?

2024-04-27 at 09 46 52@2x

Something we can solve in a separate PR though - I can shepherd this out now.

jim:
You know, I had missed that we had a link (and actually, a different controller) already 🤦🏻. Truth be told we can probably revert the button addition.

I might do that and clean up the `Admin::Reports::NotificationsController` to not have a `member_id` filter on it since we have that functionality available in `Admin::Members::NotificationsController` already.

phinze:
Haha fair! Though... the fact that both you and Vivek both missed it being there within the last few months might be signal that we should do something about that three dots menu 💭 - will keep noodling on it.

PR #1470, opened 2024-04-27
Use newer default value for config.action_dispatch.cookies_same_site_protection

jim:
# What it does

Starting in Rails 6.1, the default value of this setting was changed to `:lax`. Our app predates that default, and the current behavior is less secure than the default.

[The Rails guides have more info](https://edgeguides.rubyonrails.org/configuring.html#config-action-dispatch-cookies-same-site-protection).

You can see [the security warning from GitHub](https://github.com/chicago-tool-library/circulate/security/code-scanning/3) as well.


phinze (approved):
Good call. 🔒

PR #1471, opened 2024-04-27
Change name display on appointments

phinze:
# What it does

Updates member name display on admin appointments page

* Desktop: Preferred name w/ full name in tooltip
* Mobile: Preferred name w/ full name in parens (since tooltips on links don't work on mobile as there's no hover)

# Why it is important

Staff wants to address people by preferred name, but they sometimes need to differentiate between multiple people with the same name.

# UI Change Screenshot

Desktop:

2024-04-27 at 10 12 27@2x

Mobile:

2024-04-27 at 11 50 06@2x


jim (approved):
A very pragmatic solution to the name issue we're having while we sort out the broader naming concerns. :shipit:

PR #1472, opened 2024-04-28
Add left hook config for JS and trailing newlines.

jim:
# What it does

Adds to our [lefthook](https://github.com/evilmartians/lefthook) config two new checks that run on `pre-commit`. One checks that [standard](https://standardjs.com) (the original JS one, not the Ruby one) has formatted any JavaScript code, and the other checks that all text files end with a newline.

# Why it is important

Consistency is good.


phinze (approved):
👏🏻

PR #1478, opened 2024-05-01
Attempt fix for InvalidAuthenticityToken errors

phinze:
# What it does

Implements a change suggested [here](https://blog.alex-miller.co/rails/2017/01/07/rails-authenticity-token-and-mobile-safari.html) and [here](https://github.com/rails/rails/issues/21948#issuecomment-147738473) which should hopefully reduce the number of InvalidAuthenticityToken errors we see.

# Why it is important

ActionController::InvalidAuthenticityToken is one of the more common errors we see, and though some of them probably come from random web traffic, there are enough to suggest we have an actual issue.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,17 @@ class ApplicationController < ActionController::Base
   include ActiveSupport::Testing::TimeHelpers
   include PageAttributes
 
+  # Move CSRF token storage from its default location in the session (which is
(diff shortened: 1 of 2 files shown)


phinze:
😮 Wow did not expect tweaking a caching header to absolutely decimate system tests. Looking into this 🕵🏻

jim (approved):
Seems worth a shot!

jim (approved):
I like the new implementation. I'll deploy this week and keep an eye out for any issues.

Issue #1147, opened 2023-10-10
Item loan history page uses item layout.


jim:
The Loan History page for an item is accessed via this link on an item page:

![Image](https://user-images.githubusercontent.com/3331/268805690-d83eb7ca-8240-4a9a-9ed1-d89b376e3f9f.png)

Unfortunately, it doesn't use the same item page layout that the other tabs do:

![Image](https://user-images.githubusercontent.com/3331/268805699-f9019040-1fb5-499e-84a1-c04fd9735f8a.png)

To fix this, we should update the page so that it renders the loan history within the correct tab on the item page.

PR #1479, opened 2024-05-03
Item loan history page uses item layout

katielentes:
# What it does

Adds an admin `LoanHistoriesController` and corresponding route to use same layout as the rest of the tabs in the admin/item view. 

# Why it is important

More seamless navigation on Item Loan History Page

# UI Change Screenshot
Screenshot 2024-05-02 at 10 04 35 PM

# Implementation notes

Modeled `LoanHistoriesController` based on `HistoriesController` in `app/controllers/admin/items/histories_controller.rb`

Also made the due/returned status a chip and added an extra "Date" column to the table to work nicely with the chip.

Is `app/views/admin/loan_histories/show.html.erb` used anywhere else? I haven't been able to find where the existing loan history page is used elsewhere but I want to confirm before removing.

Resolves [#1147](https://github.com/chicago-tool-library/circulate/issues/1147)

### DIFF
--- a/app/controllers/admin/items/loan_histories_controller.rb
+++ b/app/controllers/admin/items/loan_histories_controller.rb
@@ -1,0 +1,10 @@
+module Admin
+  module Items
+    class LoanHistoriesController < BaseController
+      def show
+        @item = Item.find(params[:item_id])
+        @loans = @item.loans.includes(:member).order(created_at: :desc)
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


jim (commented):
This looks great!

I don't think we use `app/views/admin/loan_histories/show.html.erb` anywhere else- or more accurately, I don't think we have any other links that go to the controller that renders it. So it should be safe to remove that file and the associated controller and route now that we have an improved version in place.

There used to be a few funky older controllers like [`Admin::LoanHistoriesController`](https://github.com/chicago-tool-library/circulate/blob/main/app/controllers/admin/loan_histories_controller.rb) that were trying to serve multiple purposes at once, but we've mostly replaced them with separate controllers (and corresponding views) at this point.

katielentes:
Thanks @jim! I've removed the old layout, associated controller, and route.

jim (approved):
🚀 Awesome!

PR #1480, opened 2024-05-04
Add standardjs check to CI

jim:
# What it does

We added standard to check our front end code in #1472. This extends our existing CI checks to also run that linter.


phinze (approved):
Nice hopefully this will help us keep JS stuff more consistent 👨‍⚖️

PR #1485, opened 2024-05-09
Convert rotation to integer when fetching variant

phinze:
Vips does not like strings


phinze:
Skipping CI since we're time sensitive and I'm testing on staging

PR #1488, opened 2024-05-10
Add reservation index and show pages

jim:
# What it does

Builds out a more usable index of reservations and the start of a reservation UI. These are modeled on the existing tables we have in the app and the "profile" page layout that we use on item/member/item pool pages.

I also moved the pickup edit screen into a tab on the reservation page.

# UI Change Screenshot

Screenshot 2024-05-09 at 11 30 44 PM

Screenshot 2024-05-09 at 11 30 54 PM

Screenshot 2024-05-09 at 11 30 35 PM


phinze (approved):
LGTM! 🚀

PR #1489, opened 2024-05-11
Flatten Pickup into Reservation

jim:
# What it does

This is a stacked PR that should be merged after #1488 lands. The new code is in 44e09e6b3af6cad9b049db07ccdb5428a7c3d489.

After looking more closely at how both Reservation and Pickup had status columns and thinking about how we'd model the logic between the two of them, I decided that it's going to be best to have a single status field track the entire reservation lifecycle. Accordingly, the Pickup model doesn't provide much except a level of indirection.

Note that I went the brute force route in the migration. It doesn't seem worth it to write a data migration for any existing Pickup or ReservationLoan data this early in development.


phinze (approved):
This makes sense to me - it seems more straightforward to model this as just one state machine as opposed to two interlocking ones.

PR #1490, opened 2024-05-11
Use a state machine to model Reservation status

jim:
# What it does

This is a stacked PR that should be merged after #1489. The new code is:

* Initial Reservation state machine added in a6f1d6e28cab9beff92ee970697c0830c1f46225
* Updates to the UI to drive it from the state machine in a62dbef17e8c86b195643de0cf22a22c30a8e988 and 52b0feb51865da1027b891188890854e50fe0597)

# Implementation notes

I wanted to try to model the states and transitions for a reservation in a single place. It's a somwhat linear path, with a few points of branching. We'll always want to preserve an escape hatch for strange situations, but I think this might be a good way to consolidate the logic of what can happen when throughout a Reservation's lifecycle.

I'm fairly happy with this approach so far. It remains to be seen how the state machine and other conditionals will interact- for example, what if you have to have added a certain number of associated items before it's OK to transition to a new state?


phinze (approved):
This is looking nice and clean so far - I like the direction!

Issue #1475, opened 2024-04-30
Add pagination to account holds history


jim:
A member's hold history page within their account (`/account/holds/history`) isn't paginated, which means for long-time members this page is very long (and slow to load).

We should add pagination using [pagy](https://github.com/ddnexus/pagy) (which is what we already use elsewhere).

PR #1491, opened 2024-05-12
Added pagination to account holds history

crismali:
# What it does

Adds pagination to the accounts hold history page.

# Why it is important

This will prevent long page load times for long time users. See #1475. 

# UI Change Screenshot

Screenshot 2024-05-12 at 3 01 48 AM

# Implementation notes

I wasn't sure how many holds should be shown on each page, so I just left it as the pagy default. 

The test I added might be considered redundant, but I figured I'd leave it in for now. It technically doesn't test the pagination functionality itself. I'm happy to make it do so, but I decided to hold off after I couldn't find a similar test for the other uses of pagy in the app.

I also left a small change in the Gemfile.lock regarding the installation of Nokogiri on laptops like mine.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -326,6 +326,8 @@ GEM
     nio4r (2.7.0)
     nokogiri (1.16.4-arm64-darwin)
       racc (~> 1.4)
+    nokogiri (1.16.4-x86_64-darwin)
+      racc (~> 1.4)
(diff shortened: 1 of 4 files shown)


jim (approved):
This looks good to me!

jim:
I think it's OK to not test the pagination directly- in theory the Pagy library should do that itself.

Issue #1251, opened 2024-01-21
Add filters to users page under Settings


jim:
We need to have filters on the users page so that admins can see only admins, only staff, or everyone. These can be done similarly to the filter we have on the admin member's index (`/admin/members`).

PR #1495, opened 2024-05-16
Added filtering by role to the users index page for admins

crismali:
# What it does

These changes allow admin users to filter the users index page in the admin interface by role. By default they'll see all users, but they can choose admin only, staff only, or members only.

# Why it is important

This should make some admin tasks easier when it comes to users. See #1251 for more. 

# UI Change Screenshot

With the members only filter active:
Screenshot 2024-05-15 at 9 46 00 PM

With the admin only filter active:
Screenshot 2024-05-15 at 9 45 41 PM

With no filters active (ie all users):
Screenshot 2024-05-15 at 9 45 29 PM

# Implementation notes

The card didn't specifically mention being able to filter by the member role but I included it anyway. ~~Not sure if pagination or sorting would be helpful here as well.~~ I just went through the pagination card for this page (#1496).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -326,6 +326,8 @@ GEM
     nio4r (2.7.0)
     nokogiri (1.16.4-arm64-darwin)
       racc (~> 1.4)
+    nokogiri (1.16.4-x86_64-darwin)
+      racc (~> 1.4)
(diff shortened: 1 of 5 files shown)


jim (approved):
This is awesome! The test coverage is nice to see 🚀 !

Issue #1249, opened 2024-01-21
Add pagination to Users page under Settings


jim:
The "Users" page (which can be found at `/admin/users` or in the admin menu under "Settings") lists all users on a single page 😱 ! We need to add pagination to this page so that we don't load all of the users at once.

We're currently seeing 2 second response times for this page:

![Image](https://github.com/chicago-tool-library/circulate/assets/3331/a7a2e707-7386-4711-80fe-1df094b08337)

See also #1251

PR #1496, opened 2024-05-16
Added pagination to the admin users index page under settings

crismali:
# What it does

Adds pagination with pagy to the users index page under settings.

# Why it is important

Currently this page takes just over 2 seconds to load so this should speed things up significantly. See #1249 for more.

# UI Change Screenshot
Screenshot 2024-05-16 at 10 55 35 AM

# Implementation notes

Just as basic a pagy implementation as I could manage. I also worked on #1495 which involves filtering this same page. Once one of these is merged it might take a little work on the other to make sure the filters and pagination don't cause any issues with one another, which I'm happy to take care of. Alternatively, these two could be combined? Depends on how useful one is without the other.

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -1,10 +1,12 @@
 module Admin
   class UsersController < BaseController
+    include Pagy::Backend
+
     before_action :set_user, only: [:show, :edit, :update, :destroy]
     before_action :require_admin
 
     def index
-      @users = users_index_query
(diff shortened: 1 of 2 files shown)


jim (commented):
This change looks good! Like you said, it'll just need to be adjusted a bit due to merge conflicts with the other PR.

phinze (approved):
Looks great, thanks @crismali! I'll get this deployed shortly.

PR #1502, opened 2024-05-20
Prevent same-day drop offs and pickups

### LINKED ISSUE #1258: Prevent same-day drop offs and pickups
Based on [Paul's comment](https://github.com/chicago-tool-library/circulate/pull/1248#pullrequestreview-1837413463), we should explore how to prevent folks from scheduling pickups and drop offs on the same day.

katielentes:
# What it does

Adds validation on `Reservation` to prevent `ended_at` from being before or the same day as `started_at`. Adds front end validation on the Reservation form to disable selecting invalid dates.

# Why it is important

Closes #1258 

# UI Change Screenshot
Screenshot 2024-05-19 at 8 53 42 PM
Screenshot 2024-05-19 at 8 54 12 PM

# Implementation notes
To make the front end validation work, I just added a script tag in `app/views/admin/reservations/_form.html.erb`. I saw the JS controller pattern using stimulus and attempted to re-create it, but had trouble getting the controller to connect to the view. I included this attempt with my console.logs.

If anyone can help me figure out what I'm missing I'd greatly appreciate it.

Also, if the front end validation is overkill I am happy to remove it! Open to any and all feedback.


app/javascript/controllers/reservation_dates_controller.js
```diff
@@ -0,0 +1,19 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Connects to data-controller="reservation-dates"
+export default class extends Controller {
+  targets = ["startDate", "endDate"]
```
jim, line 5: I poked at this Stimulus controller for a bit. It looked correct to me, but after some head scratching and doc reading I noticed that the `static` keyword was missing before the `targets` declaration. Making this change allowed the targets to be properly loaded (and accessed within controller methods via `this.startDateTarget`, etc.):

```suggestion
  static targets = ["startDate", "endDate"]
```

jim (commented):
The reservation validation looks good to me.

I like having some client side validation on this- we tend to do most of that on the server, but for something like this it's such a better experience to avoid the round trip before telling folks that they picked something invalid. Thanks for taking the initiative to get something setup for that!

I do think it would be nice to get the JS moved into that controller before merging this.

app/javascript/controllers/reservation_dates_controller.js
katielentes: Thanks for your help here - I definitely needed to add the `static` keyword! I also realized the reason I didn't see any of my console.logs in the conntect function or changes reflected after creating this controller was due to needing to run `rails assets:precompile`.

jim (approved):
This looks good. Thanks for refactoring it into a Stimulus controller!

app/javascript/controllers/reservation_dates_controller.js
jim: That would do it! In case you hadn't encountered it, using `bin/dev` to run the server and also keep the JS and CSS up to date can be handy.

katielentes: Ah thank you for that! I had just been using `rails s`!

PR #1503, opened 2024-05-20
Update nix setup to support vips

phinze:
# What it does

Pulls dependencies of vips into scope in the Nix development setup.

# Why it is important

For those of us using the Nix dev setup (only me 🙃) this makes image processing work locally.

# Implementation notes

 * Switches upstream to `nixpkgs-unstable` because we tend to track pretty closely with upstream dependencies on the Ruby and JS side thanks to dependabot, so it's helpful for the underlying libraries to stay tight with upstream as well.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -54,16 +54,16 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1710021367,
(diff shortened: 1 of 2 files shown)


phinze:
Only me using this so self-merging

PR #1506, opened 2024-05-22
Edit reservation holds in place

jim:
# What it does

Adds the ability to edit the items (`ReservationHolds`) on a reservation in place in the table. This is likely going to be the way that folks work with these models as we build out other UIs as well.

Validations are shown on the page properly. There is some UI cleanup work to do, of course- this is still rough.

# UI Change Screenshot


app/controllers/admin/reservations/item_pool_searches_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Admin
+  module Reservations
+    class ItemPoolSearchesController < BaseController
```
jim, line 3: This controller is mostly a clone of `Admin::ItemPoolsController`. If we add more/better searching we can do it in such a way that both views are improved.

app/controllers/application_controller.rb
```diff
@@ -85,4 +85,13 @@ def after_sign_in_path_for(user)
       referer || default_path
     end
   end
+
+  # Render a turbo-stream response without all the boilerplate
+  def render_turbo_response(*args)
```
jim, line 90: We don't need to support non-JS client in the admin, but we should be careful about depending on it on public pages.

app/controllers/admin/reservations/item_pool_searches_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+      include Pagy::Backend
+
+      def show
+        item_pool_scope = ItemPool.all
+
+        if params[:category]
+          @category = CategoryNode.where(id: params[:category]).first
```
crismali, line 10:
```suggestion
          @category = CategoryNode.find_by(id: params[:category])
```
I think this syntax is preferred but it honestly makes no difference

app/controllers/admin/reservations/item_pool_searches_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+          redirect_to(items_path, error: "Category not found") && return unless @category
+
+          item_pool_scope = @category.item_pools
+        end
+
+        item_pool_scope = item_pool_scope.includes(:categories).order("name ASC")
```
crismali, line 16:
```suggestion
        item_pool_scope = item_pool_scope.includes(:categories).order(name: :asc)
```
Similar to the above. I've been on teams that enforce it one way or the other so it jumps out at me 😅

crismali (approved):
Looks solid!

app/controllers/admin/reservations/item_pool_searches_controller.rb
jim: I don't think we've been too strict about this one way or the other so far... if there's a linter we could use to enforce this kind of thing, that would be nice!

PR #1511, opened 2024-05-28
Update donation link

phinze:
# What it does

Changes the donation link to a new one provided by @tessavierk


phinze:
Self-merging as this is just a link change and it's important we get it deployed soon.

PR #1514, opened 2024-05-29
Consolidate setup and reset scripts

### LINKED ISSUE #1487: Consolidate bin/setup and bin/reset
We have both `bin/reset` and `bin/setup` scripts. They are mostly the same, and the difference between then (as when to use each one) is not obvious.

Let's combine these into a single script, `bin/setup`. I think we can mostly just use what's in `bin/reset` for this, but it's worth making sure that it works when there isn't yet a database, etc.

jim:
# What it does

Combines `bin/reset` and `bin/setup` into a single idempotent script. `setup` is the default Rails one, and they were minimally different. This way, it's always the same command to get to a workable baseline.

Fixes #1487.

### DIFF
--- a/bin/reset
+++ b/bin/reset
@@ -1,31 +1,9 @@
 #!/usr/bin/env ruby
-require 'fileutils'
 
-# path to your application root.
-APP_ROOT = File.expand_path('..', __dir__)
+puts <<~WARNING
+    !!! WARNING !!!
+    bin/reset is deprecated and will be removed! Please use bin/setup to reset your environment.
+WARNING
+puts
 
-def system!(*args)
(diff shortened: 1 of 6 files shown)


bin/reset
```diff
@@ -1,31 +1,9 @@
 #!/usr/bin/env ruby
-require 'fileutils'
 
-# path to your application root.
-APP_ROOT = File.expand_path('..', __dir__)
+puts <<~WARNING
+    !!! WARNING !!!
```
crismali, line 4: 🔥

crismali (approved):
Looks great!

mdecelle (approved):

PR #1517, opened 2024-05-31
Revert "Update donation link"

phinze:
Reverts chicago-tool-library/circulate#1511


phinze:
On request from @tessavierk we're to putting this back now, as the other link is for a specific purpose

Issue #1542, opened 2024-06-12
Create Notification record for Devise interactions


crismali:
We create `Notification` records to track when we email members, but we don't do this for emails generated by Devise (such as password reset emails). It would be nice to track these so when someone says they didn't get an email, we can verify that it did or didn't go out (and when that happened if it did).

PR #1543, opened 2024-06-12
Create Notifications whenever Devise sends an email

crismali:
# What it does

This creates `Notification`s whenever the Devise mailer is used (`UserMailer`). It uses the set `@subject` when present and falls back to the mail object's subject.

# Why it is important

This takes care of [#1542](https://github.com/chicago-tool-library/circulate/issues/1542) (this was a draft issue when I started so I converted it into an issue).

# UI Change Screenshot

Nothing much to see here.

# Implementation notes

I used the `@subject` instance variable as the default but fell back to `mail.subject` because the emails that are not overridden lack `@subject`. 

When I was first looking at this I assumed that Devise or Rails just automatically picked up on `@subject` and used that as the email's subject line but that doesn't seem to be the case. The email's actual subject in this case is "Reset password instructions" and `@subject` is only used for the `Notification`. Not calling this out because it's wrong or anything, it just confused me for a moment so I wonder if maybe there's a better name or if the original intent was for `@subject` to be the email's subject line.

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -2,19 +2,13 @@ class UserMailer < Devise::Mailer
   default from: "Chicago Tool Library <team@chicagotoollibrary.org>"
   after_action :store_notification
 
-  def reset_password_instructions(record, token, opts = {})
-    @user = record
-    @subject = "You requested a password reset"
-    super
-  end
-
   def store_notification
     Notification.create!(
-      member: @user.member,
-      uuid: SecureRandom.uuid,
       action: action_name,
       address: @user.email,
-      subject: @subject.presence || mail.subject
+      member: @user.member,
+      subject: mail.subject,
+      uuid: SecureRandom.uuid
     )
   end
 end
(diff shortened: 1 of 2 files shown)


phinze (changes requested):
Nice catch on the `@subject` stuff! It looks like it's all controlled from `devise.en.yml` which [explains the subject you saw](https://github.com/chicago-tool-library/circulate/blob/df98b6caec63f2514f72bc8587f821963f871c45/config/locales/devise.en.yml#L22-L23).

Since we know it's not working, do you think you can remove the `@subject` stuff altogether instead of including it in a conditional? It looks like [Devise sets `@user` as well](https://github.com/heartcombo/devise/blob/a259ff3c28912a27329727f4a3c2623d3f5cb6f2/lib/devise/mailers/helpers.rb#L24) so I think that entire overridden method can go away now!

crismali:
Sounds good!

phinze (approved):
Looking good 😎 will get this deployed

Issue #665, opened 2021-09-16
Staff can still put items on hold or check things out when members have overdue items.


CTLvivek:

CTLvivek:
I'm assuming this title (originally created by Jim) is the desire to hardcode this policy rule so that it's easier for librarians to follow and enforce. If that's the right understanding, I'd say that at the moment it's better to have the flexibility but I'm sure it'd be fine either way.

jim:
We want to allow staff to check out and put things on hold regardless of if a person has things that are overdue. The system currently doesn’t allow this, and it’s more annoying than helpful.

PR #1544, opened 2024-06-13
Staff can create holds or loans for members even if that member has overdue loans

crismali:
# What it does

This allows anyone with staff/admin access to create a hold or a loan for a member that already has an overdue loan. It removes a check that prevented this.

# Why it is important

It addresses [#665](https://github.com/chicago-tool-library/circulate/issues/665) by removing a roadblock for staff.

# UI Change Screenshot
Before this change, staff would see this toast message regarding the overdue loan.

Screenshot 2024-06-13 at 11 26 32 AM

After this change, staff will see the usual item "Lend or Hold Item" form.

Screenshot 2024-06-13 at 11 26 14 AM

# Implementation notes

I just exposed the form in the HTML and didn't have to change any validations or controller checks.

### DIFF
--- a/app/views/admin/members/_lookup.html.erb
+++ b/app/views/admin/members/_lookup.html.erb
@@ -1,15 +1,6 @@
 <div class="columns member-lookup-items">
   <div class="column col-12">
     <h2 id="checkout" class="h4">Lend or Hold Item</h2>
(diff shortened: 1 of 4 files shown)


phinze (approved):
LGTM! Makes sense that staff should be able to manipulate holds/loans regardless of the member's situation.

Issue #1527, opened 2024-06-05
Reservation validations can be configured


jim:
We want the specific values used for validations in #1463 to be configurable in the admin. We can probably store these fields on the library for now. It might be worth prefixing them with "reservation_" so that their use is obvious?

The values we'll need are:
* maximum reservation length
* minimum reservation start distance
* maximum reservation start distance

Issue #1463, opened 2024-04-22
Add additional validations for reservations


jim:
Here are the validations we'd like to add:

* Limits reservation duration to 21 days
* Require that reservations start no sooner than two days in the future (we do want to use days here and not hours, so that the soonest a reservation can be created for on Tuesday is Thursday).
* A reservation can't start more than 3 months in the future.

PR #1545, opened 2024-06-13
Allow admins to configure reservation starts and ends at through reservation policies

crismali:
# What it does

~~I added 3 non-nullable columns with defaults to the `libraries` table to store the configurations (`maximum_reservation_length`, `minimum_reservation_start_distance`, and `maximum_reservation_start_distance`), `Library#valid_reservation_started_at?`, and validations for `Library` and `Reservation`.~~

Adds a `ReservationPolicy` model with CRUD that admins can configure to set acceptable reservation dates per `ItemPool`. Right now they show up as toast message warnings on the `Reservation` details page when the reservation has a `status` of `requested`, `approved`, or `building`.

# Why it is important

Addresses #1527 and #1463 by allowing ~~super~~ admins to configure some validation rules around reservation start and end days.

# UI Change Screenshot

Reservation policies index: 
index

Reservation policies show: 
show

Reservation policies edit: 
edit

Item pool edit: 
edit-item-pool

Item pool show: 
show-item-pool

Reservation with many warnings:
show-reservation-warnings

# Implementation notes

~~Adding the validations to `Reservation` seemed like the most straightforward approach, but I wasn't sure if anything else should be factored in, like the `Reservation#status` or if the `Reservation` has already ended based on its start and end days.~~

~~Also, there were a few templates that didn't have `.html` as part of their file extension which caused some surprising behavior when adding and editing forms so I renamed them, but maybe they were named that way to do something I didn't catch. Not sure, but figured it might be worth calling out.~~

I tried to match `BorrowPolicy` as much as I could (ie how there can only be one `default` policy at a time, a description, associated with a library). I added a `ReservationDateValidator` to `app/lib` to contain all of the logic around comparing reservations and reservation policies and tried to call out all of the possible issues that could come up. 

Mentioned above, but the warnings will only show up when the reservation's `status` is `requested`, `approved`, or `building`. We can always adjust those statuses since this was my best guess at when it would be most relevant.

Any feedback is appreciated!


phinze:
This is a good start!

The fields might be something we want to store on `BorrowPolicy` instead of on the `Library` directly, though borrow polices are not yet wired into the group lending part of the code base.

Since @jim is a lot more active in this area I’m going to leave it to him to review this.

jim (approved):
If you're up for it, I think it would be worth trying to put the new values in a `BorrowPolicy`-esque model that `has_many :item_pools` instead of on the `Library`. I realize the original ticket did say exactly what you did here, so props on following those instructions to a tee 😄 .

Also, it's really awesome to see the tests you added. It's always great to see our coverage improve.

jim (approved):
Cool! I read through everything and clicked around and it works well! I think we'll likely iterate a bit on when and how we surface this information, but the abstractions here should allow us to express the rules for items properly.

Issue #1546, opened 2024-06-14
Log the user_id during signup


crismali:
We have logging setup in Appsignal so that we can see how members use the application. This is helpful when we need to troubleshoot issues.

However, the current approach relies on logging the value of `current_user`, which means that during signup, we get empty values logged (users aren't technically logged in during the signup flow):

```
10-15-2023 18:31:45.760[application][INFO][web.1][rails]method=POST path=/signup/documents/acceptance format=js controller=Signup::AcceptancesController action=create status=200 allocations=2052 duration=10.17 view=0.00 db=4.78 location=https://app.chicagotoollibrary.org/signup/payments/new user_id= request_id=ffc95285-d7ea-4182-8745-ab070b6f4dd2
```

It would be really useful to have the `user_id` of the current user logged.

PR #1547, opened 2024-06-14
Log the user id when available during the sign up flow

crismali:
# What it does

This adds the user_id to lograge during the signup process by falling back to the `user` associated with the controller's `member` (if member is defined).

# Why it is important

This addresses [#1546](https://github.com/chicago-tool-library/circulate/issues/1546).

# UI Change Screenshot

It's small, but you can see the `user_id` is `17` in the picture for the `/signup/documents/agreement` path.

Screenshot 2024-06-14 at 5 01 14 PM

# Implementation notes

`@member` was being set in the signup controllers so I just added an `attr_reader` for it. I leaned on `try` pretty heavily so controllers that don't implement `member` won't be impacted.

### DIFF
--- a/app/controllers/signup/base_controller.rb
+++ b/app/controllers/signup/base_controller.rb
@@ -3,6 +3,8 @@ module Signup
     before_action :load_steps
     before_action :set_page_title
 
+    attr_reader :member
+
     layout "steps"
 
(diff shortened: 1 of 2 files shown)


phinze (approved):
LGTM!

Issue #1507, opened 2024-05-22
Fix code scanning alerts


jim:
Tracking issue for:
- [ ] https://github.com/chicago-tool-library/circulate/security/code-scanning/1
- [ ] https://github.com/chicago-tool-library/circulate/security/code-scanning/2

The code is actually safe in that it is checking that the values are in a set of allowed values, but we may as well update the code to avoid security warnings like this.

PR #1548, opened 2024-06-14
Prevent codescanning warnings by ensuring methods return constant values and not technically user input

crismali:
# What it does

Refactors some filtering methods to use `case` statements and constants to so that user input is never technically passed to the `send` method.

# Why it is important

Addresses [#1507](https://github.com/chicago-tool-library/circulate/issues/1507). Even if the code was safe before, it's nice to have less noise.

# UI Change Screenshot

Nothing to see here.

# Implementation notes

I added an `INDEX_FILTERS` constant so all of the values are grouped in an array but nothing uses it. I've seen other patterns for managing constants and kind of just went with this one, but if there's a preferred way to do this I'm happy to make those changes.

### DIFF
--- a/app/controllers/admin/loan_summaries_controller.rb
+++ b/app/controllers/admin/loan_summaries_controller.rb
@@ -13,10 +13,23 @@ module Admin
 
     private
 
+    INDEX_FILTERS = [
+      CHECKED_OUT = "checked_out",
+      OVERDUE = "overdue",
+      RETURNED = "returned"
+    ]
+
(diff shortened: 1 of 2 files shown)


phinze (approved):
Ah thanks I’ve been meaning to get around to these 😌 

I don’t think it’s worth bikeshedding the constant storage pattern here since it’s got such a tiny surface area. The way you have it set up is a reasonable improvement from where it was so let’s ship it! 🚢

PR #1553, opened 2024-06-18
Switch to GoodJob for background work

phinze:
# What it does

* Switches job system from [Sucker Punch](https://github.com/brandonhilkert/sucker_punch) to [GoodJob](https://github.com/bensheldon/good_job)

# Why it is important

* GoodJob gives us a few implementation benefits over Sucker Punch
  - It's Postgres-backed, so in-flight jobs won't be lost on server restart
  - It can be configured with an external worker, which will reduce memory pressure on the web dyno
  - It includes scheduled job features, which will let us migrate from Heroku Scheduler to jobs defined in code
  - It includes a dashboard we can use to debug potential issues with background jobs

# Implementation notes

* We will now have a persistent worker dyno running in both staging and production, which will add $14/mo ($168/yr) to our hosting costs. Initial discussion amongst the team has concluded this is a worthwhile tradeoff for the above benefits.
* The Heroku Scheduler to GoodJob Cron conversion will come in a future PR, assuming this transition goes well.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -114,7 +114,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:11
+        image: postgres:15
         env:
           POSTGRES_HOST: localhost
           POSTGRES_USER: postgres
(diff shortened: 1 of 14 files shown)


jim (approved):
LGTM! I'm excited to have. real background work system.

Issue #1556, opened 2024-06-21
Square: implement webhooks for payments received


jim:
There have been a couple instances over the past few months where someone was charged by Square for a membership but this wasn't recorded in the Circulate database. The current implementation relies on someone being redirected from the Square checkout page back to Circulate for a `Membership` to be created and the payment to be recorded. We don't have great logs for this (yet), but the most likely cause is that the member didn't get redirected successfully back to Circulate.

Square supports sending webhooks to applications when various events occur. To solve the above issue and make the app more resilient, we should implement a webhook so that we can perform the various post-payment actions more reliably and not rely on the user's browser to complete payment flow processing.

To work on this, you'll need to setup a developer account with Square.

jim:
Attempted this earlier in 2024 and ran into a bunch of issues. Rolled the change back while we decide how to proceed.

PR #1557, opened 2024-06-21
Extend signup timeout to 2 hours from 30 minutes

jim:
# What it does

We had an instance of a member taking longer than 30 minutes to complete the signup steps, and it timed out just as they were paying for their membership. You can see [the sample of this occurrence in AppSignal](https://appsignal.com/chicago-tool-library/sites/60596f9214ad662e17191689/performance/incidents/59/samples/60596f9214ad662e17191689-158228531489123241821718233200?selectedTime=2024-06-12T23%3A05%3A00.000Z&activeFilter=date). This meant that the callback from Square was rejected and their membership payment wasn't recorded correctly in our system. The logic to calculate if the time had passed was also hard to follow as it was previously implemented.

This change gives people 2 hours to get through signup. The original time was chosen arbitrarily, and in hindsight I can't say why it needs to be so short. I'm open to making it even longer; the main risk I can think of is on shared computers where folks might enter some of their info, walk away, and leave the window open for someone else to complete.

The other part of a complete solution to this problem is to implement web hooks for payments, which it captured by #1556.

### DIFF
--- a/app/controllers/signup/base_controller.rb
+++ b/app/controllers/signup/base_controller.rb
@@ -1,5 +1,7 @@
 module Signup
   class BaseController < ApplicationController
+    SIGNUP_TIMEOUT = 2.hours
+
     before_action :load_steps
     before_action :set_page_title
 
@@ -16,7 +18,7 @@ module Signup
     private
 
     def load_member
-      if session[:member_id] && session[:timeout] && session[:timeout] > Time.current - 15.minutes
+      if session[:member_id] && session[:timeout] && session[:timeout] > Time.current
         @member = Member.find(session[:member_id])
       else
         reset_session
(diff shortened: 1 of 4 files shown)


app/controllers/signup/base_controller.rb
```diff
@@ -16,7 +18,7 @@ def is_membership_enabled?
     private
 
     def load_member
-      if session[:member_id] && session[:timeout] && session[:timeout] > Time.current - 15.minutes
```
jim, line 19: This logic was really hard for me to wrap my head around. Now, `session[:timeout]` contains a time, after which the current session is considered invalid.

crismali (approved):
Looks great to me!

PR #1558, opened 2024-06-21
Remove one day gap when renewing a membership

jim:
# What it does

We have an interesting bug where the system creates renewal memberships that start one day after their current membership ends. This creates a strange situation where someone can have an expired membership, but their new one hasn't yet started, so parts of the system just say they need to renew their membership despite them already doing so.

To fix this, I removed the one day gap, which was being explicitly added.


phinze (approved):
Code looks good! That system test failure might be legit though

PR #1559, opened 2024-06-22
Add a member-facing reservation booking workflow

jim:
# What it does

Adds a new flow that allows members to create reservations.

## Implementation Notes

* This just assigns reservations to the first `Organization`, we'll want to make that the current user's organization (or perhaps they can choose which one) when we have organization members.
* There is some UI jankiness that we'll iron out over time.
* Some basic system tests are included for this flow.

## UI Changes

Screenshot 2024-07-08 at 9 52 14 PM


phinze (approved):
Looking good! Just smoke tested this locally and the scaffolding all works for me!

PR #1567, opened 2024-06-25
Install Blazer for custom analytics dashboards

phinze:
# What it does

* Installs the [Blazer](https://github.com/ankane/blazer) gem
* Admins can now visit the Blazer dashboard at `/blazer`

# Why it is important

* Let non-Heroku-admins arbitrarily query the database
* Lets app admins create dashboards
* Opens up ability for us to explore the data we're recording with Ahoy

# UI Change Screenshot

2024-06-25 at 11 30 01@2x

# Implementation notes

* Blazer recommends creating a readonly user for it to use, but our Heroku database plans do not allow us anything beyond the default user, so we'll have to rely on Blazer's transaction wrapping and our own carefulness to avoid losing data.
* Blazer has several other features controlled via its config file that we can explore once we've had a chance to play with it.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -106,3 +106,5 @@ gem "twilio-ruby", "~> 7.1"
 gem "ahoy_matey", "~> 5.1"
 
 gem "good_job", "~> 3.29"
+
+gem "blazer", "~> 3.0"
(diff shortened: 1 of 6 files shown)


jim (approved):
I'm excited to see how we can expose some of the data we are collecting. It's too bad about Heroku's database credential limitations, but hopefully the transaction wrapping that blazer includes will prevent any disasters.

PR #1570, opened 2024-06-25
Block all bots in robots.txt

### LINKED ISSUE #1569: Add robots.txt to block web crawlers

As we're chasing down both InvalidAuthenticityToken and MemoryQuotaExceeded errors, it would be good to filter out bot traffic as much as we can.

I think we can try a [robots.txt](https://www.robotstxt.org/robotstxt.html) that entirely blocks bots for now. I don't believe we get any massive benefits by having search engines indexing the whole tool library. We can let bots index the squarespace site for our search engine presence.

phinze:
# What it does

Resolves #1569 

# Why it is important

Removes source of potential noise from logs and traffic.

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,1 +1,2 @@
-# See http://www.robotstxt.org/robotstxt.html for documentation on how to use the robots.txt file
+User-agent: *
+Disallow: /


jim (approved):
Good call!

PR #1572, opened 2024-06-29
Process Square payments using webhooks

jim:
# What it does

Replaces the way we handle Square payments (and ultimately create `Membership` objects) with one based on accepting webhook notifications from Square.

This should improve reliability (we don't need someone to follow the redirect back form Square to record their payment) and observability (we can now track these jobs with GoodJob to know if/when they fail, etc.)

# Why it is important

We've had a number of issues over the years with folks not making it back to our site from Square and Circulate then being out of sync with Square.

# Implementation notes

* There are a bunch of new `Rails/EnvironmentVariableAccess` errors from standard-rails that I'll need to sort out.

# Deployment notes

* I'd like to make sure that staging is setup with the sandbox so we can test the payment flow there.
* We need to configure values for `SQUARE_PAYMENT_WEBHOOK_NOTIFICATION_URL` and `SQUARE_PAYMENT_WEBHOOK_SIGNATURE_KEY` in both staging and production.


app/controllers/renewal/payments_controller.rb
```diff
@@ -36,35 +36,7 @@ def skip
...
-        session[:attempts] += 1
-        if session[:attempts] <= 10
-          render :wait, layout: nil
-          return
-        end
-      end
-
-      Rails.logger.error(errors)
-      session.delete(:attempts)
-      flash[:error] = "There was an error processing your payment. Please come into the library to complete signup."
```
jim, line 67: This logic has all moved to `SquarePaymentJob`- we just render the confirmation page here and don't attempt to process the payment further.

app/controllers/signup/payments_controller.rb
```diff
@@ -18,10 +18,14 @@ def create
 
       result = checkout.checkout_url(amount: @form.amount, email: @member.email, return_to: callback_signup_payments_url, member_id: @member.id, date: Date.current)
       if result.success?
+        if Rails.env.development?
+          # It takes the sandbox a while to be ready
+          sleep 2
+        end
```
jim, line 24: This is nasty, but I couldn't come up with a better way to make testing against the sandbox work more than 1/10 of the time.

app/lib/square_checkout/client.rb
```diff
@@ -0,0 +1,109 @@
...
+              base_price_money: {
+                amount: amount.cents,
+                currency: "USD"
+              }
+            }],
+
+            metadata: {
+              member_id: member_id.to_s,
+              created_by: "circulate"
+            }
```
jim, line 49: All payments (including those made using the POS system to sell shirts) will send us web hooks, so we use this metadata to detect if payments were created by our app. If they weren't, we ignore them in the webhook handler.

config/environments/development.rb
```diff
@@ -71,6 +71,7 @@
   config.file_watcher = ActiveSupport::EventedFileUpdateChecker
 
   config.hosts << ".circulate.local"
+  config.hosts << /[a-z0-9-.]+\.ngrok-free\.app/
```
jim, line 74: I use ngrok when testing this, and figure others might as well. Open to other services as well if they' assist with local dev.

app/jobs/square_payment_job.rb
```diff
@@ -0,0 +1,47 @@
...
+  class MetadataMissingError < StandardError; end
+
+  class OrderLookupError < StandardError; end
+
+  def perform(order_id:)
+    existing_adjustment = Adjustment.find_by(square_transaction_id: order_id)
+    if existing_adjustment
+      Rails.logger.info("Adjustment already exists for order #{order_id}, skipping.")
+      return
+    end
```
jim, line 11: We might need a better lock here in case a message is delivered more than once close enough together that it's getting processed by two workers at the same time. [The docs say](https://developer.squareup.com/docs/webhooks/overview#requirements-and-limitations):

> **Webhooks can be sent more than once.** You can bypass the processing of repeated notifications using the idempotency value included as the event_id field in the body of each event notification.

phinze:
> There are a bunch of new Rails/EnvironmentVariableAccess errors from standard-rails that I'll need to sort out.

It looks like [standard-rails disabled this recently](https://github.com/standardrb/standard-rails/issues/46), along with a few other rules we have in our current todo list. I'll bump us to their main branch temporarily to sidestep this extra work.

jim:
@phinze awesome, much appreciated!

app/views/renewal/confirmations/show.html.erb
```diff
@@ -7,8 +7,8 @@
     <div class="verification-instructions">
       <p>Thank you for renewing your Chicago Tool Library membership.</p>
 
-      <% if @amount %>
-        <p><strong>Your payment of <%= @amount.format %> has been processed and applied to your account.</strong></p>
+      <% if @payment %>
+        <p><strong>Your payment has been processed and applied to your account.</strong></p>
```
crismali, line 11: Should this be something like "Your payment is being processed and applied to your account." in case the job fails?

app/controllers/signup/payments_controller.rb
crismali: 🍳

crismali (approved):
Looks good to me! Just a small copy question

app/views/renewal/confirmations/show.html.erb
jim: Oh yes, that is a lot better! It also handles the fact that there might be a small delay which could happen for a variety of reasons.

app/controllers/signup/payments_controller.rb
jim: I'm also realizing that this won't handle staging correctly, so I guess I'll make this an env var that we can set in each environment.

Issue #1513, opened 2024-05-29
As an admin, I can specify questions that will be asked when creating a reservation


jim:
To start, these questions can be of two types: those with text answers and those with number answers (which must be integers).
We'd like these to be customizable, so they'll need to be stored in a `JSON` column on `Library`.

PR #1576, opened 2024-07-03
Admins can create questions that are answered when creating/updating reservations

crismali:
# What it does

Allows admins to create, update, archive, and unarchive questions that are answered when creating or updating a reservation. The questions have all of the usual CRUD features while answers can be viewed under a tab on the reservation's show page.

# Why it is important

Addresses #1513

# UI Change Screenshot

Questions' index
list questions

New Question
new question

Edit Question
edit question

Show Question
show question

Show archived Question
show archived question

New Reservation
new reservation

Edit Reservation
edit reservation

Show Reservation
show reservation

# Implementation notes

I've added question, stem, and answer models (["stem" is the term for the question or prompt text in a multiple choice question](https://cft.vanderbilt.edu/guides-sub-pages/writing-good-multiple-choice-test-questions/)). 

Right now all unarchived questions will appear when creating/updating a reservation with their most recently created stem (stems are never updated).

I leaned a lot on nested attributes for creating new stems and creating/updating the answers but it's been a while since I've used them so there might be better implementations (or different approaches altogether).

The `jsonb` `result` column on answers should be flexible enough to support multiple choice questions later on.

The things that seem the oddest to me are:

- the `result` column name (I didn't want there to be situations where something like `answer.answer` is possible and as a generally rule I try to avoid anything "response")
- the `answer_type` lives on the stem so the answer has to talk to the stem quite a bit. I thought about copying it over but that seemed redundant. Plus these models are all pretty small right now.

Always open to any and all feedback!


jim:
Cool! This is great in that it handles questions being re-phrased nicely and will make aggregating answers fairly straight forward. I wonder how we might deal with the order of the questions and removing a question that we no longer want? I imagine we can add `active` and `position` columns when we get there, but wanted to mention it while you're sorting out the data model.

Also TIL about the term `stem`. I don't think that the tool library folks specifically requested multiple choice questions, but it does seem like a really nice feature to have (and will make things like . I think we can probably get away without needing to track the versions of stems to start and just assume that we don't edit them to change their meaning once they are created.

While we wait [to get feedback from tool library staff on specific questions](https://chicagotoollibrary.slack.com/archives/C05RB4B5M1N/p1719883395328159), I tried to put together a list of the types of questions we might want to handle below. We don't necessarily need to support all of these, but I figured having a list might help us discuss what we can support and what is maybe complicated enough to add later on if we need it.

| Question Type | Data Type | Example Question | Example Answer |
| ------------- | ------------- | ----- | ----- |
| Multiple Choice, multiple answers | string | What kind of project are you doing? | Community improvement, volunteering, city-sponsored |
| Multiple Choice, single answer | integer | What is your zip code? | 60609 |
| Free-form | string  | Is there anything you need but couldn't find? | A ladder small enough to fit in my little car |
| Free-form | integer | How many attendees do you expect at your event? | 100 |

crismali:
I think `active` and `position` make a lot of sense. I'll add `active` in this round and maybe `position` can be down the road a bit. For this version I was thinking about sorting them by the question's `name` so it's consistent but the `name` probably isn't something people see while making a reservation. 

Those examples are super helpful! I'm not going to jump into adding all of the multiple choice stuff here, but I think it's good to keep in mind. Maybe the options/alternatives could be tossed into a jsonb column on the stem? ¯\_(ツ)_/¯

jim (approved):
This is great. I like how this handles questions mutating over time gracefully.

In terms of alternative names for `result`. The best I can think of is `value`- it's what we use in the Ruby code and `answer.value` reads pretty well. When I see `result` in this app I tend to think of the [Result class](https://github.com/chicago-tool-library/circulate/blob/main/app/lib/result.rb). To me this naming isn't a deal-breaker, though; naming is always hard and a compromise a lot of the time.

For adding eventual multiple choice questions, I imagine we'd store a map of `stem_id` to boolean value in `result`, like `{1: true, 2: false}` to represent which values were selected? Definitely can build that later, I agree the model provides enough flexibility to do that.

PR #1580, opened 2024-07-08
Removed the hold_request_status enum and some no longer used host request files

crismali:
# What it does

Cleans up a postgres enum and some files related to a hold request model that is otherwise removed.

# Why it is important

A little less noise

# Implementation notes

I tried `drop_enum` with arguments in a `change` in the migration but rolling it back didn't work that way. So I just went with up and down.


jim (approved):
Thanks for cleaning this up!

PR #1584, opened 2024-07-15
Disable Turbo links prefetching

### LINKED ISSUE #1581: "Item viewed" ahoy event fires every time you hover over an item
**Describe the bug**
The "Item Viewed" event fires every time you hover over an item, e.g., in search results.

**To Reproduce**
Follow these steps in the dev app (or in the production app if you have access to query the production database).

* Search for something in the search bar.
* Hover over the item description hyperlinks or images for a few items.
* Query the `ahoy_events` table (e.g., `select * from ahoy_events order by time desc limit 20;`) and note that there is one "Item Viewed" event for each hover that you performed.

**Expected behavior**
I suspect this was not the intended functionality, and that the events are supposed to fire each time the user clicks on an item to view its page.

**Screenshots**
Here's an example where I searched for "saw" and then hovered over a bunch of the results. Note the large number of "Item Viewed" events in rapid succession.

image

**Additional context**
Here's where the "Item Viewed" event gets fired in the code.

https://github.com/chicago-tool-library/circulate/blob/0bd00cf74145cb707704e1e2aa1398f943f7e1b7/app/controllers/items_controller.rb#L27-L38

phinze:
# What it does

Closes #1581

# Why it is important

See linked issue for further discussion

# Implementation notes

Docs for this feature are here: https://turbo.hotwired.dev/handbook/drive#prefetching-links-on-hover

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -9,6 +9,7 @@
     <% unless request.get? && !form_request? # prevent turbo for caching form redisplays, which will often contain validation messages %>
       <meta name="turbo-cache-control" content="no-cache">
     <% end %>
+    <meta name="turbo-prefetch" content="false"> <%# ask turbo not to prefetch links on hover %>
     <%= javascript_include_tag "application", "data-turbo-track": "reload", defer: true %>
     <%= stylesheet_link_tag "application", media: "all", "data-turbo-track": "reload" %>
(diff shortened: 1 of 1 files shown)


crismali (approved):
Classic magic meta tags!

jim (approved):
Yeah, I think we can do without this.

PR #1585, opened 2024-07-15
More attempts to fix InvalidAuthenticityToken issues

jim:
# What it does

1. Configures AppSignal to track the `USER_AGENT` header so we know if the issue affect only certain clients. Mobile Safari is a mentioned a lot in the prior art here, but we don't know if our problems also match that pattern.
2. Handle InvalidAuthenticityToken errors by redirecting folks back to where they came from and asking them to try again. In theory this will help if it turns out that folks are hitting this issue by using a cached page that's lost its session, although I think the work done in #1478 should have fixed that.

I couldn't figure out how to succinctly trigger an `InvalidAuthenticityToken` in a controller test, but this code should only affect folks who have already hit an error anyway.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,6 +16,8 @@ class ApplicationController < ActionController::Base
   #    made before the above feature landed: https://github.com/demarches-simplifiees/demarches-simplifiees.fr/pull/6332
   protect_from_forgery store: NonSessionCookieStore.new(:csrf_token), with: :exception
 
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -52,8 +54,6 @@ def current_library
     @current_library ||= Library.find_by(hostname: request.host.downcase) || Library.first
   end
 
-  private
```
jim, line 55: This is redundant; there's another `private` up on line 37.

phinze (approved):
Reviewed live!

crismali (approved):
Looks good 🔥

Issue #1554, opened 2024-06-20
Add a "Missing" status to Item


jim:
This will need to be made to [this file](https://github.com/chicago-tool-library/circulate/blob/main/app/models/concerns/item_statuses.rb#L3) along with a database migration to add the new option to the `item_status` enum.

phinze:
Done in #1587 - thanks @crismali!

phinze:
In testing this on staging I noticed a couple of things we need to round this out.

Currently, missing items are recorded using the "in maintenance" status with a Ticket that says "missing".

![2024-07-17 at 16 47 26@2x](https://github.com/user-attachments/assets/b8883c7a-a5bd-4f85-a6b5-b010ad053cf3)

With #1587 we have the new status, but I think we want to add:

 1. A report to list all missing items
 2. An extra clause in the status display so we're surfacing the "missing" status instead of [the "unavailable" else case](https://github.com/chicago-tool-library/circulate/blob/7d6c1b3c29e45c3fa8888948dc8f57033f63eb12/app/helpers/items_helper.rb#L168-L170)

PR #1587, opened 2024-07-16
Added "missing" to item statuses

crismali:
# What it does

Adds missing as an option to item statuses

# Why it is important

Takes care of #1554

# UI Change Screenshot

Screenshot_2024-07-16_at_12_27_18 AM

# Implementation notes

Just changed the enum and its concern. 🤷‍♂️


phinze (approved):
LGTM! It's a little change but I think it'll be super helpful to staff & volunteers.

Issue #1533, opened 2024-06-09
Show reservation review notes on admin screen


jim:
We need a way to see the reservation review once it has been submitted.

PR #1588, opened 2024-07-16
Added a review notes tab to the reservation show page

crismali:
# What it does

After a reservation is reviewed (approved or rejected) a tab called "Review Notes" appears that lets admins see the review notes.

# Why it is important

Takes care of #1533 

# UI Change Screenshot

Screenshot 2024-07-16 at 1 00 50 AM

# Implementation notes

My text editor auto-formatted the scss. I can undo that so it's only the class I added, but since the code and the HTML have their own linters the changes seem reasonable to me. 

I continued the pattern for the other tabs on the reservation show page. Feels a little heavy to make a full request just for the review notes, but I think keeping the tab implementations consistent is probably more important.

### DIFF
--- a/app/assets/stylesheets/admin.scss
+++ b/app/assets/stylesheets/admin.scss
@@ -4,12 +4,12 @@ body.admin {
   }
 
   table .table-stripe {
-    background: rgba(255,255,255,.7);
+    background: rgba(255, 255, 255, 0.7);
   }
 
   .admin-search-form {
     margin-top: 0.6rem;
-  } 
+  }
 
   .off-canvas-sidebar .nav-item {
     margin-top: 0.2rem;
@@ -25,7 +25,7 @@ body.admin {
     }
   }
 
- .app-header {
+  .app-header {
     background: $primary-color;
     color: white;
     padding-top: 1rem;
@@ -37,7 +37,7 @@ body.admin {
       padding-bottom: 0;
     }
     ul.breadcrumb {
-      margin: -.3rem 0 0 0;
+      margin: -0.3rem 0 0 0;
       padding: 0;
       a {
         color: white;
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/admin.scss
```diff
@@ -71,32 +72,33 @@ body.admin {
       padding-bottom: 0;
     }
   }
+
+  .reservation-review {
+    padding: 1rem 0;
+  }
```
crismali, line 78: This class is the only real change to this file.

app/views/admin/reservations/_profile.html.erb
```diff
@@ -42,13 +42,14 @@
   <div class="column col-sm-12 col-8">
 
     <% if Reservation.statuses.values_at(:requested, :approved, :building).include?(@reservation.status) %>
-      <%= render partial: "reservation_policies_warnings", locals: {reservation: @reservation} %>
+      <%= render partial: "admin/reservations/reservation_policies_warnings", locals: {reservation: @reservation} %>
```
crismali, line 45: I got an error at some point switching between the tabs since this partial didn't have a (more) absolute path.

phinze:
> My text editor auto-formatted the scss. I can undo that so it's only the class I added, but since the code and the HTML have their own linters the changes seem reasonable to me.

Do you know what linter your editor is using? We can add it to our deploy hooks and CI so it runs automatically everywhere, which would match what we're doing for Ruby, JS, and ERB.

phinze (approved):
LGTM!

crismali:
> Do you know what linter your editor is using? We can add it to our deploy hooks and CI so it runs automatically everywhere, which would match what we're doing for Ruby, JS, and ERB.

Mine is set up to use [Prettier](https://prettier.io/). I think of it as a JS ecosystem thing so I'm not sure what the Rails integration looks like.

jim (approved):
:shipit:

jim:
I'm on-board with adding a linter/formatter for the SCSS, too. We can probably spin that out into a separate task.

Issue #1425, opened 2024-03-28
"Sort by" buttons have no effect on search results screen


djcunningham0:
**Describe the bug**
The "sort by" buttons at the top right of the search results screen seemingly have no effect. Toggling between "date added", "number", and "name" does not change the order of the search results.

**To Reproduce**
1. Search for any item. (For example, I searched for "drill" in the screenshots below.)
2. Toggle the "sort by" selection to the other available values, and note that the order of the search results does not change.

**Expected behavior**
If we're going to keep this UI element, the order of the search results should change to reflect the selected "sort by" value.

**Screenshots**
Note that the order of search results is the same in each screenshot, despite the different "sort by" selections.

![image](https://github.com/chicago-tool-library/circulate/assets/38900370/7af6eb8f-dfc8-4cb6-8c5f-07392c9f165d)

![image](https://github.com/chicago-tool-library/circulate/assets/38900370/ddd42408-36ec-4ae3-a549-efb323e47d40)

![image](https://github.com/chicago-tool-library/circulate/assets/38900370/0b5d7dff-e656-40d4-a23e-adce29077f7c)

**Desktop (please complete the following information):**
 - OS: macOS 14.1.2
 - Browser: Safari 17.1.2

**Additional context**
We discussed this bug at the weekly meeting on March 25th, 2024. That discussion is reflected in the options reflected below.

## Options

There are several ways we could fix this bug:

**Option 1: Remove the "sort by" UI element from the search results screen.** 
The included options for sorting don't make much sense from a user perspective, so the easiest fix would be to remove the element entirely. **This would be my preference** because we should always be sorting the results in order of relevance to the search term -- the user doesn't need the option to change the order.

**Option 2: Fix it with the current sorting options**
Make the search results order change when the user selects one of the options ("date added", "name", or "number"). I have not scoped out how difficult the fix would be.

**Option 3: Fix it *and* make improvements.**
This would include Option 2 and extend it by making additional improvements. I can think of two categories of improvements we should make if we decide to keep this UI element:
1. UI changes: The current UI element design doesn't pop and is easy to miss. We could change the colors, placement, and/or type of UI element (e.g., use a dropdown instead of toggle buttons).
2. Sorting options: The current sorting options ("date added", "name", and "number") aren't useful for users. We should default the sort order to something like "Most Relevant", and maybe add additional options like "Availability" that would serve some purpose for the user.

PR #1589, opened 2024-07-16
Fixes the sort buttons on the items index page when there's a search query

crismali:
# What it does

Forces the given sort order on the query and capitalized the sort by buttons. 

# Why it is important

Addresses #1425 

# UI Change Screenshot

Without a query, there is no Relevance sort button:
Screenshot 2024-07-18 at 12 07 27 PM

Relevance:
Screenshot 2024-07-18 at 12 07 38 PM

Date Added:
Screenshot 2024-07-18 at 12 07 45 PM

Number:
Screenshot 2024-07-18 at 12 07 51 PM

Name: 
Screenshot 2024-07-18 at 12 07 59 PM

# Implementation notes

For the sorting, I just changed `order` to `reorder`. The issue was that PGSearch would add a default order by rank that would take priority ahead of any other ordering applied. `reorder` disregards the previous orders on the query and applies our new one. 

For the Relevancy sort I tried always using `reorder` and adding relevance to the hash of sort orders using this [PGSearch utility](https://github.com/Casecommons/pg_search?tab=readme-ov-file#search-rank-and-chained-scopes) but the rank results were different then what PGSearch applies by default. 

I think there's likely a good way to refactor this index action overall, but that's probably left for later. This works for now and it gets some more testing around this behavior.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -101,7 +101,7 @@ class ItemsController < ApplicationController
   end
 
   def apply_order(query)
-    params.delete(:sort) if params[:sort] == "relevancy"
+    params.delete(:sort) if params[:sort] == "relevance"
 
     if params[:sort].blank? && params[:query].present?
       query # we don't need to apply the ordering ourselves since pg_search has already done it
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -114,7 +114,7 @@ class ItemsController < ApplicationController
     options = {
       "name" => "items.name ASC",
       "number" => "items.number ASC",
-      "added" => "items.created_at DESC",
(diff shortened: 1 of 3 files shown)


phinze (changes requested):
I think with this change we lose the ability to ever sort by the search rankings.

Testing it out, here's an example of searching for "drill" and the top hit being "Reamer" (because that was added more recently):

2024-07-17 at 19 04 19@2x

I believe what we need here is an additional scope called "Relevancy" that sorts just by search rank and is selected by default.

crismali:
@phinze 

Took a pass at a relevance sort button, let me know what you think!

crismali:
Thanks, but credit on that change should go to my partner @erinhochstatter who noticed while I was chatting with her about it 😄

Issue #867, opened 2021-11-16
Improve tool embedding metadata


jim:
**Is your feature request related to a problem? Please describe.**

When someone posts a link to a tool on Facebook/Twitter/etc, it would be nice for there to be a nice preview of the item (with a photo if there is one) that is shown.

phinze:
Some example docs talking about this kind of functionality: https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started

Those are Twitter specific but I think there's some Open Graph standard discussed on there that might cover multiple services. Ideally we'd figure out what collection of tags we need so things look decent on the major social media platforms.

phinze:
![2024-07-17 at 17 04 11@2x](https://github.com/user-attachments/assets/05de5d61-741e-41e8-aecb-24c9037511d4)

Looking great! I think we can call this done and open fresh issues for any tweaks we want to make.

Issue #1505, opened 2024-05-20
Switch to Active Storage Variants


jim:
* Move to named variants **with preprocessing on** instead of inline transforms in the views
* Generate all variants in a Rails console session, etc.

Slack discussion: https://chicagotoollibrary.slack.com/archives/C05RB4B5M1N/p1698633739922959

PR #1590, opened 2024-07-16
Added basic open graph meta tags to the item show page

crismali:
# What it does

This will make the share preview look a little friendlier when people share links in apps that support Open Graph protocol (facebook, twitter/x, LinkedIn, Slack, Apple Messages, etc)

# Why it is important

This helps with #867 

# UI Change Screenshot

Most of the tools that let you preview this functionality want a URL to scrape the information from which won't really be available until this is deployed to an environment. So for now, here's the tags in the dom:

Screenshot_2024-07-16_at_4_26_47 PM

# Implementation notes

I actually worked on pretty much the same thing last Friday so I can say that Slack will at least pick up these tags.

I used the [Open Graph protocol site](https://ogp.me/) as a guide and didn't bother adding any platform specific tags for now.

### DIFF
--- a/app/views/items/show.html.erb
+++ b/app/views/items/show.html.erb
@@ -1,3 +1,17 @@
+<% content_for :head do %>
+  <meta property="og:title" content="<%= page_attr(:title) %>">
+  <meta property="og:url" content="<%= request.original_url %>">
+
+  <% if @item.description.present? %>
+    <meta name="description" content="<%= @item.description.to_plain_text %>">
+    <meta property="og:description" content="<%= @item.description.to_plain_text %>">
+  <% end %>
+
+  <% if @item.image.attached? %>
+    <meta property="og:image" content="<%= item_image_url(@item.image, resize_to_limit: [600, 600]) %>">
+  <% end %>
+<% end %>
+
 <div class="items-show">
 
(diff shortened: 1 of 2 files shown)


app/views/items/show.html.erb
```diff
@@ -1,3 +1,17 @@
...
+  <meta property="og:title" content="<%= page_attr(:title) %>">
+  <meta property="og:url" content="<%= request.original_url %>">
+
+  <% if @item.description.present? %>
+    <meta name="description" content="<%= @item.description.to_plain_text %>">
+    <meta property="og:description" content="<%= @item.description.to_plain_text %>">
+  <% end %>
+
+  <% if @item.image.attached? %>
+    <meta property="og:image" content="<%= item_image_url(@item.image, resize_to_limit: [600, 600]) %>">
```
phinze, line 11: Just saying out loud that I checked to make sure this was a `resize_to_limit` pair that we've already used, so we don't have to generate a new variant just for this use case. Once I get around to #1505 these will all switch to named variants that are preprocessed when the item is saved. Ideally we'll end up with a relatively limited set of sizes to save storage space and processing time.

phinze (approved):
Looks great! This will be a nice improvement.

PR #1591, opened 2024-07-17
Switch system tests to use playwright

phinze:
# What it does

Changes from Selenium to Playwright to run our browser based system tests.

# Why it is important

The hope is that this is a smidge more reliable / less flaky.

# Implementation notes

* This is mostly based on a [recent blog post](https://justin.searls.co/posts/running-rails-system-tests-with-playwright-instead-of-selenium/) that reported a good experience switching.
* There were a few places that Playwright had trouble with ambiguous links - it seems to "see" both the mobile and desktop nav menus. Fixed these with `within` scopes.
* The drag/drop test needed an extra delay to work... it seemed like it didn't drag far enough w/o it. I guess playwright is slower to move the virtual mouse!? 😂 
* I decided to yank out Docker support for system tests for now. There is a [story for setting this back up](https://playwright-ruby-client.vercel.app/docs/article/guides/playwright_on_alpine_linux), but it's high effort and nobody is using it right now. We can always pick it up down the road.
* Cleaned up unused / unnecessary helpers:
   - Richtext selector: included in Capybara / Rails now
   - JS Error detection: no tests fail without it, so seems fine now

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -32,7 +32,10 @@ Capybara.register_driver :headless_chrome do |app|
   Capybara::Playwright::Driver.new(
     app,
     browser: :chrome,
-    headless: true
+    headless: true,
+    # We only use this env var for the Nix dev environment. For other platforms
+    # we can let playwright use its own downloaded browser.
+    executablePath: ENV.fetch("PLAYWRIGHT_CHROME_BINARY", nil)
   )
 end
 
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -20,49 +20,40 @@ Capybara.add_selector :rich_text_area do
   end
 end
 
-Capybara.register_driver :headless_chrome do |app|
-  Capybara::Selenium::Driver.new(
+Capybara.register_driver :headed_chrome do |app|
+  Capybara::Playwright::Driver.new(
     app,
     browser: :chrome,
-    options: Selenium::WebDriver::Chrome::Options.new.tap do |opts|
-      opts.add_argument "--headless=new"
-      opts.add_argument "--no-sandbox"
-      opts.add_argument "--disable-gpu"
-      opts.add_argument "--disable-dev-shm-usage"
-      opts.add_argument "--window-size=1400x1800"
-
-      # This configurable is used in the Nix development set up, see NIX.md.
-      # It's safe to remove if the Nix pieces are ever ripped out.
-      if ENV.has_key?("SELENIUM_CHROME_BINARY")
-        opts.binary = ENV["SELENIUM_CHROME_BINARY"]
-      end
-    end
+    headless: false
   )
 end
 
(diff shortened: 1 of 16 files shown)


test/application_system_test_case.rb
```diff
@@ -20,49 +21,43 @@
...
+# Capybara.register_driver :chrome_in_container do |app|
+#   Capybara::Selenium::Driver.new(
+#     app,
+#     browser: :remote,
+#     url: "http://selenium_chrome:4444/wd/hub",
+#     options: Selenium::WebDriver::Chrome::Options.new.tap do |opts|
+#       opts.add_argument "--window-size=1400x1800"
+#     end
+#   )
+# end
```
crismali, line 60: Do we want to keep this as comments in case there are some bumps down the road with this transition? I lean towards just deleting it and pulling it from git if it becomes important later on.

test/test_helpers/stdout_helpers.rb
```diff
@@ -0,0 +1,34 @@
+# Taken from https://gist.github.com/searls/9caa12f66c45a72e379e7bfe4c48405b
+# Referenced in https://justin.searls.co/posts/running-rails-system-tests-with-playwright-instead-of-selenium/
```
crismali, line 2: I haven't used it before, but maybe something like [this gem](https://github.com/Shopify/minitest-silence) might be good for this?

For some reason I thought Rails just baked in a helper to help with this but I couldn't dig up. I did find a [different approach to this here](https://gist.github.com/moertel/11091573). I don't have strong feelings about this sort of thing because it's always weird, but this other approach does avoid monkey-patching `Kernel` 🤷

crismali (approved):
Looks like there's a fair bit of commented config floating around, but nothing major. I was able to pull this down,  run `bin/setup`, and run the tests without issue, so real solid 👍 

I also left a little note on suppressing `puts` in tests. I'm certain there are a thousand more implementations and all of them are fine, but just wanted to present an alternative in case there's ever an issue.

test/application_system_test_case.rb
phinze: Agreed! This was leftover from Jim's initial experimentation I think. Will clean up.

test/test_helpers/stdout_helpers.rb
phinze: Took a peek at both of those options and I think I prefer this one because it's more conservative in two ways:

 1. It only redirects output when you explicitly call it instead of globally
 2. It only suppresses the regexp you tell it to

The aesthetic debate around whether global variable juggling or Kernel monkey patching is "less gross" might be a fun diversion, but I too don't feel that strongly and am happy to conclude by agreeing "It's always weird." 🤝 😄

test/application_system_test_case.rb
phinze: Done in c9b6e3d547aa95e45348251ce1eca594bee7b2f5

Issue #1439, opened 2024-04-03
Allow limiting to available items on admin search results page


phinze:
**Is your feature request related to a problem? Please describe.**
The toggle to limit results to available items only exists on member search pages and the admin items list page. It would be helpful to be able to have the same toggle when you search for a tool in the admin page.

**Describe the solution you'd like**

![2024-04-03 at 12 55 40@2x](https://github.com/chicago-tool-library/circulate/assets/37534/20a306c8-6e9f-40db-b38f-d849e155c831)

PR #1592, opened 2024-07-18
Added toggle to limit admin search results to available items only

crismali:
# What it does

Added toggle to limit admin search results to available items only.

# Why it is important

Takes care of #1439

# UI Change Screenshot

With the toggle off:
Screenshot 2024-07-18 at 4 10 01 PM

With the toggle on:
Screenshot 2024-07-18 at 4 10 12 PM

# Implementation notes

I reused the partial from the items#index and admin/items#index even though it has some hidden field tags we don't really need in this context.

I also removed searches#create since it just passed through to searches#show anyway. Just had to change the forms to be `get` requests.

### DIFF
--- a/app/controllers/admin/searches_controller.rb
+++ b/app/controllers/admin/searches_controller.rb
@@ -3,12 +3,6 @@ module Admin
     def create
       query = params[:query]
 
-      # item = Item.where(number: query).first
-      # redirect_to(admin_item_path(item)) && return if item
-
-      # member = Member.matching(query).first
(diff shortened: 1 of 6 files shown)


phinze (approved):
Looks great!

> I also removed searches#create since it just passed through to searches#show anyway. Just had to change the forms to be get requests.

This will likely have the side effect of sidestepping one of our common sources of InvalidAuthenticityToken errors! /cc @jim for visibility

PR #1596, opened 2024-07-20
GoodJob 4.0 migrations

phinze:
# What it does

Adds missing GoodJob migrations as I didn't notice that Dependabot took us over the major boundary in #1582.

# Why it is important

Fixes job errors we're seeing in production that look like https://github.com/bensheldon/good_job/issues/1404

# Implementation notes

Generated by pinning good_job to 3.99 and running

```
bin/rails good_job:update
bin/rails db:migrate
```

Follow up tasks to consider: 

 * Smoke testing GoodJob in tests (I think we'll get this as a matter of course once we start converting some of our cron tasks over)
 * Splitting out major version upgrades to separate dependabot group

### DIFF
--- a/db/migrate/20240720134319_create_good_job_settings.rb
+++ b/db/migrate/20240720134319_create_good_job_settings.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+class CreateGoodJobSettings < ActiveRecord::Migration[7.1]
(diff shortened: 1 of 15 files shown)


phinze:
Going to rush this through sans review because jobs are exploding in prod, but requesting review like normal for visibility.

PR #1601, opened 2024-07-20
log additional information for search and view events

djcunningham0:
# What it does

Add more useful information to the "Search" and "Item Viewed" events logged by ahoy.

Specifically, this PR adds the following information:
* Search
  + "params": any parameters included with the search, such as "sort" and "category" (this will also include "query", which we're already logging, but it's probably not too big a deal to include it here too)
  + "n_results": number of results returned by the query
  + "first_page_results": a list of the ordered item IDs on the first page results
    + I'm not sure whether this will be useful/necessary. **See implementation note below.**
* Item Viewed
  + "referrer_url": add the referrer URL so we can connect Item Viewed events to search queries, or see which other places are leading to item views

# Why it is important

The additional information will help us understand how users are interacting with the site, specifically related to search.

# UI Change Screenshot

No UI changes, but here's what the `ahoy_events` table looks like after performing a few actions in the dev app:

image

# Implementation notes

### first_page_results

First I tried to figure out a way to add an "item_index" field to the Item Viewed event that would tell us whether the user clicked on the 1st, 2nd, ... item in the results. But I couldn't figure out how to do that. Instead, I decided to add the "first_page_results" field to the Search event -- we can theoretically connect the Item Viewed event to the Search event, and then look up the item ID in "first_page_results" to figure out the item index.

Another benefit is that we get (quasi-)impression data for search results. We can assume that the user saw at least the first several results from the list. This could possibly be useful in some search analysis (for example, if we find that item *X* is always the first result for query *Q* but it never gets clicked on, that indicates that item *X* is not relevant to query *Q*).

The downside is we're logging a lot more data (relative to what we were previously logging), and this is probably the least important field.

**Questions:**
1. Is the volume of data (a list of 20 item IDs per search event) too large to justify keeping this field? I'm not sure what kind of table size constraints we're working with.
2. Does anyone know how we might add an "item_index" field to the Item Viewed event instead? That accomplish the same thing (minus the impression data) with a much smaller footprint and would make analysis easier.

### referrer_url

I noticed this field doesn't behave the same way in all browsers. Specifically, if I'm using Safari and I open the link in a new tab, the referrer URL is null (but if I open the link in the same tab the referrer URL is populated). In Chrome, the referrer URL was always populated, whether I opened the link in the same tab or a new tab. 

Let me know if anyone knows a way to ensure the referrer URL is always included in Safari -- but that would be super low priority.

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -20,7 +20,7 @@ class ItemsController < ApplicationController
 
     # Track that a search was performed if we are on the first page
     if @query && @pagy.page == 1
-      ahoy.track "Searched", query: @query, params: (@filter_params || {}), n_results: @pagy.count, first_page_results: @items.map(&:id)
+      ahoy.track "Searched", query: @query, params: @filter_params || {}, n_results: @pagy.count, first_page_results: @items.map(&:id)
     end
   end
 
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -20,7 +20,7 @@ class ItemsController < ApplicationController
 
     # Track that a search was performed if we are on the first page
     if @query && @pagy.page == 1
-      ahoy.track "Searched", query: @query, params: @filter_params || {}, n_results: @pagy.count, first_page_results: @items.map(&:id)
+      ahoy.track "Searched", query: @query, params: @filter_params || {}, n_results: @pagy.count
     end
   end
 
@@ -37,7 +37,7 @@ class ItemsController < ApplicationController
     # the `request.referer != request.url` condition prevents logging the view after
     # placing a hold redirects back to the item page
     if request.referer != request.url
-      ahoy.track "Item viewed", item_id: @item.id, referrer_url: request.referer
+      ahoy.track "Item viewed", item_id: @item.id, referrer_url: request.referer, search_result_index: params[:search_result_index].to_i + 1
     end
   end
 
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -20,7 +20,7 @@ class ItemsController < ApplicationController
 
     # Track that a search was performed if we are on the first page
     if @query && @pagy.page == 1
-      ahoy.track "Searched", query: @query
+      ahoy.track "Searched", query: @query, params: @filter_params, n_results: @pagy.count, first_page_results: @items.map(&:id)
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/items_controller.rb
```diff
@@ -34,7 +34,11 @@ def show
       @current_hold_count = current_member.active_holds.active_hold_count_for_item(@item).to_i
     end
 
-    ahoy.track "Item viewed", item_id: @item.id
+    # the `request.referer != request.url` condition prevents logging the view after
+    # placing a hold redirects back to the item page
+    if request.referer != request.url
```
jim, line 39: We could write a wrapper of some kind (it would be a great use for Redis, but we don't have it in the stack) that could limit the tracking dynamically according to a rule like "only track page views once every n minutes". But I think this is a good place to start.

jim (approved):

jim:
Sorry, I missed answering your questions in my review!

> Is the volume of data (a list of 20 item IDs per search event) too large to justify keeping this field? I'm not sure what kind of table size constraints we're working with.

It is storing more data, but I don't think 20 integers is a big deal right now. We might need to think about deleting old events/visits at some point anyway. How many `events` are we storing in an average month?

> Does anyone know how we might add an "item_index" field to the Item Viewed event instead? That accomplish the same thing (minus the impression data) with a much smaller footprint and would make analysis easier.

I think we can probably add a query parameter to the links to the item pages on the index:

```erb
<% @items.each.with_index do |item, index| %>
  <%= link_to item.name, item_path(item, search_result_index: index) %>
<% end %>
```

And then in the `ItemsController#show` action, `params[:search_result_index].to_i` should be the index. We might want some validation on the field to ensure it's an integer between 0 and whatever he max on a page is.

I think I agree with you that the second option is the better way to go, and it means not needing to store the list of search results.

djcunningham0:
I implemented the change as Jim described in his comment and as we discussed on the call. I removed the "first_page_results" field from the search event and added the "search_result_index" field to the item viewed event.

Screenshot demonstrating the new logs (running locally on the dev database):

image

phinze:
Changes look good, but [this test failure](https://github.com/chicago-tool-library/circulate/actions/runs/10154096146/job/28078647133?pr=1601) looks real:

```
Failure:
ItemsControllerTest#test_only_shows_one_of_a_given_item_regardless_of_how_many_categories_it_is_in [test/controllers/items_controller_test.rb:40]:
Expected exactly 1 element matching "a[href='/items/309']", found 0.
Expected: 1
  Actual: 0
```

I think the assertion here needs to be updated with the same index parameter we added:

https://github.com/chicago-tool-library/circulate/blob/da11149a38787137d484a68a9903184c85535b9c/test/controllers/items_controller_test.rb#L40

phinze (approved):
Looking good! Lets see what the additional info can tell us

Issue #1532, opened 2024-06-09
Send an email when a reservation is approved or rejected


jim:
* This should be sent to the person who submitted the reservation.
* It should include any notes entered by a staff member on the approval form.

PR #1603, opened 2024-07-22
Send an email when a reservation is approved or rejected

crismali:
# What it does

This adds an approval/rejection email that's sent to whomever originally submitted the reservation.

# Why it is important

Takes care of #1532 

# UI Change Screenshot

HTML approval:
Screenshot 2024-07-22 at 4 45 31 PM

Plain text approval:
Screenshot_2024-07-22_at_4_45_41 PM

HTML rejection:
Screenshot 2024-07-22 at 4 47 04 PM

Plain text rejection:
Screenshot 2024-07-22 at 4 47 12 PM

# Implementation notes

The plain text emails have the old address because that's what's in the setup data, while the HTML mailer layout has the new address hardcoded.

I had to add `submitted_by` to the reservation to have someone to send the email to. Right now this is just set to `current_user` in the `create` action, but this will likely have to happen somewhere else. I also made the association required. I could see it being optional but I'm not sure how `submitted_by` would be added later given the current flow. Easy enough to drop though!


crismali:
Just realized that this might want to wait for #1559 to land. `submitted_by` might become the organization or the user and the organization (assuming the organization ends up with some contact information attached to it).

test/mailers/reservation_mailer_test.rb
```diff
@@ -0,0 +1,65 @@
...
+  test "#reservation_requested includes HTML and text parts when a reservation is submitted" do
+    reservation = create(:reservation)
+
+    ReservationMailer.with(reservation:).reservation_requested.deliver_now
+
+    mail = ActionMailer::Base.deliveries.last
+
+    subject = "Reservation submitted for #{reservation.started_at.to_fs(:short_date)}"
+
+    assert mail.multipart?, "mail should be multipart"
```
jim, line 17: There's an [`assert_delivered_mail`](https://github.com/chicago-tool-library/circulate/blob/c94929772bd3ad971f2ce55edca6db09dfc2a88b/test/application_system_test_case.rb#L62)  utility function ([used like this](https://github.com/chicago-tool-library/circulate/blob/c94929772bd3ad971f2ce55edca6db09dfc2a88b/test/system/admin/gift_memberships_test.rb#L37)) currently used in system tests that might make this more concise.

jim (approved):
Looks good to me!

crismali:
Thanks! Definitely looks better with `assert_delivered_email` ✨ 

Is this the sort of thing where we want to merge now and make changes after other PRs land or hold off on merging?

jim:
@crismali We've been merging this stuff in since the group lending isn't actively used yet and having everything in `main` tends to unblock other work.

PR #1604, opened 2024-07-22
Reorder sass rules to prevent warnings

crismali:
# What it does

This reorders some style rules to prevent deprecation warnings from SASS.

# Why it is important

Reduces noise and surfaces any issues with SASS now instead of when it upgrades.

An example warning:
```
Deprecation Warning: Sass's behavior for declarations that appear after nested
rules will be changing to match the behavior specified by CSS in an upcoming
version. To keep the existing behavior, move the declaration above the nested
rule. To opt into the new behavior, wrap the declaration in `& {}`.

More info: https://sass-lang.com/d/mixed-decls

   ┌──> app/assets/stylesheets/globals.scss
73 │     letter-spacing: $button-letter-spacing;
   │     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ declaration
   ╵
   ┌──> app/assets/stylesheets/mixins.scss
64 │ ┌     a {
65 │ │         color: white;
66 │ │         text-decoration: none;
67 │ │     }
   │ └─── nested rule
   ╵
    app/assets/stylesheets/globals.scss 73:3          @import
    app/assets/stylesheets/styles.scss 1:9            @import
    app/assets/stylesheets/application.sass.scss 1:9  root stylesheet
```

# Implementation notes

I just rearranged the lines so that the behavior remains the same post future sass upgrade as it does now. It's possible this changes some attempts at overriding mixins, but didn't notice anything different when I was clicking around. I'll double check, but it'd be great if someone else took a look. 👀 

Also, this PR makes much more sense in the "unified" Github view, instead of "split."

### DIFF
--- a/app/assets/stylesheets/globals.scss
+++ b/app/assets/stylesheets/globals.scss
@@ -69,9 +69,9 @@ $pri-btn-color-hover: darken($button-color, 7%);
 
 // BUTTONS
 .primary-btn {
-  @include primary-button;
   letter-spacing: $button-letter-spacing;
   border: none;
+  @include primary-button;
   a {
     height: 50px !important;
   }
@@ -82,8 +82,8 @@ $pri-btn-color-hover: darken($button-color, 7%);
 }
 
 .secondary-btn {
-  @include secondary-button;
   letter-spacing: $button-letter-spacing;
+  @include secondary-button;
 }
 
 .link-btn {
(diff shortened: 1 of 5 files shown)


phinze (approved):
Took this for a quick smoke test and LGTM!

Issue #1593, opened 2024-07-19
Members with Overdue Items Report


phinze:
Via @tessavierk -

> Could we add a report for members with overdue items, the items they have, and their contact info?
> 
> Sometimes if we have down time we call or email members with pretty overdue items and it would be MUCH easier if we had a spreadsheet to work from instead of clicking around in the overdue loans section of Circulate.

Table I'm imagining:

| Member | Overdue Tools | Email | Phone |
| --- | --- | --- | --- |
| Penelope | Saw (5 days)<br>Drill (2 weeks) | penelope@example.com | 773-555-1234 |
| Jasper | Ladder (3 months) | jasper@example.com | 773-555-1234 |

phinze:
We should include a CSV download button on this report:

> **sometimes we will do a bulk email and it is nice to just copy-paste all the overdue folks**, but sometimes we do personal outreach and it's helpful to see what we're calling/emailing them about and who to prioritize

PR #1605, opened 2024-07-24
Added a members with overdue loans report

crismali:
# What it does

Adds a report that shows each member with overdue loans.

# Why it is important

Takes care of #1593 

# UI Change Screenshot

Report page:
Screenshot 2024-07-24 at 10 32 49 AM

CSV:
Screenshot 2024-07-24 at 10 33 07 AM

# Implementation notes

More or less copied the pattern from the memberships report.


jim (approved):
This is great!

PR #1606, opened 2024-07-26
Fix undefined method error in CSRF handler

jim:
# What it does

There is still a bug in the custom CSRF handling code:

```
undefined method `Appsignal' for #<Appsignal::Transaction:0x00007f8f23094ae8 ...>

      transaction.Appsignal.set_custom_data(
                 ^^^^^^^^^^
```

I added a test for `ApplicationController#handle_invalid_authenticity_token`. It's worth noting that the block passed to `Appsignal.report_error` in that method isn't executed in the test environment, but I manually hacked things up and validated that at least the exception shown above (nor one like it) is no longer happening.


jim:
The system tests are failing on all PRs right now. I'm going to merge this to prevent this error from happening in production.

Issue #1598, opened 2024-07-20
Speed up bin/setup


phinze:
When I already have all dependencies installed, `bin/setup` takes 1m39s on my machine. Ideally, `bin/setup` should be something that feels easy to run without thinking of it. I think sub 30s w/ all deps installed would be a good goal, but even getting it down to sub-1m would be a win.

The majority of time looks to be spent in `devdata.rake` and I bet there's some low hanging fruit in there.

PR #1614, opened 2024-07-30
devdata: Reuse image blobs to save space and maybe a little time

phinze:
# What it does

Prevents us from copying the fixture image once per ~400 items loaded during devdata

# Why it is important

Part of work on #1598 

# Implementation notes

* I was also able to rework the AnalyzeJob workaround in a way that should prevent AnalyzeJobs from flying around in the dev env.
* As mentioned in the comments, this means the fixture image will get rotated everywhere if it's rotated. I think this is a reasonable tradeoff for how much repeated churn this saves.

### DIFF
--- a/lib/tasks/devdata.rake
+++ b/lib/tasks/devdata.rake
@@ -51,7 +51,17 @@ namespace :devdata do
       offset = (index + 1) * 10000
       admin = library.users.where(role: "admin").first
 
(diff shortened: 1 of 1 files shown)


jim (approved):
image

I think I agree that the speed improvement here is worth the slightly weird situation with images.

PR #1615, opened 2024-07-30
Move development and CI postgres to tmpfs ramdisk

phinze:
# What it does

Moves postgres to run on a tmpfs ramdisk. This yields a modest improvements in bin/setup and test time locally, and hopefully similar speedups in CI.

# Why it is important

Faster tests means less human time waiting!

# Implementation notes

This means that the development database contents will not persist across reboots for people using Docker. The main instructions have folks installing postgres with homebrew, so it won't affect most people.


jim (approved):
image

PR #1616, opened 2024-07-31
Handle members with no associated user.

jim:
# What it does

Fixes the Members with Overdue Loans Report, which is currently [broken in production](https://chicagotoollibrary.slack.com/archives/C05U0JSJ6AD/p1722191013493979).

# Why

It turns out that we have 48 `Member` records without associated `User` objects:

```sql
SELECT COUNT(members.id)
FROM members
LEFT JOIN users on users.member_id = members.id
WHERE users.id IS NULL;
=> 48
```

This data situation is breaking the Members with Overdue Items report, so this PR makes the template more forgiving. Since there is such a small number of members with this issue, we can go in and fix the database at some point. But I'd like to fix the report ASAP so that staff can start using it.

### DIFF
--- a/app/views/admin/reports/members_with_overdue_loans/index.html.erb
+++ b/app/views/admin/reports/members_with_overdue_loans/index.html.erb
@@ -26,7 +26,7 @@
             <% end %>
           </td>
           <td>
-            <%= member.user.email %>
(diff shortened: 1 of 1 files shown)


crismali (approved):
Looks good!

Issue #1349, opened 2024-02-20
Consolidate notifications to cut down on the max notifications a user gets in a day


phinze:
## Is your feature request related to a problem? Please describe.**

In investigating [a question on a PR](https://github.com/chicago-tool-library/circulate/pull/1337#pullrequestreview-1886486850), I discovered that we're regularly sending quite a few notifications to some users.

![2024-02-19 at 18 23 15@2x](https://github.com/chicago-tool-library/circulate/assets/37534/392d59c2-d411-4cd2-8bce-542cc3b74352)

## Describe the solution you'd like

We should analyze the patterns of notifications and figure out if we can consolidate some of them.

## Additional context

This applies to both email and SMS notifications. It costs us money per text message to send SMS so any reduced notifications will also save us money running the app.

PR #1617, opened 2024-08-01
Consolidate hold available notifications

crismali:
# What it does

Instead of members receiving an email and text for every hold that has become available, they'll receive one email and one text describing all of the holds that have become available for them.

# Why it is important

Helps with #1349 

# UI Change Screenshot

Single hold html email:
Screenshot 2024-08-01 at 2 16 46 PM

Single hold plain text email:
Screenshot 2024-08-01 at 2 17 12 PM

Multiple holds html email top part:
Screenshot 2024-08-01 at 2 23 44 PM

Multiple holds html email bottom part:
Screenshot 2024-08-01 at 2 23 52 PM

Multiple holds plain text email:
Screenshot 2024-08-01 at 2 23 06 PM

# Implementation notes

I adjusted the existing mailer and texter to accept multiple holds with a little bit of logic around pluralizing the copy. 

I also added a smoke test for `Hold.start_waiting_holds` method and refactored it a bit so it shouldn't have to pull quite as many holds out of the database. Looks like between 500 and 600 holds will be pulled into memory now (down from 800-900).


phinze:
Directionally this looks really good! However, I think it's missing a piece.

AFAICT even with these changes we're still running `holds:start_waiting_holds` every 10 minutes, so unless somebody gets assigned multiple holds in the same 10m span, they'll still mostly be getting individual messages.

That being said, I think this is still probably worth merging as-is, since it's a necessary step towards further consolidation.

@jim I'm curious to hear your thoughts on this too, so I'm gonna hold off on approving for the time being.

(Also notifying members of holds is one of the most important flows in Circulate, so having multiple eyeballs on reviews in this area is a good idea anyways!)

app/models/hold.rb
```diff
@@ -116,22 +116,7 @@ def self.extend_started_holds_until(date)
...
-      unless hold.item.holdable?
-        Rails.logger.debug "[hold #{hold.id}]: item is not holdable"
-        next
-      end
-
-      if hold.started?
-        Rails.logger.debug "[hold #{hold.id}]: already started"
-        next
-      end
-
```
jim, line 134: All of these debug lines were useful back in the beginning, but I think cleaning them up at this point is nice. Probably better test coverage is the better way to ensure that things are working as intended (thanks for adding another test).

I think pulling these checks up into the query makes sense.

jim (approved):
The logic looks good here. I agree with @phinze that it's worth thinking through how to better batch notifications, and also that this is good to merge as it's a big improvement.

One nice thing is that since [all holds expire at the end of the day](https://app.chicagotoollibrary.org/blazer/queries/9-when-active-holds-expire), the notifications for any holds that "age out" should get collapsed by this new approach. Holds that are started when items are returned at other times, though, will still get sent individually.

I think we should probably have a design chat about batching to figure out the right balance between fewer notifications and more immediate delivery. This might get us far enough that we don't need to build a real notification batching system yet, although I am not opposed adding one if it turns out we do need it.

phinze (approved):
Nice thanks @jim that's the insight I was missing and makes me feel a lot more confident about this PR as-is. So holds becoming available via another hold's expiration will tend to be batched and holds becoming available via returns will be more spread out. Let's merge this and see what we learn!

PR #1620, opened 2024-08-04
Turn on standard-rails and fix timezone-unsafe method calls

phinze:
# What it does

It turns out you need to activate standard-rails in a config, so we've had it installed for a while but haven't been using it! I noticed this as I was chasing down time-related test flakiness. So here we get standard-rails to snap everything into consistency and flush out all outstanding issues relating to non-timezone-honoring method calls.

# Why it is important

This should fix outstanding time-related test flakes and help protect us against future footguns.

# Implementation notes

* Steps I took
  - I activated standard-rails and ran `standardrb --fix` to catch all the automatically fixable errors.
  - Then I ran `standardrb --generate-todo` to capture the remaining issues that need manual fixing.
  - Then I removed all the timezone related issues from the ignores in the todo file and manually fixed them.
* To prevent the already-huge diff from getting bigger, I'll spin up an issue to fix the remainder of the issues in the todo list.
* I recommend reviewing each commit individually and turning on whitespace ignore.


app/models/hold.rb
```diff
@@ -110,40 +110,40 @@ def self.next_waiting_hold_dates
   end
 
   def self.extend_started_holds_until(date)
-    Hold.active.started.where("expires_at < ?", date).update_all(expires_at: date)
+    Hold.active.started.where(expires_at: ...date).update_all(expires_at: date)
```
jim, line 113: TIL about this syntax for time comparisons in queries. Neat.

jim (approved):
Lots of little improvements in here, and it's cool the way it generates a TODO file with issues to work through.

PR #1624, opened 2024-08-05
Pin standard-rails to main to cut down TODO

phinze:
# What it does

standard-rails has recently disabled several rules we have in our todo list, but they have yet to cut a release, so temporarily pinning to their latest commit

# Why it is important

Helps us prevent some busywork complying with rules that will be turned off soon.


phinze:
Nothing fancy here; going to self-merge so #1572 can benefit from it.

Issue #1370, opened 2024-02-24
Require email confirmation before borrowing items


jim:
We have an issue where we end up with invalid email addresses for members.

When folks sign up online, we show them a confirmation page, so we'll use that as the starting point to begin confirming member email addresses.

* Add a banner to this page that says to check their email for a confirmation link (and to click that link).
* Send the confirmation link via email when creating the user (Devise should be able to do this automatically).
* We'll need to show a notice at the top of every page of the member account portal that reminds folks to confirm their email address (and optionally re-send the email if needed).

mdecelle:
The feature to implement is part of https://github.com/heartcombo/devise - they call it "confirmable"

PR #1626, opened 2024-08-07
Require email confirmation before "borrowing" items

crismali:
# What it does

This adds a check to prevent users from placing holds on items before confirming their email address (via devise). It also adds some warning toast messages to encourage users to confirm.

# Why it is important

Handles #1370 

# UI Change Screenshot

Verify email message on the search page:
Screenshot 2024-08-07 at 3 21 06 PM

Verify email message on the item page:
Screenshot 2024-08-07 at 3 20 58 PM

Message where a user would normally see the "Place Hold" button:
Screenshot 2024-08-07 at 3 21 51 PM

Member login page:
Screenshot_2024-08-07_at_3_22_07 PM

Resend confirmation instructions page:
Screenshot 2024-08-07 at 3 22 18 PM

# Implementation notes

It doesn't look like members can create loans directly, so I prevented members with unconfirmed email addresses from being able to make holds. Admins could still create loans for these members, so maybe we want to add a message on the admin side that warns them about creating loans for these members?

Right now the migration will mark all existing users as verified so these changes should only impact new members. I'm not sure if there's a heuristic we could use to make this smarter.

Take a look and let me know what you think!

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -1,16 +1,16 @@
-<h2>Resend confirmation instructions</h2>
+<div class="users-confirmations-new container">
+  <h2>Resend Confirmation Instructions</h2>
 
-<%= form_for(resource, as: resource_name, url: confirmation_path(resource_name), html: {method: :post}) do |f| %>
-  <%= render "devise/shared/error_messages", resource: resource %>
+  <%= form_for(resource, as: resource_name, url: confirmation_path(resource_name), html: {method: :post}) do |f| %>
+    <%= render "devise/shared/error_messages", resource: resource %>
 
-  <div class="field">
-    <%= f.label :email %><br>
-    <%= f.email_field :email, autofocus: true, autocomplete: "email", value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email) %>
-  </div>
+    <div class="field text mb-2">
+      <%= f.label :email %><br>
+      <%= f.email_field :email, autofocus: true, autocomplete: "email", value: (resource.pending_reconfirmation? ? resource.unconfirmed_email : resource.email) %>
+    </div>
 
-  <div class="actions">
-    <%= f.submit "Resend confirmation instructions" %>
-  </div>
-<% end %>
-
-<%= render "devise/shared/links" %>
+    <div class="actions text-center">
+      <%= f.submit "Resend Confirmation Instructions", class: "btn btn-primary" %>
+    </div>
+  <% end %>
+</div>
--- a/app/views/devise/sessions/new.html.erb
+++ b/app/views/devise/sessions/new.html.erb
@@ -37,6 +37,7 @@
         <li>If you don't have a membership with us yet, <%= link_to "sign up online", signup_path %>.</li>
       <% end %>
         <li>If you need to choose or don't remember your password, <%= link_to "reset your password", new_user_password_path %>.</li>
+        <li>If you need haven't confirmed your email address, <%= link_to "request a confirmation email", new_user_confirmation_path %>.</li>
       </ul>
   </div>
 </div>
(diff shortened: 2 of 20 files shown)


app/views/devise/confirmations/new.html.erb
```diff
@@ -1,16 +1,16 @@
...
-    <%= f.submit "Resend confirmation instructions" %>
-  </div>
-<% end %>
-
-<%= render "devise/shared/links" %>
+    <div class="actions text-center">
+      <%= f.submit "Resend Confirmation Instructions", class: "btn btn-primary" %>
+    </div>
+  <% end %>
+</div>
```
crismali, line 16: Changed this template to look more like the other devise pages.

config/initializers/devise.rb
```diff
@@ -134,7 +134,7 @@
   # without confirming their account.
   # Default is 0.days, meaning the user cannot access the website without
   # confirming their account.
-  # config.allow_unconfirmed_access_for = 2.days
+  config.allow_unconfirmed_access_for = 2.days
```
crismali, line 137: 2 days is the default here, but maybe it should be more like a week?

phinze: Since we're explicitly disallowing the thing we don't want people to do without confirming, I wonder if we should set it to [`nil` so it never kicks them out?](https://github.com/heartcombo/devise/pull/2278)

phinze:
This is looking really good!

> Admins could still create loans for these members, so maybe we want to add a message on the admin side that warns them about creating loans for these members?

I think something on the admin side member page to indicate that the user has not confirmed their email address is a good idea.

app/views/devise/confirmations/new.html.erb
jim: Good catch!

config/initializers/devise.rb
jim: I agree that if we're going to go with the model of preventing actions within the account based on if an email is confirmed, it's OK to always let folks log in.

phinze:
I'm flushing out a few Circulate tasks now, so I can pick up the two outstanding tweaks and bring this home. 🤝

phinze:
After discussing this with staff, we have two changes to make here:

 * Prevent users from logging in when verification is incomplete
 * Change message copy to use active voice e.g. "You have not verified your email address yet! Please click the confirmation link in the email you received from us. Click here to request a new email."

And one change to investigate:

 * Can we consolidate the confirmation link in with the new member email?

crismali:
I've made some of the changes we talked about the other night. A new question is how much do we want to surface folks who have changed their email addresses but haven't confirmed the new ones yet. I've changed the members so that when they change the member email they're also updated the user unconfirmed email so there shouldn't be in any place where a user's email being changed directly.  I've also made it so the message on the admin member page appears if the user has never been confirmed or if they just haven't confirmed their new email. 

What do folks think? Should we call this out more for members or admins?

app/views/devise/sessions/new.html.erb
```diff
@@ -37,6 +37,7 @@
         <li>If you don't have a membership with us yet, <%= link_to "sign up online", signup_path %>.</li>
       <% end %>
         <li>If you need to choose or don't remember your password, <%= link_to "reset your password", new_user_password_path %>.</li>
+        <li>If you need haven't confirmed your email address, <%= link_to "request a confirmation email", new_user_confirmation_path %>.</li>
```
phinze, line 40:
```suggestion
        <li>If you haven't confirmed your email address, <%= link_to "request a confirmation email", new_user_confirmation_path %>.</li>
```

Found a typo, though I think it might be more confusing to have this showing to every logging in user vs only showing it in flash messages when the email is not verified.

phinze: It looks like we can modify the devise settings so the flash message includes a link instead. https://stackoverflow.com/questions/4008059/rails-devise-customize-flash-message-devise-en-yml-with-a-link-to

jim (approved):
I think this is looking good. I just want to think through if we need to do any communication before merging and deploying, since every existing user will suddenly not be able to log in.

crismali:
@jim They should all be able to log in with this. The migration makes it so that if you're already an existing user you're automatically confirmed even though that hasn't actually happened.

Issue #1311, opened 2024-02-07
As a user, I would like a Renew All Loaned Items Button


jpeace-pro:
**Is your feature request related to a problem? Please describe.**
When I borrow many items at a time and need to renew all of them, I have to press Renew Loan for each individual item. This can be tedious.

**Describe the solution you'd like**
I'd like to renew all items I have on loan in batch with one "Renew All" button.

**Describe alternatives you've considered**
Borrow fewer items. 

**Additional context**
Maybe items get grouped by return date and you can renew them in groups per date of extension period. I.e. I have 3 items due 12/4 and 4 items due 12/7. I can hit renew all for all items on 12/4 and renew all for all items on 12/7 separately.

phinze:
Looks like Admins can do this so we'll duplicate the behavior for members: https://github.com/chicago-tool-library/circulate/blob/da11149a38787137d484a68a9903184c85535b9c/app/views/admin/members/show.html.erb#L50-L52

PR #1627, opened 2024-08-08
Added a renew all button for loans

crismali:
# What it does

Adds a button and controller action to renew all loans.

# Why it is important

Takes care of #1311 

# UI Change Screenshot

The loans page with the button:
Screenshot 2024-08-08 at 3 45 23 PM

The loans page with the button and the flash message:
Screenshot 2024-08-08 at 3 45 35 PM

# Implementation notes

Right now the button only appears if any one of the loans can be renewed. 

It didn't seem to make sense to lump the request renewal aspect of some of the loans in with this button, but maybe that could be useful at some point.

### DIFF
--- a/app/controllers/account/loans_controller.rb
+++ b/app/controllers/account/loans_controller.rb
@@ -1,11 +1,21 @@
 module Account
   class LoansController < BaseController
+    include Lending
+
     def index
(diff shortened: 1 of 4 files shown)


jim (approved):
Code looks great!

I'm wondering if members will find it confusing when this button only renews certain items, but I don't think that's a problem with the PR as much as something we might need to communicate better in other ways.

PR #1628, opened 2024-08-08
Standard rb fixes

crismali:
# What it does

This is just a bunch of standard rb fixes for everything left in the todo list except for the old migrations.

# Why it is important

Takes us a little closer to being fully standard rb compliant

# Implementation notes

I added a migration for missing uniqueness constraints. I did a smoke test in blazer to make sure there weren't already records that violated the constraints but we might want to check those tables out again before deploying.


jim (approved):
This looks good to me. We could pull down a database from production and run the migration to verify there aren't any problematic records.

phinze (approved):
Mmm code cleanup 🍩

phinze:
I'm comfortable watching for any issues with migrations and addressing them in the deploy, so I'm shepherding this out now 🐑

Issue #1512, opened 2024-05-29
Edit reservation items during pickup


jim:
When someone comes to pickup their reservation, we might need to tweak what items they are checking out. This means we can edit items on the reservation before checkout.

Editing _after_ checkout would just mean checking items in or out.

Related notes from a Q&A:

What happens when you add an item that wasn’t on the original reservation? Similarly, can you view the original reservation or see what’s different?
* Not fully implemented yet, but:
* It’s going to let you add it, but it’s going to highlight it in a certain way and it’s going to make a different sound and make it very clear that it’s not in the expected things
* You’ll be able to say “yes this is what I mean”
* Important for us to see why: is this an extra item? Is this too many of expected item?
* MyTurn creates a popup for this where you have to go click before you can get back to scanning - important for us to be able to stay in flow while scanning, especially when adding something at last minute
* There will be some issues you can override and others (like item in maintenance) that you’ll have to resolve before you can check out. Over time we can add automation to help with common edge cases.

jim:
Closed by #1639

PR #1639, opened 2024-08-15
Add PendingReservationItem and a few uses of it

jim:
# What it does

Adds graceful handling of items that aren't a part of a reservation being scanned while someone is being built. This takes place e.g. the morning that a reservation is scheduled to be picked up.

We store these items in the database so that they are safely persisted even if you close/reload the page. As this feature develops further, there will be a way to update the reservation's holds so that the item represented by a pending item can be added to the reservation pickup normally (and the pending item deleted).

This is the first part of #1512.

# Why it is important

If we're scanning hundreds of items, the app needs to persist which ones were unexpected so the staff member can review them and decide what to do.

# UI Change Screenshot

Screenshot 2024-08-14 at 11 13 37 PM


app/controllers/admin/reservations/reservation_loans_controller.rb
```diff
@@ -16,6 +16,11 @@ def create
             quantity: @reservation_hold.quantity
           )
         else
+          if reservation_loan_params[:reservable_item_id].blank?
```
crismali, line 19: This looks like it might make sense as an `elsif` in the larger `if`

crismali (approved):

PR #1641, opened 2024-08-18
Send reservation requested emails to all admin users

### LINKED ISSUE #1531: Send reservation requested emails to all admin users
We want all staff to be notified when a reservation is requested.
* Initially will can send all “staff” related emails to admin users
* Eventually we might more control, but that is enough for now.

This needs to wait to be worked on until we're sending those emails at the appropriate time.

katielentes:
# What it does

Sends email to admins when a member submits a reservation

# Why it is important

Closes #1531 

# UI Change Screenshot
Screenshot 2024-08-17 at 10 49 22 PM

# Implementation notes
This doesn't send the mailer when an admin requests a reservation through `ReservationsController`, since it doesn't seemed like review is required when a reservation is created this way.

### DIFF
--- a/app/mailers/reservation_mailer.rb
+++ b/app/mailers/reservation_mailer.rb
@@ -2,8 +2,9 @@ class ReservationMailer < ApplicationMailer
   def reservation_requested
     @reservation = params[:reservation]
     @library = @reservation.library
+    @to_emails = User.select { |user| user.has_role?("admin") && user.library_id === @library.id }.pluck(:email)
     @subject = "Reservation submitted for #{@reservation.started_at.to_fs(:short_date)}"
-    mail(to: @reservation.submitted_by.email, subject: @subject)
+    mail(to: @reservation.submitted_by.email, subject: @subject, bcc: @to_emails)
   end
 
   def reviewed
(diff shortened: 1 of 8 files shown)


app/mailers/reservation_mailer.rb
```diff
@@ -6,6 +6,14 @@ def reservation_requested
     mail(to: @reservation.submitted_by.email, subject: @subject)
   end
 
+  def review_requested
+    @reservation = params[:reservation]
+    @library = @reservation.library
+    @admins = User.where(library_id: @library.id).where(role: ["admin", "super_admin"]).pluck(:email)
```
jim, line 12:
```suggestion
    @admins = User.where(library_id: @library.id).where(role: ["admin"]).pluck(:email)
```

I think we don't want to send emails to `super_admin` users. That wasn't super well specified in the task, but super admins are the folks who (in theory) would edit which libraries are in the system. It's a mostly unused role since we are only supporting CTL currently.

jim (changes requested):
This is awesome! The code looks good and I love the test coverage. Just one little tweak before we merge this in.

app/mailers/reservation_mailer.rb
katielentes: Ah okay that makes sense! Thanks for the context! Change has been made :)

jim (approved):
Awesome! Thanks for making that tweak. This looks good.

Issue #1634, opened 2024-08-12
Reservation reviewing should happen within the reservation layout


jim:
Entering a reservation review currently looks like this:

![Image](https://github.com/user-attachments/assets/9956012b-b576-4238-afdd-73aa13170307)

It should happen within a tab in the reservation layout (maybe in a tab like the questions).

PR #1646, opened 2024-08-19
Move reservation reviewing into a tab

crismali:
# What it does

Instead of reviewing a reservation on its own page, now it can be reviewed in a "Review" tab

# Why it is important

Takes care of #1634 

# UI Change Screenshot

Pending reservation:
Screenshot 2024-08-19 at 4 44 04 PM

Requested reservation on the items tab:
Screenshot_2024-08-19_at_4_44_16 PM

Requested reservation on the review tab:
Screenshot 2024-08-19 at 4 44 29 PM

Immediately after approving a requested reservation:
Screenshot 2024-08-19 at 4 44 47 PM

The review notes tab of an approved reservation:
Screenshot 2024-08-19 at 4 44 57 PM

# Implementation notes

This is implemented the way the other tabs on the reservation show page are. I think from an accessibility perspective these tabs are probably more like a subnav rather than capital-T "Tabs" but that might be a more general thing to investigate.


phinze (approved):
LGTM!

PR #1648, opened 2024-08-20
Swapped the yarn add playwright command with a generic yarn install

crismali:
# What it does

Instead of running `yarn add` for playwright, this just runs `yarn install` in the `bin/setup` script. This way the version of playwright that's listed in the package.json is ultimately installed.

# Why it is important

I was running into an issue where running `bin/setup` would change the playwright version from 1.46.0 to 1.45.0 in the package.json. If I cleared those changes it would trip up the capybara/playwright immigration and tell me playwright wasn't actually installed.

# Implementation notes

I think there's a case to be made to lock playwright to what the ruby implementation considers a compatible version (via `bundle exec ruby -e 'require "playwright"; puts Playwright::COMPATIBLE_PLAYWRIGHT_VERSION'`). Maybe a general dependency update changed the package.json and these got out of sync and we just need to figure out a better way to keep them in sync?

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -16,9 +16,8 @@ FileUtils.chdir APP_ROOT do
   puts "== Installing dependencies =="
   system("bundle check") || system!("bundle install")
 
-  puts "== Installing playwright yarn package =="
-  playwright_version=`bundle exec ruby -e 'require "playwright"; puts Playwright::COMPATIBLE_PLAYWRIGHT_VERSION'`.strip
-  system!("yarn add -D 'playwright@#{playwright_version}'")
+  puts "== Installing yarn packages =="
+  system!("yarn install")
 
   if ENV["USING_NIX"] == "1"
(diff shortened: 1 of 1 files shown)


phinze (approved):
Duplicating some stuff I said in slack...

I've been having the same issue - the original blog post we were following said to do this, but I think it's just not the best approach https://justin.searls.co/posts/running-rails-system-tests-with-playwright-instead-of-selenium/

Hm but yes dependabot has taken us ahead of the ruby driver's stated compatibility

Here's the Ruby driver's equivalent update, still in draft mode https://github.com/YusukeIwaki/playwright-ruby-client/pull/301

I wonder if it's better for us to have dependabot ignore playwright on the javascript side and then we just remember to follow ruby updates with equivalent JS ones

(We have the same issue with Rails because it's released as simultaneous Ruby + JS updates and dependabot can't do cross-dependency-type batches)

Regardless this is the right next step I think!

Issue #1526, opened 2024-06-05
Add pickup and drop-off times to reservations


jim:
These represent the appointment slot that will be used for pickup and are separate from the started_at and ended_at times on the reservation. This can be based on how we do it with appointments, but it might be better to have an association between the reservation and the appointment slot event instead of copying the dates onto the reservation.

* A pickup time must be selected before the reservation is submitted, but it can be created in the pending state without one.
* A drop-off time can be set either before submitting the reservation for review or at a later time.

PR #1652, opened 2024-08-25
Add pick-up and drop-off times to reservations for admins and members

crismali:
# What it does

This allows admins to select pick-up and drop-off times when creating or updating a reservation and adds a requirement that the pick-up time must be selected before a reservation becomes "requested"

Members can also add these times when creating a reservation but at the moment there doesn't seem to be a way for them to edit a reservation without typing in the url.

# Why it is important

Implements #1526 

# UI Change Screenshot

New reservation selecting a pickup time:
Screenshot 2024-08-25 at 12 01 06 PM

New reservation selected pickup time:
Screenshot 2024-08-25 at 12 01 16 PM

New reservation validation errors:
Screenshot 2024-08-25 at 12 01 47 PM

Show reservation with pickup/dropoff times
Screenshot 2024-08-25 at 12 02 10 PM

Requesting a reservation without a pickup time:
Screenshot 2024-08-25 at 12 12 25 PM

Member reservation show with pickup/dropoff times:
Screenshot 2024-08-25 at 12 17 30 PM

# Implementation notes

The reservation now belongs to a pickup event and dropoff event (both optionally). Moving the reservation from "pending" to "requested" (but really basically any other status) requires the pickup event. This will probably need to be updated once canceling or duplicating or other reservation features are implemented.

If someone goes to edit a reservation where the pickup or dropoff events have already passed they'll still appear in the edit dropdowns.

### DIFF
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -21,7 +21,7 @@ class Reservation < ApplicationRecord
   has_many :answers, dependent: :destroy
   belongs_to :reviewer, class_name: "User", optional: true
   belongs_to :organization
-  belongs_to :submitted_by, class_name: "User", optional: false
+  belongs_to :submitted_by, class_name: "Member", optional: false
   belongs_to :pickup_event, class_name: "Event", optional: true
   belongs_to :dropoff_event, class_name: "Event", optional: true
 
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -31,8 +31,8 @@ class Reservation < ApplicationRecord
   validates :name, presence: true
   validates :started_at, presence: true
   validates :ended_at, presence: true
-  validates :pickup_event, presence: true, if: :must_have_pickup_event?
   validates_associated :reservation_holds
+  validate :must_have_pickup_event
 
   after_initialize :restore_manager
   before_validation :move_ended_at_to_end_of_day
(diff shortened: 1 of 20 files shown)


test/factories/reservations.rb
```diff
@@ -7,19 +7,70 @@
...
+      pickup_event { association(:event) }
     end
 
     trait :rejected do
       status { "rejected" }
       notes { "This reservation has been rejected :(" }
+      pickup_event { association(:event) }
+    end
+
+    trait :obsolete do
```
jim, line 35: Thanks for adding these missing traits to the factory!

app/models/reservation.rb
```diff
@@ -70,4 +75,18 @@ def move_ended_at_to_end_of_day
   def validate_reservation_dates
     errors.add(:ended_at, "end date must be after the start date") if started_at.present? && ended_at.present? && started_at.to_date >= ended_at.to_date
   end
+
+  def must_have_pickup_event
+    return if manager.pending? || pickup_event.present?
+
+    errors.add(:pickup_event_id, "can't be blank")
+  end
```
jim, line 83: This seems like a pretty nice way to have validations conditionally apply based on the reservation's state.

jim (approved):
Awesome!

We'll likely want to iterate on the UI here eventually- like probably the pickup event has to be on/after the reservation start date, etc. But we can add more of that later on since it'll complicate the UI if things get too dynamic.

PR #1663, opened 2024-09-05
Swap member and user relationship

crismali:
# What it does

With these changes members will belong to users instead of users belonging to members.

# Why it is important

Having users belong to members is more of a historical artifact than a real design choice. This keeps things to be more like what people would expect.

# Implementation notes

I added some SQL to the migration to make sure the current member/user relationships are transferred over. There are probably some other improvements that could be made (like the columns could be non-null if we're sure there's a user for every member or pulling some of the overlap between members and users into a profile, etc), but I focused mainly on inverting the relationship.


jim (approved):
Incredible! Thanks for taking on this bit of tech debt- it'll be nice to not get tripped up by this any more.

PR #1664, opened 2024-09-05
Put the item's name ahead of its number on the member view of the items index

crismali:
# What it does

Rearranges some markup and bumps up the item name's font size.

# Why it is important

[People were suggesting this change in the recent survey](https://chicagotoollibrary.slack.com/archives/C05RB4B5M1N/p1725532818050629)

# UI Change Screenshot

Screenshot 2024-09-05 at 4 32 49 PM

# Implementation notes

This doesn't impact the admin items index since the style changes use a class name that's only on the member view.


jim (approved):
:shipit:

Issue #1520, opened 2024-06-03
As a group owner, I must answer reservation questions before submitting a reservation for review


jim:
These are the questions created by staff. They must be answered in order to submit a reservation. Always use the latest set of questions saved on the Library, regardless of what was already stored in the reservation if it is edited.

PR #1674, opened 2024-09-16
Member facing reservation questions

crismali:
# What it does

Members now have to answer any questions in the system when making a reservation.

# Why it is important

#1520 

# UI Change Screenshot

The new reservation form with a text and number question:
Screenshot 2024-09-16 at 2 57 21 PM

The new reservation form with errors:
Screenshot 2024-09-16 at 2 57 32 PM

The show reservation page with answered questions
Screenshot 2024-09-16 at 2 58 07 PM

# Implementation notes

Right now a `nil` value for a `number` field is coerced to a `0` so `number` questions/answers can never really be `blank?`. Not sure if this is nice or annoying but I'm happy to change it up if that's what folks prefer.

I didn't get into editing/updating since we're not likely going to have that as a flow in this part of the app. I thought about removing those actions/routes/views but I figured that'd likely end up happening once other ways of interacting with the reservation are worked out.

In order to make the questions optional for admins but required for members, I tried something a little weird but I think it works well enough: I subclassed `Answer` into a `RequiredAnswer` that has an additional validation (same table and everything else though). The risks for this are if we forget to use `RequiredAnswer` and instead use `Answer` in member-facing context. I don't think that's particularly more risky than other implementations of this kind of conditional validation logic that I've seen, but I'm always open to learning new approaches!

### DIFF
--- a/app/controllers/account/reservations_controller.rb
+++ b/app/controllers/account/reservations_controller.rb
@@ -14,10 +14,12 @@ module Account
 
     def new
       @reservation = Reservation.new
+      set_required_answers
       set_reservation_slots
     end
 
     def edit
+      set_required_answers
       set_reservation_slots
     end
 
@@ -26,8 +28,9 @@ module Account
 
       if @reservation.save
         # TODO use the current user's actual organization
-        redirect_to account_reservation_url(@reservation), notice: "Reservation was successfully created."
+        redirect_to account_reservation_url(@reservation), success: "Reservation was successfully created."
       else
+        set_required_answers
         set_reservation_slots
         render :new, status: :unprocessable_entity
       end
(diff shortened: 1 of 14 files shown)


jim (commented):
Subclassing `Answer` with `RequiredAnswer` is an interesting approach. I think the only other way we have handled this in the app is setting an attribute on a model to toggle which "mode" it's in [like we do in `Appointment`](https://github.com/chicago-tool-library/circulate/blob/1ea9f114f7ae65fa631a1ef9aea8244d06983b21/app/models/appointment.rb#L9-L10). But I'm OK with us trying this new way and seeing how far it gets us, especially since these models are pretty simple.

> Right now a nil value for a number field is coerced to a 0 so number questions/answers can never really be blank?. Not sure if this is nice or annoying but I'm happy to change it up if that's what folks prefer.

I think we'd want to not coerce to zero so that a required field has to have a value entered. Otherwise someone can just hit the submit button without entering a value, but my understanding is we'd like to ensure that all of the questions have been actually answered.

crismali:
Just took care of the coercion issue (I was wrong about it converting `nil`, it was actually converting empty strings).

jim (approved):

PR #1678, opened 2024-09-21
Resolve pending reservation items

### LINKED ISSUE #1676: `bin/setup` opens browser windows
(At least on a Mac), when running `bin/setup`, several browser windows are opened by `letter_opener` for the email confirmation emails send to users created by the script. We should make it so this doesn't happen.

jim:
# What it does

* This adds the start of "resolving" `PendingReservationItems`, which means we either delete them or expand the reservations to include them.
* I added some reservations to `bin/setup` to make development easier.
* I also fixed #1676 in the process of the last task.

### DIFF
--- a/test/system/admin/reservations_test.rb
+++ b/test/system/admin/reservations_test.rb
@@ -339,20 +339,46 @@ class AdminReservationsTest < ApplicationSystemTestCase
     refute_text hammer.name
   end
 
-  test "adding unreserved items to a pickup" do
+  test "adding unreserved items to a pickup and then removing them" do
     hammer_pool = create(:item_pool, name: "Hammer")
     hammer = create(:reservable_item, item_pool: hammer_pool)
 
     reservation = create(:reservation, :building)
     visit admin_reservation_pickup_path(reservation)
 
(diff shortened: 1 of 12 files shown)


test/system/admin/reservations_test.rb
```diff
@@ -339,20 +339,46 @@ def create_events
...
+  test "adding unreserved items to a pickup and merging a new item into the reservation" do
+    hammer_pool = create(:item_pool, name: "Hammer")
+    hammer = create(:reservable_item, item_pool: hammer_pool)
+
+    reservation = create(:reservation, :building)
+    visit admin_reservation_pickup_path(reservation)
 
-    refute_text "1 item scanned that did not match the reservation"
+    assert_difference -> { reservation.reservation_holds.count } => 1,
+      -> { reservation.pending_reservation_items.count } => 0 do
```
crismali, line 371: I didn't know this was a thing you could do with `assert_difference`!

jim: I only learned about it when writing this! I had a bunch of nested `assert_difference` calls and was like "there must be a better way". I think this is better?

PR #1683, opened 2024-09-24
Admin organization members

crismali:
# What it does

# Why it is important

# UI Change Screenshot

# Implementation notes


app/models/organization_member.rb
```diff
@@ -1,4 +1,28 @@
...
+    admin: "admin",
+    member: "member"
+  }
+
+  validates :full_name, presence: true
+  validates :organization_id, uniqueness: {scope: :user_id}
+
+  def self.create_with_user(email:, **organization_member_attributes)
+    organization_member = new(organization_member_attributes)
+    user = User.find_by_case_insensitive_email(email) || User.new(email:, password: Devise.friendly_token)
```
crismali, line 18: If a user is created here, a confirmation link is automatically sent to them, but they probably need to do a "forgot password" flow to actually confirm themselves and sign in.

app/models/user.rb
```diff
@@ -42,4 +59,14 @@ def self.serialize_from_session(key, salt)
...
+
+  private
+
+  def password_required?
+    !persisted? || !password.nil? || !password_confirmation.nil?
+  end
+
+  def email_required?
+    true
+  end
```
crismali, line 71: These private methods are also taken from Devise's Validatable.

app/views/admin/organization_members/new.html.erb
```diff
@@ -0,0 +1,21 @@
...
+    <%= user_form.text_field :email, type: "email" %>
+  <% end %>
+
+  <% role_options = OrganizationMember.roles.values.sort.map { |role| [role.capitalize, role] } %>
+  <%= form.select :role, options_for_select(role_options, @organization_member.role) %>
+
+  <div>
+    <%= form.submit %>
+  </div>
+<% end %>
```
crismali, line 21: The `new` and `edit` templates could share a `form` partial, but there'd need to be a decent amount of logic in it (figuring out the form url, whether or not to include the email field) so this might be a small amount of duplication but I think it's a bit easier to reason about.

app/views/admin/organizations/show.html.erb
```diff
@@ -1,15 +1,36 @@
...
+    <% @organization.organization_members.order(:full_name).each do |organization_member| %>
+      <tr>
+        <td>
+          <%= link_to organization_member.full_name, admin_organization_member_path(organization_member) %>
+        </td>
+        <td>
+          <%= organization_member.role.capitalize %>
+        </td>
+        <td>
+          <%= link_to organization_member.user_id, admin_user_path(organization_member.user) %>
```
crismali, line 31: I added this link to the user's page before I added the role because the table looked weirdly empty. Not sure if we want to keep it or if it's convenient/confusing.

config/routes.rb
```diff
@@ -74,7 +74,10 @@
   get "/auth/failure", to: "volunteer/sessions#failure"
 
   namespace :admin do
-    resources :organizations
+    resources :organizations do
+      resources :organization_members, only: [:new, :create]
+    end
+    resources :organization_members, only: [:show, :edit, :update, :destroy]
```
crismali, line 80: Two sets of resources:

- one that includes the `organization_id` in the url
- one that contains the `organization_member_id`

db/migrate/20240924190657_ensure_unique_email_addresses_per_library.rb
```diff
@@ -0,0 +1,11 @@
...
+  def up
+    remove_index :users, %i[email library_id]
+    execute "CREATE UNIQUE INDEX index_users_on_lowercase_email_and_library_id ON users USING btree (lower(email), library_id);"
+  end
+
+  def down
+    execute "DROP INDEX index_users_on_lowercase_email_and_library_id;"
+    add_index :users, %i[email library_id]
+  end
+end
```
crismali, line 11: This is less related to the actual issue, but it's something that needs to be done.

app/models/organization_member.rb
```diff
@@ -1,4 +1,28 @@
...
+    organization_member.user = user
+
+    transaction do
+      raise ActiveRecord::Rollback unless user.save
+      raise ActiveRecord::Rollback unless organization_member.save
+    end
+
+    organization_member
+  end
```
crismali, line 27: This whole method could probably be moved into a service or something like that, but right now it would have no other functionality and the `OrganizationMember` is pretty small anyway.

app/models/user.rb
```diff
@@ -16,9 +18,24 @@ class User < ApplicationRecord
   }
 
   has_one :member
+  has_many :organization_members, dependent: :destroy
+  has_many :organizations, through: :organization_members
+
+  # Adapted from https://github.com/heartcombo/devise/blob/main/lib/devise/models/validatable.rb
+  # so we can scope email uniqueness to library_id
+  validates :email, presence: {if: :email_required?}
+  validates :email, uniqueness: {allow_blank: true, case_sensitive: false, if: :devise_will_save_change_to_email?, scope: :library_id}
```
crismali, line 27: This validation is why we needed to drop `:validatable` above. `:validatable` adds this validation without the `scope: :library_id`

app/models/organization_member.rb
jim: It might confuse people a bit if they get an email that says they need to confirm but it doesn't work.

OTOH, we're likely going to start testing with single person orgs, so we can improve this invite messaging/flow in the future.

app/views/admin/organization_members/new.html.erb
jim: I agree that the duplication is easier to work with than a lot of template logic.

app/views/admin/organizations/show.html.erb
jim: Hmm, it's probably OK to leave for now. I'm not sure how useful it will be, but it's not hurting anything.

db/migrate/20240924190657_ensure_unique_email_addresses_per_library.rb
jim: Excellent, thanks for fixing this.

jim (approved):
This all looks really good to me.

Regarding extracting a `Profile` from `Member`, that's probably the right thing to do? I don't think it needs to be in this PR, though. Probably worth thinking through the ramifications a bit before making that shift.

PR #1684, opened 2024-09-27
Don't allow items that aren't active to be added to appointments.

jim:
# What it does

Fixes a long-standing issue where members can add items that are in maintenance to their appointments. This resulted from some gaps in logic in the appointment and hold code.

# Why it is important

Items in maintenance can't be checked out, and it's a bad experience to allow someone to think they can pick something up when they can't. It also created more work for staff, who have to sort it all out when a member is confused by this.

# Implementation notes

* Only items that have a `status` of `active` are shown on the appointment editing page.
* The logic on the holds page has been fixed so that it says "you are # 1 in line" instead of "ready for pickup" for items in maintenance.

We don't actually let members place new holds on items in maintenance, and at once point we deleted all holds for items when they entered this state. We stopped doing that because it confused members to have their holds disappear, but didn't fix this gap in logic that resulted from that change.

### DIFF
--- a/app/controllers/account/appointments_controller.rb
+++ b/app/controllers/account/appointments_controller.rb
@@ -69,7 +69,7 @@ module Account
     end
 
     def load_holds_and_loans
-      @holds = Hold.active.includes(member: {appointments: :holds}).where(member: @member)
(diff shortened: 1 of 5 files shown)


crismali (approved):
👍 👍 👍 👍 👍

PR #1690, opened 2024-10-01
Only update user unconfirmed email if user exists and email was changed.

jim:
# What it does

Fixes an issue that was reported by staff. If a member updates their profile, their email is always marked as unconfirmed. This PR limits this to situations where the email has actually change.


crismali (approved):
👍

Issue #893, opened 2021-12-04
Reports admin menu drop down scroll on library laptops


jim:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

PR #1692, opened 2024-10-03
Remove unnecessary spacing resulting from specter upgrade

jim:
# What it does

The spectre.css upgrade in #1686 ended up messing up a bunch of spacing in the app, mostly due to changes to the default styles of `p`, `ul`, and `li` elements. I clicked around and fixed the most glaring visual issues, but I wouldn't be surprised if there are more.

I also ended up fixing #893 in the process.


jim:
I'm going to merge this without review in order to fix these visual bugs ASAP.

PR #1695, opened 2024-10-05
Vendored spectre-css to update its scss syntax around colors and math

crismali:
# What it does

This swaps out deprecated sass functions and syntax for their modern counterparts.

# Why it is important

The times are a-changin' and so must we.

# UI Change Screenshot

An example of the deprecation message:
Screenshot 2024-10-05 at 3 48 51 PM

# Implementation notes

[Sass has a migration tool for these changes](https://sass-lang.com/documentation/cli/migrator/) that I ran on the affected files. The TL;DR of these changes is that `darken`, `lighten`, and using `/` for division are all deprecated.

This PR looks super huge and it kind of is. Our primary css framework (Spectre CSS) hasn't had an update in about 4 years and deprecation warnings in the package were showing up in our build process. I copied the main scss code from the repo into our `app/assets` and ran the migration tool on those. 

This is kind of a short term fix to get rid of some noise. Longer term, we should migrate off of Spectre to something else ([I've added an issue for this](https://github.com/chicago-tool-library/circulate/issues/1694)).


jim (approved):
This is a good stop-gap that'll buy us some time to figure out the right direction long term.

Issue #1682, opened 2024-09-24
Tweaks to individual signup process


jim:
1. Change text underneath "Tool Library memberships are pay-what-you-can" to:
> This means you choose how much your one-year membership costs.
Please choose how much you can contribute to this community service. Membership contributions are the Tool Library’s main source of income to pay our staff, pay our rent, and buy and maintain our tools.
We recommend contributing $100 for a one-year membership, but this is only a recommendation. Many people pay more and many people pay less.
Please note that your membership contribution is the only payment you will make to the Tool Library this year. There are no late fees, replacement fees, or rental fees.

2. Also a few changes here:
![Image](https://github.com/user-attachments/assets/c49ff83f-1590-4d8e-b856-99ede72027df)

3. Also is it possible to set the online payment minimum to $5?

[Slack reference](https://chicagotoollibrary.slack.com/archives/C05RB4B5M1N/p1726671579813779)

PR #1699, opened 2024-10-08
Tweak sign up for clarity around payments

crismali:
# What it does

Changes some copy and has the form require a minimum of a $5 online donation

# Why it is important

Takes care of the requests in #1682 

# UI Change Screenshot

New copy:
Screenshot 2024-10-07 at 7 18 53 PM

Attempting to donate $4:
Screenshot 2024-10-07 at 7 19 05 PM

# Implementation notes

Happy to tweak any formatting or anything else. 

I put a validation in the form for the $5 minimum but if someone tries to get around that I also adjusted the minimum in `MembershipPaymentForm`. This also changes the behavior for renewals which isn't in the card but seems like the kind of thing we'd want changed everywhere.

### DIFF
--- a/app/forms/membership_payment_form.rb
+++ b/app/forms/membership_payment_form.rb
@@ -4,7 +4,7 @@ class MembershipPaymentForm
   attr_accessor :amount_dollars
 
   validates_numericality_of :amount_dollars,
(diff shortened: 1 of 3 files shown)


jim (approved):
Awesome. Thanks for getting this done!

PR #1707, opened 2024-10-16
Add started_at and ended_at dates to reservation_hold.

jim:
# What it does

Adds `started_at` and `ended_at` dates to `ReservationHold` as the first step of making the reservation system more flexible. By moving these dates onto the reservation holds, we'll be able to represent situations like subsets of items on a reservation being extends (say, when someone doesn't return them along with the rest of their order).


crismali (approved):
👍

PR #1709, opened 2024-10-17
Improve UI of reservation items screen

jim:
# What it does

* Renames two reservation tabs: `items` becomes `reservation` and `pickup` becomes `items`. I think this will make more sense to staff (it already does to me).
* Rebuilds the Items tab to use a table and be more information-dense. This will help the UI scale, be more scannable, and provide a space for other controls as needed.
* Adds the concept of a "default tab" when viewing reservations in the admin. This is so we can have the more relevant tab already selected when folks go to visit a reservation. Which tab this is will vary over the lifecycle of a reservation. The code here is a start but will need adjustment as we continue to build out functionality.

# UI Change Screenshot
Screenshot 2024-10-17 at 10 21 34 AM
Screenshot 2024-10-17 at 10 21 23 AM

### DIFF
--- a/app/controllers/admin/reservations/reviews_controller.rb
+++ b/app/controllers/admin/reservations/reviews_controller.rb
@@ -3,7 +3,7 @@ module Admin
     class ReviewsController < BaseController
       def edit
         unless @reservation.manager.can?(:approve)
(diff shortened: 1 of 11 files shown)


crismali (approved):
👍

PR #1723, opened 2024-10-31
Move color overrides into the main spectre variables file.

jim:
# What it does

When we updated the Sass to use `@use` instead of `@import` in #1716, some of the built-in framework colors stopped being overridden properly. We were using `@import` to just replace global variables for `$primary-color`, etc, but with `@use` everything is namespaced so that no longer works.

I couldn't figure out a nicer way to do this, so I just moved our customization into the main `spectre-css/_variables` file and commented out what was there originally.

Here are a couple places I noticed odd colors:
Screenshot 2024-10-30 at 9 53 44 PM
Screenshot 2024-10-30 at 9 53 50 PM

And after this PR, things are back to normal:
Screenshot 2024-10-30 at 9 54 05 PM
Screenshot 2024-10-30 at 9 54 10 PM

### DIFF
--- a/app/assets/stylesheets/globals.scss
+++ b/app/assets/stylesheets/globals.scss
@@ -2,35 +2,10 @@
 
 @use "mixins";
 
-$primary-color: #2f7890;
-$primary-color-light: color.adjust(#2f7890, $lightness: 20%, $space: hsl);
-$secondary-color: color.adjust(#cdabbe, $lightness: -10%, $space: hsl);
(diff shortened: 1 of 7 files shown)


crismali (approved):
👍  I think this is a good approach for now. We can come up with something better down the road or make a more concerted effort to move into way of styling that isn't getting deprecated every other week. I can look around at what more recent, idiomatic approaches to this in Rails are.

PR #1724, opened 2024-11-02
Assorted improvements to user email reset flow.

jim:
# What it does

We found a couple issues with the way we handle users resetting their emails. The main changes are:

* Show the new email in the "resend confirmation email" modal to staff
* Tweak the app layout so that the confirmation success flash message is shown to members
* Store the email that a notification is going to instead of the value of `user.email` in `Notification#address`.


crismali (approved):
👍

PR #1736, opened 2024-11-07
Tweak lograge config

### LINKED ISSUE #1568: Appsignal logs contain the string "user_id" with no actual ID attached
**Describe the bug**

It looks like the raw token `user_id` is landing in our production logs. We should investigate!

**Screenshots**

2024-06-25 at 14 58 29@2x

jim:
Update our [lograge](https://github.com/roidrage/lograge) configuration in two ways:

1. Disable lograge in development and test environments. I think the default Rails log format is easier to read for a human, and we mostly use log rage in production to be compatible with AppSignal's log parsing.
2. Adjusts the way we add additional properties to the logs so that we only add a `user_id` when there is a value. I think this will fix #1568.

### DIFF
--- a/config/initializers/lograge.rb
+++ b/config/initializers/lograge.rb
@@ -1,12 +1,12 @@
 Rails.application.configure do
-  config.lograge.enabled = true
+  config.lograge.enabled = Rails.env.production?
 
   # Set a few custom parameters on log lines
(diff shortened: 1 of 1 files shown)


crismali (approved):
Looks great

PR #1738, opened 2024-11-08
Add flash_highlight and clean up bits of old highlighting.

jim:
# What it does

* Adds a new/better way to highlight areas of the page that are updated. This is mostly useful on screens where there are long lists of items, as they sometimes move around as changes are made. I added a `flash_highlight` helper function to manage this- it's easy to just call it in a controller with an `ActiveRecord` object.
* Enables [morphing for page refreshes](https://jonsully.net/blog/turbo-8-page-refreshes-morphing-explained-at-length/#stay-on-the-rails), which should make some interactions where a user is redirected back to the same page they're already on nicer.
* Cleans up some old styles that weren't being used.

# UI Change Screenshot

### DIFF
--- a/app/assets/stylesheets/styles.scss
+++ b/app/assets/stylesheets/styles.scss
@@ -288,68 +288,6 @@ $photo-width: 187px;
   max-width: 100%;
 }
 
-.categories-table {
-  display: grid;
-
-  grid-template-columns: 1fr 100px 200px;
-  grid-row-gap: 0.3rem;
-  grid-column-gap: 0.6rem;
-
-  .categories-table-row {
-    display: contents;
-    &:hover .categories-table-name,
-    &:hover .categories-table-count {
-      background: variables.$bg-color-dark;
-    }
-    .categories-table-name.highlight {
-      animation: highlight-fade-out 1.5s;
-    }
-  }
-
-  .categories-table-header {
-    font-weight: bold;
-  }
-
-  .categories-table-name {
-    grid-column-start: 1;
-    padding-left: 0.2rem;
-    &.nested-1 {
-      margin-left: 1rem;
-    }
-    &.nested-2 {
-      margin-left: 2rem;
-    }
-    &.nested-3 {
-      margin-left: 3rem;
-    }
-    &.nested-4 {
-      margin-left: 4rem;
-    }
-    &.nested-5 {
-      margin-left: 5rem;
-    }
-    &.nested-6 {
-      margin-left: 6rem;
-    }
-    &.nested-7 {
-      margin-left: 7rem;
-    }
-    &.nested-8 {
-      margin-left: 8rem;
-    }
-    &.nested-9 {
-      margin-left: 9rem;
-    }
-    &.nested-10 {
-      margin-left: 10rem;
-    }
-  }
-
-  .categories-table-count {
-    text-align: center;
-  }
-}
-
 .responsive-table {
   display: flex;
   flex-wrap: wrap;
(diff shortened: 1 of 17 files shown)


app/assets/stylesheets/styles.scss
```diff
@@ -288,68 +288,6 @@ $photo-width: 187px;
   max-width: 100%;
 }
 
-.categories-table {
```
jim, line 291: This wasn't used anywhere.

app/controllers/admin/loans_controller.rb
```diff
@@ -19,6 +19,7 @@ def create
       @loan = create_loan(@item, @member)
 
       if @loan.persisted?
+        flash_highlight(@loan)
```
jim, line 22: This is how to register a highlight on the next page load.

app/controllers/admin/base_controller.rb
```diff
@@ -35,5 +37,15 @@ def are_appointments_enabled?
         render_not_found
       end
     end
+
+    def flash_highlight(id_or_object)
+      identifier = if String === id_or_object
```
crismali, line 42:
```suggestion
      identifier = if id_or_object.is_a?(String)
```

Not a serious change or anything, I just find the `is_a?` to be a bit more readable than `===` (there might be some subtle difference in behavior but I don't think it would impact this)

app/javascript/lib/highlight.js
```diff
@@ -0,0 +1,18 @@
+// If an element to highlight is declared in the document head, add a CSS class
+// to highlight it. Remove the class when the highlight animation completes
+// so it can be triggered on the same element again if needed.
+export function highlightElement () {
+  const metaTag = document.querySelector("meta[name='highlight-element']")
```
crismali, line 5: I probably just missed it, but what is actually putting this meta tag into the dom?

app/javascript/lib/highlight.js
```diff
@@ -0,0 +1,18 @@
...
+  if (metaTag) {
+    const elementIDs = metaTag.getAttribute('content').split(',')
+    elementIDs.forEach(elementID => {
+      const element = document.getElementById(elementID)
+      if (element) {
+        element.addEventListener('animationend', function (event) {
```
crismali, line 11: I don't think it's bad to clean up the class once it's done, but another option to consider is [specifying an `animation-fill-mode` of `forwards` in the css](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode) (I think you can pass it as part of `animation:`). With that you could drop this `animationend` event listener, although the highlight class would remain. I'm also not sure if some of that morphing stuff would mess this more css-centric approach up.

crismali (approved):
Looks good to me! I really like these (no way around the pun) flashy kinds of interactions. 👍

app/controllers/admin/base_controller.rb
crismali: also works: 

```rb
case id_or_object
when String
  #...
else
  #...
end
```

I guess the real lesson is that the `===` feels odd to me, but I'm pretty sure that's just a me thing 😆

app/javascript/lib/highlight.js
jim: You didn't! I forgot to push one commit. e56808c contains the changes to the admin layout that are required to make both the flash highlighting as well as the morphing refresh work.

app/javascript/lib/highlight.js
jim: Interesting. If the class wasn't removed I think it would mean that a future update done via morphing replacement wouldn't trigger the animation. I'll keep this for now to remove the class, but we can re-visit this if it's ever a problem.

PR #1739, opened 2024-11-08
First pass at cleaning up SASS

crismali:
# What it does

- splits the css between admin and member facing ("application")
- removes old homepage code
- removes generally unused styles
- removes unused JavaScript controllers
- reformats a few files

# Why it is important

This is a good first step towards moving off of Spectre CSS. It eliminates a lot of noise in the styles and clarifies how much effort it will take to make future changes.

There's plenty more that could be done but I think making this smaller if still quite large PR would make things easier than a PR that rewrites the whole world of styles and templates.

# UI Change Screenshot

Everything should look and work the same!

# Implementation notes

My understanding is that when Rails compiles the assets it just runs the `build:css` command so I made sure that would still result in all of the css being built. There are a few more css related package.json commands and I added [Concurrently](https://www.npmjs.com/package/concurrently) to make running commands at the same time easier.

Reformatting was done by [Prettier](https://prettier.io/) using the default settings. 

To figure out if a style was unused I grepped the project for the class name minus the `.`. It's possible there's someplace that's catting strings together to build up some classes but I haven't seen much if any of that floating around the app and couldn't find anything broken on the member side in any case. 

The next steps could be any of a few things:

- another pass at removing unused application styles / refactoring the existing application styles to be more concise
- remove Spectre from the application scss to see what the damage is (and then repair said damage)
  - this could be on a gradual page-by-page basis
- reorganize the styles to be a bit more straightforward (they're not so disorganized now, but there's room for improvement)
- Rework the dev styles page to be an up-to-date resource for volunteers and make it dev or admin only


jim (approved):
Looks good to me. I clicked around the admin and nothing looked off.

Issue #1654, opened 2024-08-25
Add renewal count to renewal requests page


jim:
Here's the current page:

![Image](https://github.com/user-attachments/assets/e5d1ba02-7e6d-4ae9-9bd6-9e76317f352b)

* Add a column that shows how many times the item has already been renewed.
* It's OK to rearrange the data or combine columns if needed (the item number could be combined with the item name, etc.)

PR #1740, opened 2024-11-08
Added renewal count to the renewal requests index page

crismali:
# What it does

Added the renewal count to the renewal requests index page and rearranged a few of the columns.

# Why it is important

#1654 

# UI Change Screenshot

The renewal requests index page on a iPad mini
Screenshot 2024-11-08 at 3 28 23 PM

# Implementation notes

If this page needs to support a narrower screen size we can combine some of the columns but the iPad mini (768px wide in portrait mode) seems like it would cover most devices. Let me know if not!

### DIFF
--- a/app/views/admin/renewal_requests/index.html.erb
+++ b/app/views/admin/renewal_requests/index.html.erb
@@ -24,11 +24,9 @@
       <table class="table table-scroll">
         <thead>
           <tr>
-            <th>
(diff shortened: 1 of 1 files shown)


jim (approved):
Looks good to me! I think folks mostly access this on laptops/desktops but we can check with staff to see if that assumption is incorrect.

PR #1751, opened 2024-11-11
Use category name when linking to categories.

jim:
# What it does

Just a very small fix. This bug was preventing the display of category links in the admin items index. We were rendering a link around an empty `name` tag, which meant that the links were effectively invisible. With this fix, they show the category name again.

# UI Change Screenshot

Before: 
Screenshot 2024-11-10 at 7 56 44 PM

After:
Screenshot 2024-11-10 at 7 57 01 PM

### DIFF
--- a/app/views/admin/items/_items.html.erb
+++ b/app/views/admin/items/_items.html.erb
@@ -37,7 +37,7 @@
 
     <%= tag.div class: "items-table-categories" do %>
       <% item.categories.each do |category| %>
-        <%= link_to tag.name, admin_items_path(category: category.id) %>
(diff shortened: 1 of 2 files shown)


crismali (approved):
Looks good, but one test failed. Looks like the test data might need to be made to look a little different.

jim:
@crismali One of these days I'll open a PR with passing tests... 🎳  I added an argument to `click` so it will pick the first matching link. The first thing I tried was adding a wrapper div to target which link to click, but the css for `,items-table` is written in a way that makes adding more nesting tricky.

Issue #1743, opened 2024-11-09
Require staff to create a maintenance ticket to set an item's status to  "in maintenance "


jim:
1. Disable the "In Maintenance" option on the item edit form.
2. Add something to the hint text for that field that links to the page to create a maintenance ticket. "To set this item to be in maintenance, create a ticket."

![Image](https://github.com/user-attachments/assets/36cf2d31-6547-4ae5-94b1-8688b48a8c82)

crismali:
@jim 

If I remember correctly, you mentioned that creating a ticket should put an item into maintenance. Should that behavior be added as part of this? Locally I can create a ticket and it doesn't impact the status of the item.

jim:
@crismali Yes, that would be great! I misremembered that we already had the app automatically putting items into maintenance.

jim:
Fixed by #1755 .

PR #1755, opened 2024-11-14
Disable the maintenance option when creating an item

crismali:
# What it does

Disables the maintenance option on the admin item form and marks items as in maintenance after creating a ticket.

# Why it is important

#1743 

# UI Change Screenshot

Select dropdown for a new item:
Screenshot 2024-11-14 at 2 18 26 PM

Select dropdown when editing an item:
Screenshot 2024-11-14 at 2 17 56 PM

Expanded select dropdown when editing an item:
Screenshot 2024-11-14 at 2 18 08 PM

An item just before creating a ticket:
Screenshot 2024-11-16 at 11 36 48 AM

An item right after creating a ticket:
Screenshot 2024-11-16 at 11 36 59 AM

# Implementation notes

For now I've set it up to only put items into maintenance when creating a ticket. I could see a case for marking items as available once the ticket is resolved, but I can also imagine reasons we might not want that to happen (maybe another ticket was created afterwards or that issue is resolved because it's lost now or something).

### DIFF
--- a/app/controllers/admin/items/tickets_controller.rb
+++ b/app/controllers/admin/items/tickets_controller.rb
@@ -24,7 +24,7 @@ module Admin
         @ticket = @item.tickets.new(ticket_params)
 
         if @ticket.save
-          @ticket.item.update!(status: Item.statuses["retired"]) if @ticket.retired?
+          update_item_status!(@ticket)
           redirect_to admin_item_ticket_url(@item, @ticket), success: "Ticket was successfully created.", status: :see_other
         else
           render :new, status: :unprocessable_entity
@@ -46,6 +46,11 @@ module Admin
       end
 
       private
+
+      def update_item_status!(ticket)
+        status = ticket.retired? ? Item.statuses["retired"] : Item.statuses["maintenance"]
+        @ticket.item.update!(status:)
+      end
 
       def set_ticket
         @ticket = Ticket.find(params[:id])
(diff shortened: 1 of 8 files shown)


jim (commented):
This all looks good so far! Nice handling of conditionally adding that "new ticket" link and excellent tests (as usual 😄).

app/helpers/items_helper.rb
```diff
@@ -50,6 +52,11 @@ def category_nav(categories, current_category = nil)
     end
   end
 
+  def status_hint(item)
```
crismali, line 55: Originally I just did this in the template, but erb_lint didn't like me using `html_safe` there. Standard is okay with it though 🤷

package.json
```diff
@@ -28,7 +28,7 @@
   "devDependencies": {
     "concurrently": "^9.1.0",
     "markdown-toc": "^1.2.0",
-    "playwright": "1.46.0",
+    "playwright": "1.45.0",
```
crismali, line 31: I pulled this down a version because this needs to be in sync with the gem, but sometimes dependabot slips an update through

app/controllers/admin/items/tickets_controller.rb
```diff
@@ -47,6 +47,11 @@ def destroy
 
       private
 
+      def update_item_status!(ticket)
+        status = ticket.retired? ? Item.statuses["retired"] : Item.statuses["maintenance"]
+        @ticket.item.update!(status:)
+      end
```
jim, line 53: I wonder if there is a workflow where staff would want to create a ticket without putting an item into maintenance? Let's merge this as-is but check in at the next stakeholders about this.

jim (approved):

PR #1763, opened 2024-11-19
Fix load_production_data_for_export

jim:
# What it does

I noticed that `bin/rails load_production_data_for_export` was broken due to the change we made a while back to swap `users.member_id` with `members.user_id`. This just updates a couple queries to work with the new schema.


crismali (approved):
🌟

PR #1764, opened 2024-11-19
Return items on a reservation

jim:
# What it does

* Adds the ability for a staff member to mark items on a reservation as returned at the end of a loan.
* Restructures the "Pickups" tab to be the "Items" tab on a reservation. Now all item manipulation (both while building up the specific items to be borrowed and when they are returned later on) happen on the same page/tab.
* Deletes an unused partial: `app/views/account/reservations/_profile.html.erb`.

# Why it is important

We need to have a accept items that are returned to us 😄 

# UI Change Screenshot

# Implementation notes

* It's possible we could avoid the turbo stream shenanigans in this flow, but I started working on this before we merged in the page refresh morphing stuff in #1738. I will see if that simplification is possible while tackling future parts of the reservation flow.


crismali (approved):
full turbo 🔥

Issue #1747, opened 2024-11-09
Add sorting and filtering to the items in maintenance report


jim:
1. Use [Drew's example](https://app.chicagotoollibrary.org/blazer/queries/13-maintenance-tickets-and-items) as a template for what columns to show.
2. We'd like to click on headings to sort by that column
3. We'd like to filter by specific columns
4. The default view should show all statuses except resolved and retired.

jim:
Fixed by #1766.

PR #1766, opened 2024-11-26
Add sorting and filtering to the item maintenance report

crismali:
# What it does

Adds new data, filters, and sorting to the item maintenance report.

# Why it is important

#1747 

# UI Change Screenshot

Initial page load:
Screenshot 2024-11-26 at 12 15 47 PM

Filtered by ticket status:
Screenshot 2024-11-26 at 12 16 04 PM

Sorted by item number:
Screenshot 2024-11-26 at 12 16 30 PM

Filtered by item name:
Screenshot 2024-11-26 at 12 16 51 PM

# Implementation notes

I basically recreated the blazer report in [ActiveRecord](https://activerecord-hackery.github.io/ransack/) and put the Ransack over it. 

Right now the page will never show resolved or retired tickets since that was easier then hiding them by default and allowing a multi-select or something, but that's something I can add if it's actually useful here. 

I rolled the styles for the filter form without using Spectre as a kind of experiment and it wasn't too much, especially once they're pulled out and made more reusable.


jim (approved):
This looks good. I'm not super familiar with Ransack, but it looks like it's pretty useful for building this kind of search feature.

Re: not showing tickets in certain states at all, we can ask staff if that's something they would want to have once we've rolled this out and they've had a chance to use it.

PR #1785, opened 2024-12-11
Update to version 4 of upload-artifact

jim:
# What it does

I got an email about v3 being deprecated and will no longer work starting in January 2025. [The docs about this are here](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/). Our use case is pretty simple, so I think this is the only change that's needed.


crismali (approved):
👍

Issue #1742, opened 2024-11-09
Always show item's status value on staff item screen


jim:
We currently reuse the member-facing code to display item status to staff. This obscures the actual value of `status` for that item and makes more work.

1. Update admin item pages to always show the value of `status`.
2. Also show "checked out", "overdue", etc., but after the other status value. So in some cases, we'll show two things.

![Image](https://github.com/user-attachments/assets/de9a9736-6615-4831-b2d9-ba390945f133)

PR #1792, opened 2024-12-17
Show the raw value of each item's status in addition to the member facing version in the admin interface

crismali:
# What it does

Displays `item.status` on the admin item's index and show pages in addition to the `item_status_label` version.

# Why it is important

#1742 

# UI Change Screenshot

Admin items  index page:
Screenshot 2024-12-17 at 4 33 21 PM

Admin item show page:
Screenshot 2024-12-17 at 4 33 44 PM

# Implementation notes

Just added the capitalized version of the status to both pages. Right now it's background is always gray, but we could add some colors to make it a little nicer if it's not too noisy.


jim (approved):
:shipit:

PR #1795, opened 2024-12-18
Upgrade to Ruby 3.4.0.rc1

jim:
# What it does

Moves us to the latest version of Ruby.

# Why it is important

We're a couple versions behind, so we should upgrade to something more up to date. We might want to wait to merge this until 3.4 is officially out, although I think the risk of running on this release is pretty low for us.

I deleted the `ShortLink` model in this PR. I added it long ago when I thought we might need it to handle text messages, but we have those now and haven't needed it.


crismali (approved):
👍 Looks good

Not sure if we want to merge this until the full version comes out or not but I think we're fine either way.

Issue #1746, opened 2024-11-09
Add tags to maintenance tickets


jim:
1. Allow staff to set one of a set of tags on maintenance tickets. To start with, we'd like to work with:

    * Quick fix
    * Popular Tool
    * Urgent

2. Make it so that tags can be added by staff.

PR #1796, opened 2024-12-18
Adds tagging to tickets on items

crismali:
# What it does

Allows admins to tag tickets and see them on the ticket index and show pages.

# Why it is important

#1746 

# UI Change Screenshot

Tags on the index page:
Screenshot 2024-12-18 at 3 35 29 PM

Tags on the show page:
Screenshot 2024-12-18 at 3 37 17 PM

Tags on the edit page: 
Screenshot 2024-12-18 at 3 37 26 PM

Tags on the new page:
Screenshot 2024-12-18 at 3 37 41 PM

# Implementation notes

I rebased #1795 into this since the tagging gem requires `>= 3.2` Ruby. Once that's merged this diff should be a little easier to understand.

Right now the tags just appear as light gray chips or pills. We could add special styles for special tags (eg 'Urgent' with a red background) but I think this is a good first pass. 

The tags look case sensitive but actually aren't. So if you're the first person to tag something as `Urgent` it will show up as `Urgent` event if you change it to `urgent`. That also goes for any other tickets that are tagged `Urgent`.

In terms of tests, I just added onto some existing controller tests. Maybe system tests for this ticket stuff, but converting them felt like a lot for right now. In the process of troubleshooting the gem setup I added a model test that probably isn't _really_ necessary but that file was empty anyway.


jim (approved):
Looks good to me! I have some thoughts for followup items, but I think we can ship this and then work on those.

> The tags look case sensitive but actually aren't. So if you're the first person to tag something as Urgent it will show up as Urgent event if you change it to urgent. That also goes for any other tickets that are tagged Urgent.

I wonder if we should just force the tags to be lowercase so avoid confusion?

PR #1805, opened 2025-01-02
Update to Ruby 3.4.1

jim:
# What it does

A followup to #1795 that takes us to Ruby 3.4.1 now that it's out.

# Implementation notes

* I tried to update the nix stuff using `nix flake update`, but I think it's specific to Linux based on the `hostPlatform`, so I'm not sure how to verify that it's working. Maybe @phinze can weigh in on the best way to do that? I don't want to leave it broken and I assume the lockfile needs to be updated since we're changing Ruby version.


crismali (approved):
I don't know about the flake stuff, but everything else looks good 👍

phinze:
Ahoy! I can try the nix setup tomorrow and follow up with any changes required. No need to block on it. 🚀

PR #1807, opened 2025-01-02
Look up items by number instead of database ID

### LINKED ISSUE #1777: Reservation item lookup should use number and not ID
When entering IDs in either the checkout or checkout screen for group lending, we should use the item numbers and not the database ID. The numbers will be what is on the barcode that staff will scan with the barcode scanners.

jim:
# What it does

This replaces how we look up reservable items in the group lending code to use the human-friendly item number instead of the database ID.

This fixes #1777.

# Why it is important

We're going to import items from an external system, and we want to be able to change the number without messing with data integrity.

### DIFF
--- a/app/views/admin/reservations/reservation_loans/_reservation_hold.html.erb
+++ b/app/views/admin/reservations/reservation_loans/_reservation_hold.html.erb
@@ -10,10 +10,10 @@
   <% reservation_hold.reservation_loans.each do |reservation_loan| %>
     <tr id="<%= dom_id(reservation_loan) %>">
       <% if reservation_loan.quantity %>
+        <td>*</td>
         <td><%= reservation_hold.item_pool.name %></td>
-        <td><%= reservation_loan.quantity %></td>
       <% else %>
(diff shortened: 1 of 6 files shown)


app/views/admin/reservations/reservation_loans/_reservation_hold.html.erb
```diff
@@ -10,10 +10,10 @@
   <% reservation_hold.reservation_loans.each do |reservation_loan| %>
     <tr id="<%= dom_id(reservation_loan) %>">
       <% if reservation_loan.quantity %>
+        <td>*</td>
         <td><%= reservation_hold.item_pool.name %></td>
-        <td><%= reservation_loan.quantity %></td>
       <% else %>
-        <td><%= reservation_loan.reservable_item.id %></td>
+        <td><%= reservation_loan.reservable_item.number %></td>
```
jim, line 16: This is only somewhat related, but it makes it so the item names always appear in the same column. We don't have numbers for the "uncounted" items that are only represented by an ItemPool yet.

crismali (approved):
Looks great!

Issue #1653, opened 2024-08-25
Activity Report Improvements


jim:
* Break up the activity report by year
* Add renewals as an additional column

PR #1809, opened 2025-01-05
Added a column for renewals count to the activity report

crismali:
# What it does

This adds a renewals column to the activities report and breaks the report into different tables by year.

# Why it is important

#1653

# UI Change Screenshot
failures_test_table_is_populated_accordingly

# Implementation notes

The number of renewals for the month is the number of loans created with an `initial_loan_id` and I changed the `monthly_loans` view to only contain loans without an `initial_loan_id`.


jim (approved):
Thanks for working on this! I think it's a solid improvement.

PR #1817, opened 2025-01-11
Michael configure prettier

crismali:
# What it does

Configures [Prettier](https://prettier.io/)

# Why it is important

We discussed this a bit at the meeting on 1/7, but I couldn't find the issue we made (assuming we made one).

This reformats our JS and CSS/SASS and will keep their respective styles more consistent.

# Implementation notes

I ended up dropping StandardJS since it and Prettier didn't seem to get along as well as they did back in the day (there are some packages that try to reconcile them but that seemed like a bit much). The main differences between Prettier as it's configured here and StandardJS seemed fairly minor. Mostly around the use trailing commas and spaces between parentheses and function names for objects (`{ foo () { ... } }` vs  `{ foo() { ... } }`).  

I think it makes sense to switch to Prettier since it will also do CSS/SASS. There's a lot more CSS than JS in the project and they've never been consistently linted before.

I adjusted CI and Lefthook so that Prettier would run. I'm hoping I didn't miss any other spots.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -175,7 +175,7 @@ jobs:
     #   uses: mxschmitt/action-tmate@v3
     #   if: ${{ failure() }}
 
-  StandardJS:
+  Prettier:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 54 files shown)


jim (approved):
Awesome! It will be a lot nicer to work in the JS and CSS/SASS now.

I agree with your decision to go with Prettier over StandardJS. The only thing worse than no linter is two linters that don't get along!

PR #1822, opened 2025-01-23
Initial MyTurn import and field rearrangement

jim:
# What it does

* Moves some fields from `ItemPool` to `ReservableItem` as discussed at a recent meetup.
* Implements the first pass at a script to import item inventory from a MyTurn export.
* Adds a page that shows the list of `ReservableItem`s in an `ItemPool` so we can see that the import works 😄 

This change was a bit bigger than I was hoping for, so the diff is a bit chaotic.


app/models/concerns/item_categorization.rb
```diff
@@ -1,25 +1,13 @@
 require "active_support/concern"
 
-module ItemAttributes
+module ItemCategorization
```
jim, line 3: The story here is that the category code is reusable in both `Item` and `ItemPool`, but `power_source` isn't now, so I made this file only include the category parts.

app/views/account/item_pools/_items_index.html.erb
```diff
@@ -31,12 +31,6 @@
                 <strong>
                     <%= link_to item_pool.name, account_item_pool_path(item_pool) %>
                 </strong>
-                <% if item_pool.size.present? %>
```
crismali, line 34: You can also write this as `item_pool.size?` if that's your preference

lib/tasks/myturn_import.rake
```diff
@@ -0,0 +1,136 @@
...
+# "Serial Number"
+# "Size"
+# "Source / Supplier"
+# "Status(es)"
+# "Tax(es)"
+# "Weight"
+
+namespace :myturn do
+  task reset: :environment do
+    # Just for development
```
crismali, line 67: Might want to through a `raise unless Rails.env.development?` at the top of this task.

lib/tasks/myturn_import.rake
```diff
@@ -0,0 +1,136 @@
...
+          item_number = begin
+            Integer(row["Item ID"])
+          rescue
+            non_numeric_ids += 1
+            next
+          end
+
+          item_pool = ItemPool.create_with(creator: user).find_or_create_by!(name: row["Name"])
+
+          item_pool.update(
```
crismali, line 96: Might be helpful to make this `item.update!`. Right now this update can fail silently. (although this one isn't caught by the `rescue` below)

lib/tasks/myturn_import.rake
```diff
@@ -0,0 +1,136 @@
...
+            other_names: row["Keywords"]
+          )
+
+          begin
+            item = item_pool.reservable_items.create_with(creator: user).find_or_create_by!(
+              number: item_number,
+              status: ReservableItem.statuses[:active]
+            )
+
+            item.update(
```
crismali, line 107: This seems like a better candidate for `item.update!` since it's in the `rescue`.

crismali (approved):
Looks good. 👍  I had a couple of little notes that are really more like "maybe this would help a little?"

lib/tasks/myturn_import.rake
jim: Good call!

lib/tasks/myturn_import.rake
jim: I made this `update!` and moved it and the statement above into the `begin`/`rescue` block so that the error handling is more consistent.

lib/tasks/myturn_import.rake
jim: Agreed.

Issue #1734, opened 2024-11-05
Search results should not include "in maintenance" items


phinze:
Since we updated maintenance to be a status, it looks like we need to adjust our search query to exclude items in that status.

PR #1823, opened 2025-01-26
basic search enhancement: break ties with item status

djcunningham0:
# What it does

This PR makes a small change to search behavior. If two items have the same pg_search score, ties will now be broken by item status (currently, items with the same score are just sorted by item ID). Specifically, the sort order will be available > on hold > checked out > overdue > in maintenance.

# Why it is important

This seems like a common sense change. It has minimal impact since it's just breaking ties.

# UI Change Screenshot

Currently, it's common for overdue/checked out/etc. items to be ranked above available items in search results:

Untitled

After the change, ties will be broken by item status so available items will be ranked first. (Note that the item status is only being used to break ties for now, so overdue items can still appear at the top of the results if they have a higher pg_search score.)

Untitled 3

# Implementation notes

The logic I used to define the item status mirrors the logic defined in `css_class_and_status_label` in `app/helpers/items_helper.rb`. That appears to be the code used to generate the "Available", "Overdue", etc. labels next to the item names in the search results.

https://github.com/chicago-tool-library/circulate/blob/92f37e6792dc72c478320f7349b1d9f70eda9d7d/app/helpers/items_helper.rb#L162-L180

Some questions and discussion points before we merge this change:
* Is there a better way (performance, maintainability, etc.) of implementing the sorting logic? Any way to reuse existing logic? I had to use a lot of raw SQL, which probably isn't ideal.
  + also, check that my logic looks correct
* Do we want to make additional (larger) changes to the sorting logic? Something beyond breaking ties, like if an item is overdue multiply pg_search score by 0.9. I was thinking we'd merge this change first and then make any larger changes in a subsequent PR, but we could do it here too.
* Do we need to add/update any tests?

### DIFF
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -96,8 +96,8 @@ class ItemsController < ApplicationController
             ended_at IS NULL
             AND (expires_at IS NULL OR expires_at > NOW())
           GROUP BY item_id
-         ) AS active_holds
-         ON active_holds.item_id = items.id"
+         ) AS active_hold_counts
+         ON active_hold_counts.item_id = items.id"
       )
       .joins(
         "LEFT JOIN loans AS active_loans
@@ -121,7 +121,7 @@ class ItemsController < ApplicationController
                     WHEN active_loans.due_at < NOW() THEN 4  -- overdue
                     ELSE 3  -- checked out
                   END
-                WHEN borrow_policies.uniquely_numbered AND active_holds.item_id IS NOT NULL THEN 2  -- on hold
+                WHEN borrow_policies.uniquely_numbered AND active_hold_counts.item_id IS NOT NULL THEN 2  -- on hold
                 ELSE 1  -- available
               END
             WHEN items.status = 'maintenance' THEN 5  -- in maintenance
--- a/app/controllers/items_controller.rb
+++ b/app/controllers/items_controller.rb
@@ -85,51 +85,7 @@ class ItemsController < ApplicationController
       return item_scope
     end
 
-    scope = item_scope.search_by_anything(@query).with_pg_search_rank
-
-    scope
-      .joins(
-        "LEFT JOIN (
-          SELECT item_id, COUNT(*) as active_hold_count
-          FROM holds
-          WHERE
-            ended_at IS NULL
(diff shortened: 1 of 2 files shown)


djcunningham0:
Maybe also worth addressing #1734 here. Should we fully exclude the "in maintenance" items? Or further suppress them?

djcunningham0:
Also linking #1240, the general issue for improving search

app/controllers/items_controller.rb
```diff
@@ -85,8 +85,51 @@ def filter_by_query(item_scope)
...
+                    ELSE 3  -- checked out
+                  END
+                WHEN borrow_policies.uniquely_numbered AND active_hold_counts.item_id IS NOT NULL THEN 2  -- on hold
+                ELSE 1  -- available
+              END
+            WHEN items.status = 'maintenance' THEN 5  -- in maintenance
+            ELSE 6  -- unavailable
+          END AS search_priority
+        "
+      )
```
jim, line 131: > Is there a better way (performance, maintainability, etc.) of implementing the sorting logic? Any way to reuse existing logic?

Usually this would go inside the `Item` model since it's specific to it. Some of this will have to stay as SQL (like the CASE statements at the end), but the rest I think we can translate mostly into Ruby. Maybe we can look at this at the next meetup?

I think the logic in here looks correct, though!

jim (commented):
> Do we want to make additional (larger) changes to the sorting logic? Something beyond breaking ties, like if an item is overdue multiply pg_search score by 0.9. I was thinking we'd merge this change first and then make any larger changes in a subsequent PR, but we could do it here too.

I think I agree, let's merge this and then iterate on it. I do like that idea about demoting overdue items, though.

> Do we need to add/update any tests?

We have these tests that check sort order in `test/system/items_sorting_test.rb`. It looks like they didn't break with this change, but we could add another test case or so to check that this new ordering code continues to work.

If we're going to do a lot of testing, though, we could also write some unit tests in `test/models/item_test.rb`. That would be better if we wanted a lot of them or wanted to just focus on the search query and not how they results are presented in the UI.

> Should we fully exclude the "in maintenance" items? Or further suppress them?

I think for the public page, it makes sense to hide them. We can double check with staff to see if they agree.

jim (approved):

Issue #1793, opened 2024-12-18
As an admin, I hear distinct sounds as I scan items when creating a Pickup


jim:
- computer makes different noises for the types of scan and creates different buckets to be reviewed before checkout (success, disabled item, item not on original reservation)

PR #1827, opened 2025-01-29
Admins hear success/error sounds when adding items to a pickup for a reservation

crismali:
# What it does

Adds sounds when adding/removing items to a pickup for a reservation

# Why it is important

#1793

# Audio Interface Changes

You can test the sounds using the seed data by visiting the [approved reservation's page](http://localhost:3000/admin/reservations/2/items) and clicking on "Start Building".  

- for the "neutral" sound, add an item that is not part of the reservation
- for the "success" sound, add an item that is part of the reservation
- for the "failure" sound, add an item that does not exist
- for the "removed" sound, remove something that you've already added

# Implementation notes

I added a custom turbo stream action called `playSound`. Right now it should only play sounds if the `audio` tags are loaded on the page.

I didn't really add any tests since the general functionality of this part of the app is already tested and the sounds feel like a bonus. System testing this behavior seemed weird (I'm not really sure you can test this sort of thing with them) and so did controller testing controller instance variables or the turbo stream markup.


app/controllers/admin/reservations/pending_reservation_items_controller.rb
```diff
@@ -1,16 +1,15 @@
...
       # Merge into the reservation
       def update
         result = ReservationLending.add_pending_item_to_reservation(@pending_reservation_item)
         if result.success?
-          render_turbo_response(
-            turbo_stream: turbo_stream.action(:redirect,
-              admin_reservation_loans_path(@pending_reservation_item.reservation))
-          )
+          @sound_type = success_sound_path
+          render_turbo_response(:update)
```
crismali, line 12: Adding a file seemed like the easiest way to send down multiple turbo streams.

app/controllers/admin/reservations/reservation_loans_controller.rb
```diff
@@ -75,13 +81,14 @@ def destroy
...
       def render_form_with_error(message)
+        @sound_type = failure_sound_path
         @reservation_loan = ReservationLoan.new
         @reservation_loan.errors.add(:reservable_item_number, message)
         render_form
       end
 
       def render_form
-        render partial: "admin/reservations/reservation_loans/form", locals: {reservation: @reservation, reservation_loan: @reservation_loan}, status: :unprocessable_entity
+        render_turbo_response :create_error, status: :unprocessable_entity
```
crismali, line 91: This `create_error` pattern seemed to be hanging around other parts of the app so I just copied that.

jim (approved):
This is awesome!

I tested this out and it's working really well!

I diid notice that sometimes if you enter new items fast, a sound doesn't get played (I imagine because it's already playing?).

We can maybe call [fastSeek](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/fastSeek) or [load](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/load) on the audio just before calling play? It's not a big deal if a sound gets interrupted, but that might guarantee we get it to play every time.

In any case, I think we can iterate on the details (and get feedback from staff), but this is an exciting step into a bright multimedia future! 🚀 🔈 🎉

PR #1829, opened 2025-02-04
Capture misc metadata from imported items in a column

jim:
# What it does

Adds a new `jsonb` column to `reservable_items` to store arbitrary key/value metadata from imported items. This gives us a way to persist some data that doesn't map cleanly into our schema for future use. One example is that this gives us a way to capture the image URL for each item that we can later fetch and store properly.

### DIFF
--- a/db/migrate/20250202194446_add_myturn_meta_to_reservable_items.rb
+++ b/db/migrate/20250202194446_add_myturn_meta_to_reservable_items.rb
@@ -1,0 +1,5 @@
+class AddMyturnMetaToReservableItems < ActiveRecord::Migration[7.2]
+  def change
+    add_column :reservable_items, :myturn_metadata, :jsonb, default: "{}"
+  end
+end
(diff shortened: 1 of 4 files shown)


db/migrate/20250202194446_add_myturn_meta_to_reservable_items.rb
```diff
@@ -0,0 +1,5 @@
+class AddMyturnMetaToReservableItems < ActiveRecord::Migration[7.2]
+  def change
+    add_column :reservable_items, :myturn_metadata, :jsonb, default: "{}"
```
crismali, line 3: Might want to double check, but does the default of `"{}"` actually wind up as literally a string?

crismali (approved):
👍  Looks good, I just had one question/comment that might be a bit of a "gotcha"

db/migrate/20250202194446_add_myturn_meta_to_reservable_items.rb
jim: You're right, it does end up as a string. Using a plain empty Hash works as intended, though. Good catch!

PR #1831, opened 2025-02-06
Validate that reservation changes will cover existing loans.

jim:
# What it does

Adds a validation to close a gap in how we were validating `ReservationHolds`. Currently, it's possible to reduce the `quantity` of a `ReservationHold` beneath what's would be required by existing `ReservationLoan` objects. So you could have 4 items added to a reservation, but only 3 "reserved" via `ReservationHold` objects. This breaks the lending model's guarantees, which are based around the `ReservationHold` being the source of truth for item availability.

In the last meetup we had a good chat about other ways to handle this situation, which I think are worth exploring as we iterate on the UI. For now, though, I'd just like to maintain data integrity so we don't end up promising more items than we can actually lend out.

Previously, only some `ReservationLoan` objects had a value for `quantity` (those that were for items that aren't uniquely numbered). By always having a value for that field, though, we can simply some of the logic and  do things like `reservation_loans.sum(:quantity)` to count how many items are expected from the loans associated with a `ReservationHold`.

### DIFF
--- a/app/models/reservation_hold.rb
+++ b/app/models/reservation_hold.rb
@@ -8,6 +8,7 @@ class ReservationHold < ApplicationRecord
   validates :ended_at, presence: true
   validates :quantity, numericality: {only_integer: true, greater_than: 0}
   validate :ensure_quantity_is_available
+  validate :ensure_quantity_covers_existing_loans
 
   before_validation :fill_dates_from_reservation, unless: :persisted?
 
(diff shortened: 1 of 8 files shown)


app/models/reservation_hold.rb
```diff
@@ -20,11 +21,7 @@ class ReservationHold < ApplicationRecord
 
   # Total number of items that were reserved
   def loaned_quantity
-    if item_pool.uniquely_numbered?
-      reservation_loans.size
-    else
-      reservation_loans.pluck(:quantity).sum
-    end
+    reservation_loans.sum(:quantity)
```
jim, line 24: This is an example of the simplification we get from always setting `quantity`, even for individual items.

db/schema.rb
```diff
@@ -769,6 +769,7 @@
     t.string "location_shelf"
     t.string "purchase_link"
     t.string "url"
+    t.jsonb "myturn_metadata", default: "{}"
```
crismali, line 772: Looks like the MyTurn stuff from your other PR is leaking into this one

crismali, line 772: Looks like the MyTurn stuff from your other PR is leaking into this one

crismali (approved):
🥇 Looks good to me. I added one note about a line from another PR that seems to have leaked into this one.

db/schema.rb
crismali: Looks like the comment got made twice for some reason 😆

jim: Oh yeah, that is unintentional. Thanks for catching that!

Issue #1756, opened 2024-11-15
Add delete buttons to the item holds list


jim:
On the item holds page, there is no way to remove a hold:

![Image](https://github.com/user-attachments/assets/4320005d-09a1-496a-9d00-db31cbe4dd05)

To delete a hold, you have to click on the member name, then go to their holds list, and find the right hold, and then remove it.

Now that we can disable holds for an individual item, we'll want to be able to more easily clean up holds that aren't needed. Plus, it would be nice to streamline common hold workflows for items where holds are still being used (which is most items).

* Add a small button to each hold that, when clicked, shows a confirmation dialog. If the user accepts the dialog, the hold should be removed.
* We could use a button with a little trash can icon on it to keep the buttons small. Feather icons, the set we use, [has two to choose from](https://feathericons.com/?query=trash).

PR #1835, opened 2025-02-09
1756 add button to remove holds on an item

RudyOnRails:
# What it does

Adds a new button to the admin/item's holds view so that individual holds may be ended. Originally the ticket discussed removing the hold entirely but after some discussion it was decided to set the `ended_at` attribute so our future selves may benefit from the data.

# Why it is important

This would resolve https://github.com/chicago-tool-library/circulate/issues/1756

# UI Change Screenshot

 
### BEFORE
Screen Shot 2025-02-09 at 3 28 17 PM

### AFTER
Screen Shot 2025-02-09 at 3 33 23 PM

# Implementation notes

This is my first PR on this project, and my return to coding after taking a year sabbatical, so just a heads-up I ~~may be~~ am rusty and appreciate and feedback.

One thing to confirm is the new "actions" column (see screenshot). In other apps I've worked on in the past we've used this column for a general one-size-fits-all place to put actions. Let me know if there is another solution we'd like to use.


app/controllers/admin/items/holds_controller.rb
```diff
@@ -1,8 +1,9 @@
 module Admin
   module Items
     class HoldsController < BaseController
+      before_action :set_item
```
jim, line 4: Love to see some refactoring 😄

app/controllers/admin/items/holds_controller.rb
```diff
@@ -11,6 +12,18 @@ def index
             holds_scope.active
           end
       end
+
+      def remove
```
jim, line 16: I could also see this using the `destroy` controller action/HTTP DELETE method. I think it's a little better to use one of the default actions whenever possible, although we definitely have our share of custom ones as well in the app.

jim:
I forgot to mention it in the review, but I think the Actions column is a good pattern to use here.

app/controllers/admin/items/holds_controller.rb
RudyOnRails: That's a good thought! I think originally it was proper if we were going to destroy the hold, but since we decided to just timestamp `ended_at` I felt something else was appropriate. I considered using the `update/patch` combo but for some reason I didn't; can't remember.

Do you prefer something like this that is more generic?
```
      def update
        @hold = @item.holds.find(params[:hold_id])
        @hold.update(hold_params)
        redirect_to admin_item_holds_path(@item), flash: {success: "Hold updated."}, status: :see_other
      end
```

Additionally - when I was working on this I remembered a team I worked on where we discussed creating new controllers for variations of main resources, for example we could have a single concern `EndedHoldsController` where we simply `POST/create`.

Where does your intuition point to?

RudyOnRails:
> I forgot to mention it in the review, but I think the Actions column is a good pattern to use here.

Thanks for the excellent review! Very appreciated.

app/controllers/admin/items/holds_controller.rb
jim: I don't feel too strongly to be honest. I'd like to say we have a strong pattern in the app for this kind of thing, but it's probably a bit ad-hoc since we've had a lot of different people/perspectives in the code over the years.

I guess whether the hold is deleted or not depends on the perspective: in the database we're not deleting it, but from the UI perspective it's gone! But if we're to say that we're going to keep the controllers in line with the the database-view, then I agree that this isn't really deleting the hold.

I actually like the new controller option. Now that you mention it, I did something similar to implement reordering of the holds to avoid overloading the `update` action: https://github.com/chicago-tool-library/circulate/blob/c927219cdfcc64d9589604609334537e0caae24d/app/controllers/admin/items/holds/positions_controller.rb#L4-L18

So I guess that's my preference, but TBH this is also totally fine as-is and there are other things to implement. I'll leave it up to you if you want to change this or merge it as-is and move on to the next thing 😄. I appreciate the thought you put into this!

jim (approved):

jim:
@RudyOnRails For the erb_lint failure, you can probably run `bundle exec erb_lint --autocorrect .` and it'll fix any of those style issues for you.

phinze:
Hi @RudyOnRails - looks like that that [system test failure](https://github.com/chicago-tool-library/circulate/actions/runs/13949224684/job/39057978619?pr=1835) is real and repros on my machine. I took a stab at fixing it here: https://github.com/chicago-tool-library/circulate/commit/07c1c7c4a055c90c7822b0de4184a0c6f1651936

Take a look and snag that commit if you like it! Once the build is passing I can get this merged and deployed 👍🏻

RudyOnRails:
> Hi @RudyOnRails - looks like that [system test failure](https://github.com/chicago-tool-library/circulate/actions/runs/13949224684/job/39057978619?pr=1835) is real and repros on my machine. I took a stab at fixing it here: [07c1c7c](https://github.com/chicago-tool-library/circulate/commit/07c1c7c4a055c90c7822b0de4184a0c6f1651936)
> 
> Take a look and snag that commit if you like it! Once the build is passing I can get this merged and deployed 👍🏻

Oof, sorry I just saw this right after I merged.

Issue #1804, opened 2025-01-02
Maintenance ticket tag improvements


jim:
Now that we have tags on tickets, it would be nice to add some additional features:

* A way to filter tickets by tag on the Items in Maintenance screen

PR #1853, opened 2025-02-19
Show maintenance tags

renwah:
# What it does
- Addresses #1804: Adds the ticket tags to the "items in maintenance" report as well as a filter for the tags. This currently allows you to filter on one tag at a time--if we need to filter by multiple tags at once, we'll need to update this! 

# Why it is important

Makes tags more helpful!

# UI Change Screenshot

Screenshot 2025-02-20 at 12 27 31 PM

# Implementation notes

There is a strange bug where if you refresh the page occasionally, it'll prepopulate old params into the filter fields (it will not filter the results). This is annoying, but I thought it would not come up that often. I think it has to do with the fact that I had to switch to a non-ransack form helper because it was difficult to include the tags in the ransack search (it should be compatible, but it that part of Ransack seems stale--Ransack recently required that you have to allowlist everything you search by explicitly, which I could not do easily with tags).

### DIFF
--- a/app/views/admin/reports/items_in_maintenance/index.html.erb
+++ b/app/views/admin/reports/items_in_maintenance/index.html.erb
@@ -4,35 +4,39 @@
 
 <section class="filters-container">
   <h2>Filters</h2>
-  <%= search_form_for(@q, url: admin_reports_items_in_maintenance_index_path, class: "filters-form") do |f| %>
+  <%= form_with(url: admin_reports_items_in_maintenance_index_path, class: "filters-form", method: :get) do |f| %>
     <div class="filters-form-fields">
       <div class="filter-column item-and-ticket-filters">
         <div class="filter-row item-filters">
           <div class="filter-field name">
-            <%= f.label :item_name_cont, "Item Name" %>
-            <%= f.search_field :item_name_cont %>
+            <%= f.label :"q[item_name_cont]", "Item Name" %>
+            <%= f.search_field :"q[item_name_cont]"  %>
           </div>
 
           <div class="filter-field status">
-            <%= f.label :item_status_eq, "Item Status" %>
-            <%= f.select :item_status_eq, Item.statuses.values.sort.map { |value| [value.humanize, value] }, {include_blank: true} %>
+            <%= f.label :"[q]item_status_eq", "Item Status" %>
+            <%= f.select :"[q]item_status_eq", Item.statuses.values.sort.map { |value| [value.humanize, value] }, {include_blank: true} %>
           </div>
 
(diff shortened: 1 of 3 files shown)


jim (commented):
Thanks for working on this! It will be really nice to have this functionality.

It's too bad about the ransack form helper not cooperating. I wonder if it's possible to use the ransack form helper for most of the form and only build the tags select manually? We could even use `select_form_tag`, etc. and avoid the form object entirely. Feel free to say that won't work for whatever reason, I'm far from an expert with ransack myself 😄.

Otherwise, I figure that's worth a shot to have fewer form elements generated the hard/verbose way. But if it doesn't work, we can merge this and start using it.

I don't think the form redisplay issue is too terrible. Maybe it's a browser thing that can be disabled by setting `autofill`?

renwah:
Ok, I found a way to keep using the ransack form helper (`search_form_for`) along with the tag filter--unfortunately, I still couldn't get it to work entirely within Ransack but it's pretty close! I think this looks a lot cleaner, and it also fixes the issue with reloading the page I was seeing earlier 🎉 I did notice that it doesn't work unless the `selected: param[...]` option is added to the tag select field (see my comment), which I think must be a quirk of the interaction between the regular form field + Ransack. 
Note: I think this will require a merge commit before merging since it's been open for awhile? But not sure about the branch management in this project yet!

app/views/admin/reports/items_in_maintenance/index.html.erb
```diff
@@ -34,6 +34,10 @@
             <%= f.label :status_eq, "Ticket Status" %>
             <%= f.select :status_eq, [["Assess", "assess"], ["Parts", "parts"], ["Repairing", "repairing"]], {include_blank: true} %>
           </div>
+          <div class="filter-field tag">
+            <%= f.label :tag_eq, "Tag" %>
+            <%= f.select :tag_eq, ActsAsTaggableOn::Tag.most_used.pluck(:name), {include_blank: true, selected: params[:q][:tag_eq]} %>
```
renwah, line 39: `selected: params[:q][:tag_eq]` -- this is required for some reason. When it's removed, the page fails and returns this error: 

> NoMethodError: undefined method 'tag_eq' for an instance of Ransack::Search

jim (approved):
Awesome! Thanks for working through the quirks in this library. I think this is a good result and we should go ahead and get it merged in!

In terms of how we handle branches on the project, we squash PRs when merging into a single commit. So rebases/meges/etc. are kind of up to the PR's author, since it all ends up in a single commit in the end 😄.

Issue #1848, opened 2025-02-17
Track when appointments have been collected


jim:
> Adding a button/checkbox on the appointment page to note when the tools have been collected/the order has been fulfilled by a librarian. This is a librarian-facing feature to help them tell if they missed any orders.

1. Replace the "complete" button on the appointments index with a button that says "pull items".
2. Once this button is clicked, replace it with the current "complete" button. We'l probably want an "mark as not pulled" button somewhere as well so that librarians can undo recording that an appointment has been collected.
3. If an appointment has been collected, then show the current complete button.

Staff questions:

* We dim appointments that have been completed. Should we maybe use different colors/decoration based on the appointment state so they are easy to differentiate? What would be most useful? **YES**

This is a precursor to #1849.

PR #1860, opened 2025-02-24
Admins can mark appointments as pulled in addition to pending and complete

crismali:
# What it does

When looking at the appointments index, admins will now see three groups of appointments: pending, pulled, and completed. They can move appointments between these three states and the page will update accordingly.

They can also make these same changes on the appointment details/show page.

# Why it is important

#1848

# UI Change Screenshot

Mobile version of the appointments index:
Screenshot 2025-02-23 at 5 15 05 PM

Desktop version of the appointments index (the version before the feature flag):
Screenshot 2025-02-23 at 5 15 38 PM

Desktop version of the appointments index (the version after the feature flag):
Screenshot 2025-02-23 at 5 15 21 PM

Show page for a pending appointment:
Screenshot 2025-02-23 at 5 59 28 PM

Show page for a pulled appointment:
Screenshot 2025-02-23 at 5 58 06 PM

Show page for a completed appointment:
Screenshot 2025-02-23 at 5 58 18 PM

# Implementation notes

I added a `pulled_at` column to the `appointment` since completion as a concept is represented by a `completed_at` column. There is probably some sort of refactoring that could be done around representing this state, but I think there's a larger refactor that might be good.

There's a more than fair amount of duplication floating around here. Specifically, there are mobile and desktop versions of appointment partials, new and original versions of the index page and partials, and a few controllers that I duped for this. 

Before refactoring or drying anything up I think we need to find out more about how librarians are using these pages. Do they always use the old or the new version of the index page? Is there a better way to represent the index page that's more useful and doesn't require desktop and mobile specific markup?

That said, I think the verbosity of this is probably fine for the near term and I'd rather be sure we dry things up the right way then guess and end up painting ourselves into a corner. Plus, I think this will definitely make things better for the librarians so I'm happy to put any refactoring off to get this to them sooner.

### DIFF
--- a/app/controllers/admin/appointment_detail_completions_controller.rb
+++ b/app/controllers/admin/appointment_detail_completions_controller.rb
@@ -8,7 +8,7 @@ module Admin
       @appointment = Appointment.find(params[:appointment_id])
       @appointment.update!(completed_at: Time.current, staff_updating: true)
 
-      redirect_to admin_appointments_path, flash: {success: "Appointment completed."}, status: :see_other
+      render_appointment_to_turbo_stream
     end
 
     def destroy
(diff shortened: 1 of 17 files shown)


jim (approved):
I agree that we should probably clean up some of the duplication in here, but in the meantime, we can get this change out.

PR #1873, opened 2025-03-10
Update erb_lint config

jim:
# What it does

Addresses some deprecation warnings from `erb_lint`:

```
Calling `erblint` is deprecated, please call the renamed executable `erb_lint` instead.
The config file has been renamed to `.erb_lint.yml` and `.erb-lint.yml` is deprecated. Please rename your config file to `.erb_lint.yml`.
```

This PR makes the changes suggested above and also runs `bundle exec erb_lint --autocorrect .` to fix any outstanding linting errors.


phinze (approved):
LGTM! 🧼 🧹

PR #1874, opened 2025-03-10
Wrap user emails in our standard email layout

jim:
# What it does

This wraps the devise emails (email confirmation/password reset/etc) with the same layout we use for other emails. This has two benefits:

1. These emails now look more official to recipients
2. The emails are less minimal, which might have been a factor is our devise emails being flagged as "suspicious" by gmail.

# UI Change Screenshot

Here's an example of how devise emails now look:

Screenshot 2025-03-10 at 4 38 58 PM

# Implementation notes

* I also took this opportunity to setup mailer previews for the devise emails, so we can preview them along with our other emails.

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -1,9 +1,38 @@
 class UserMailer < Devise::Mailer
   default from: "Chicago Tool Library <team@chicagotoollibrary.org>"
+  layout "mailer"
+
   after_action :store_notification
+
+  def initialize_from_record(record)
+    super # load user
+    @library = @user.library
+    @logo = :small
+  end
+
+  def confirmation_instructions(*args)
+    @title = "Please confirm your email"
+    super
+  end
+
+  def email_changed(*args)
+    @title = "Your email has changed"
+    super
+  end
+
+  def password_change(*args)
+    @title = "Your password has changed"
+    super
+  end
+
+  def reset_password_instructions(*args)
+    @title = "Reset your password"
+    super
+  end
 
   def store_notification
     Notification.create!(
+      library: @library,
       action: action_name,
       address: mail.to.join(","),
       member: @user.member,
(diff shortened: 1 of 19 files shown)


app/mailers/user_mailer.rb
```diff
@@ -1,9 +1,38 @@
...
   default from: "Chicago Tool Library <team@chicagotoollibrary.org>"
+  layout "mailer"
+
   after_action :store_notification
 
+  def initialize_from_record(record)
+    super # load user
+    @library = @user.library
+    @logo = :small
+  end
```
jim, line 11: Overriding this method feels a little gross, but due to the way these mailers are written I couldn't find a better way to do this.

app/views/devise/mailer/unlock_instructions.mjml
```diff
@@ -0,0 +1,13 @@
...
+      <p>Hello <%= @resource.email %>!</p>
+
+      <p>Your account has been locked due to an excessive number of unsuccessful sign in attempts.</p>
+
+      <p>Click the link below to unlock your account:</p>
+
+      <p><%= link_to "Unlock my account", unlock_url(@resource, unlock_token: @token) %></p>
+    </mj-text>
+  </mj-column>
+</mj-section>
```
jim, line 13: We don't currently lock folks out, but we've had this module enabled for a while, so I went ahead and updated this email as well.

app/mailers/user_mailer.rb
phinze: I starting writing something about code or tests that try to verify that the upstream interface hasn't changed, but then I found [it hasn't changed in 14 years](https://github.com/heartcombo/devise/blame/fec67f98f26fcd9a79072e4581b1bd40d0c7fa1d/lib/devise/mailers/helpers.rb#L22) so I think it's probably fine to just let our mailer previews and the existing unit tests cover this. 🙂

phinze (approved):
Nice! Hopefully this helps with some of our deliverability gremlins

Issue #1814, opened 2025-01-08
Upgrade to Rails 8


jim:
[Rails 8 is out!](https://rubyonrails.org/2024/12/13/Rails-Version-8-0-1-has-been-released).

PR #1879, opened 2025-03-14
Upgrade to Rails 8

crismali:
# What it does

Upgrades Rails' Ruby and JavaScript dependencies to 8.0!

# Why it is important

#1814 and just general keeping up with the times.

# Implementation notes

Check out the specific comments I've left


Gemfile
```diff
@@ -4,11 +4,11 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }
 ruby "3.4.1"
 
 # Bundle edge Rails instead: gem "rails", github: "rails/rails"
-gem "rails", "~> 7.2.2"
+gem "rails", "~> 8.0.1"
 
 # Since Rails 7 sprockets is optional; we still use it so we need to depend on
 # the gem explicitly
-gem "sprockets-rails"
+gem "sprockets-rails", require: "sprockets/railtie"
```
crismali, line 11: This seems like the preferred way to do this these days 🤷

app/controllers/errors_controller.rb
```diff
@@ -16,6 +16,11 @@ def error_content
         title: "Page not found",
         message: "The page you were looking for could not be found."
       }
+    when 406
```
crismali, line 19: Rails generates html for the 406 status error page so I deleted that and added support for 406s into our existing error handling.

config/environments/development.rb
```diff
@@ -78,18 +77,26 @@
...
       ip = IPAddr.new(addrinfo.ip_address).mask(24)
-
       Logger.new($stdout).info "Adding #{ip.inspect} to config.web_console.permissions"
 
       config.web_console.permissions = ip
     end
   end
 
+  # Annotate rendered view with file names.
+  config.action_view.annotate_rendered_view_with_filenames = true
```
crismali, line 87: This adds html comments to the markup that let you know that this segment of html is partial A and this other segment is from the layout.

config/environments/production.rb
```diff
@@ -6,74 +6,91 @@
...
   # Specifies the header that your server uses for sending files.
-  # config.action_dispatch.x_sendfile_header = 'X-Sendfile' # for Apache
-  # config.action_dispatch.x_sendfile_header = 'X-Accel-Redirect' # for NGINX
+  # config.action_dispatch.x_sendfile_header = "X-Sendfile" # for Apache
+  # config.action_dispatch.x_sendfile_header = "X-Accel-Redirect" # for NGINX
 
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :amazon
 
-  config.active_storage.variable_content_types << "image/webp"
```
crismali, line 46: I can't find it now, but somewhere along the way Rails and ActiveStorage started working better with `webp` so this shouldn't be necessary anymore.

db/seeds.rb
```diff
@@ -1,3 +1,7 @@
+ActiveSupport.on_load(:action_mailer) do
+  Rails.application.reload_routes_unless_loaded
+end
```
crismali, line 3: Even though emails don't actually get sent from the seeds, the mailers blow up if the routes aren't loaded.

jim (approved):
This is great! I clicked around locally and everything seems to be working just fine.

Issue #1594, opened 2024-07-19
Special Handling for High Demand / High Value Items ("C tools")


phinze:
Original request from @tessavierk:

> We think we might actually be ready for "C" tools to be used. We may transition high-value or high-demand tools to "C" status and not let certain users reserve them. Ideally this would be an access we could toggle on/off for users. New users would automatically not be able to borrow "C" tools until they've had a few successful loans and users with frequent late returns might lose "C" status.

---

"High demand" items are tools that are popular and therefore have long lines of people waiting to use them. For these items, late pickups and returns have negative impact on more people and can make for extraordinarily high wait times.

"High value" items are tools that are expensive to replace or repair. If these tools are damaged or go missing, it's a bigger deal for us to repair or replace them.

Some of these items are also "high-touch", meaning they involve extra effort for us to loan out/process when returned.

To help improve throughput for loans on high demand item and reduce risk on high value items, we'd like to add a borrow policy type that limits who can borrow items with that policy. The "C - Special" policy was originally intended for this, but no code has yet been written to handle it.

Requirements to support C tools:

* Librarians will change certain high-value or high-demand items to the C borrow policy
* Members can't borrow or reserve C items by default
* "C" tool access has to be requested by a member and can be revoked by librarians at any time
* A maximum of 1 "C" tool can be added to an appointment (we'll handle this in #1850)

Staff questions:
* How do we communicate these limitations to members? It would be helpful to have a written policy or update the help test to explain this new system.
    * We'll notify existing members via email or something once this is ready
    * Maybe add a box to the side of the item page that summarizes the rules for the item?
* What's involved in requesting access to C items? Is it just a button or do we ask additional questions?
    * We might want to only allow folks to request access after e.g. n loans or an amount of time.
* Do we want any other limits on C items (limit on the number of current holds?)

New behavior to add:

* Ability to have a borrow policy that can be toggled on/off per-member manually
    * This can be a new boolean on `borrow_policies` called `requires_approval`.
* Changes to the hold buttons on items page:
    * If the member has approval for that borrow policy, we'll continue to show the normal hold UI
    * If the member doesn't have approval, there will be a button to request approval. The hold button will be hidden.
* There needs to be an approval request flow for members, and a request approve/deny flow for staff. We can model this on the renewal request flows already in the app.
    * We'll need a new model (maybe `BorrowPolicyApproval` to join members to these borrow policies and hold other metadata.
[...]

PR #1887, opened 2025-03-19
Support C tools

crismali:
# What it does

Allows borrow policies to be marked as requiring approval. If an item has a borrow policy that requires approval, members will have to request approval for that kind of item. Later, librarians/admins can approve/reject/revoke that approval.

# Why it is important

#1594 this one has been in the making for a bit

# UI Change Screenshot

Brace yourself, for they are legion.

## Members

When an item belongs to a borrow policy that requires approval, members will see this on the item show page:
item show no approval record

Once they've requested approval they'll see:
item show requested approval

If they end up approved:
item show approved

And if they end up rejected or have their approval revoked:
item show rejected

### Emails

Approved HTML email:
approval email html

Approved text email:
approval email text

Rejected HTML email:
rejected email html

Rejected text email:
rejected email text

Revoked HTML email:
revoked email html

Revoked text email:
revoked email text

## Librarians

Updated the borrow policies index:
borrow policies index

There was a seemingly unused borrow policy show page (you'd need to know the url to access it) has been updated:
borrow policy show

The borrow policy edit form now has the requires approval checkbox:
borrow policy edit

The "Manage Approvals" button takes librarians to the borrow policy approvals index for that borrow policy (by default they're sorted by `created_at` descending):
borrow policy approvals index no filters

The borrow policy approvals index can be filtered/paginated:
borrow policy approvals index with filters

Borrow policy approvals can be edited:
borrow policy approval edit

Updating a borrow policy approval takes you back to the borrow policy approvals index page:
borrow policy approvals index with flash message

# Implementation notes

The original card mentions a `revoked_at` and `revocation_reason`, but so far my implementation doesn't have the `revoked_at` and has a more generic `status_reason`. 

I think the `revoked_at` fell away because I rolled all of the statuses of the approval into a single enum. In place of the `revoked_at` field maybe we'd just want to add audited to the approvals? Then a librarian could see the whole history if there has been a bunch of back and forth on the approval.

I went with the more generic `status_reason` because I figured there could be reasonable notes for just about any status. Examples: "approved but probationary", "rejected because they just became a member", or "rejected because they've broken too many B tools". 

Speaking of, the `status_reason` shows up (if present) in the rejected and revoked emails.

If a librarian updates the the borrow policy approval without changing the status, no new email will be sent out.

Let me know what ya'll think! Nothing is too small or large to change!

**Note:** in the dev data, an example of a C tool is the item with the id `10893`

### DIFF
--- a/app/assets/stylesheets/reports.scss
+++ b/app/assets/stylesheets/reports.scss
@@ -1,4 +1,5 @@
-.items_in_maintenance {
+.items_in_maintenance,
+.borrow_policy_approvals {
   --border-color: rgb(206.3684210526, 195.2236842105, 188.1315789474);
   --box-shadow: 0 0 0 0.1rem rgba(47, 120, 144, 0.2);
   --blue: #2f7890;
--- a/app/controllers/account/borrow_policy_approvals_controller.rb
+++ b/app/controllers/account/borrow_policy_approvals_controller.rb
@@ -1,0 +1,14 @@
+module Account
+  class BorrowPolicyApprovalsController < BaseController
+    def create
+      borrow_policy = BorrowPolicy.find(params[:borrow_policy_id])
+
+      BorrowPolicyApproval.find_or_create_by!(
+        borrow_policy:,
+        member: current_user.member
+      )
+
+      redirect_to item_path(params[:item_id]), status: :see_other, success: "Approval requested."
+    end
+  end
+end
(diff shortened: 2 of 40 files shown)


jim (approved):
Amazing! Really nice work.

I agree with your logic around the decision to use `status_reason`.

The main thing we should get feedback on is some of the language. For example, I think that maybe the "You cannot borrow this type of item" message on the item page is a little vague. Maybe we could have a page to link to that explains the rules or something?

In any case, let's just get staff to weight in on wording. But I think we can merge this and then tweak that stuff later.

PR #1888, opened 2025-03-19
Update Nix setup

phinze:
# What it does

Spruces up the nix development environment that I use. The required change is `nix flake update` so the latest ruby version is picked up, the other changes (direnv-nix update, flake.nix formatting) are for good measure.

# Why it is important

Only important for people using Nix dev setup, which at this point is just me. :)

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -4,10 +4,9 @@
 #
 # For details on the Nix development setup, see NIX.md.
 #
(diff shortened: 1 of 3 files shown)


phinze:
Self-merging because I'm the only one using this

there are dozens of us, dozens!

Issue #1889, opened 2025-03-20
Members receive a renewal reminder email before the banner appears in the app


crismali:
**Describe the bug**
A member got an email to renew their membership but instead of clicking the renewal link in the email, they seem to have just signed into the app and tried to navigate to the renewal page. I think the renewal emails are getting sent a day before the renewal button/banner appears in the app which can be a little confusing.

**To Reproduce**
1. Have a membership that expires 30 days from now
2. Sign in with that member
3. You won't see the "Renew Membership" button and banner

**Expected behavior**
If someone has received the email, the UI should be showing the button and banner.

**Additional context**

`lib/tasks/email.rake:36` looks for memberships with `Membership.where(ended_at: (Date.current + 30.days).all_day)`, but our banner decides whether or not to render with `if current_member.last_membership && (current_member.last_membership.ended? || current_member.last_membership.ends_within?(29.days.since))` (in `app/views/account/_membership_renewal_message.html.erb:1`)

PR #1890, opened 2025-03-21
Ensure that if the renewal email has been sent, the renewal banner is also showing in app

crismali:
# What it does

Fixes a bug where a user would receive their renewal notice email and then visit the site and be unable to see the "Renew Membership" banner (they could click the link in the email but didn't)

# Why it is important

#1889 A member ran into this issue the other day

# Implementation notes

The memberships factory was steamrolling the `ended_at` so I tweaked it to prevent the steamrolling, but that led to me having to tweak the dates in a membership validation test.


jim (approved):
A sneaky bug! Thanks for fixing this.

Issue #1851, opened 2025-02-17
Track accessories on items


jim:
> Can we add sub-items (accessories) that are automatically reserved with an item by members, but require librarians to check each one in/out?
Ex. sewing machine is a single item reserved by a member. Librarians, however, have to check out the sewing machine, the power cord, and accessory kit.
We would need the ability to add or remove accessories from item entries in case an accessory is lost or added.
This would help a lot to ensure important parts of tools aren't going missing or being forgotten on the shelves during check-out.

Example tools we need this for:
* table saw (fence, guards, push stick, etc.)
* sewing machine (presser foot, power cord, accessory kit, etc.)

Staff questions:
* Do the accessories need to be numbered, or is just a checkbox enough? Do we need images of each accessory?
    * Just a list of names of accessories to start.

Implementation notes:
* Add a textarea to the item edit form. Each line in the field will be treated as an accessory (split on newlines)
* In the lending UI, display a checkbox for each accessory. Disable the lend button. When the checkboxes are all checked, enable the button.
* Do the same thing for checking in the item.
* Also, update the appointment page with the same UI.

PR #1895, opened 2025-03-29
Track item accessories

crismali:
# What it does

Allows librarians to add/edit/remove accessories from items and requires librarians to check off those accessories when lending or accepting a return.

# Why it is important

#1851 Helpful reminders for multipart items.

# UI Change Screenshot

Editing an item's accessories:
edit accessories

Librarian viewing an item's accessories:
view accessories

Checking out an item when none of the accessories are checked:
checking in all unchecked

Checking out an item when some of the accessories are checked:
checking in half checked

Checking out an item when all of the accessories are checked:
checking in all checked

After checking out an item with accessories:
checked out

Returning an item when the accessories are not checked:
returning unchecked

Returning an item when the accessories are checked:
returning checked

After returning an item:
returned

# Implementation notes

I added the functionality when checking in/out items through an appointment, but the original issue says to add this to the lending UI and the appointment page. Am I missing something here?

### DIFF
--- a/app/javascript/controllers/index.js
+++ b/app/javascript/controllers/index.js
@@ -14,6 +14,8 @@ import CollapseController from './collapse_controller'
 
 import ConditionalFieldController from './conditional_field_controller'
 
+import ConfirmItemAccessoriesController from './confirm_item_accessories_controller'
+
 import DynamicFieldsController from './dynamic_fields_controller'
 
 import EmailSettingsEditorController from './email_settings_editor_controller'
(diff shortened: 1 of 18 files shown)


app/javascript/controllers/index.js
```diff
@@ -46,6 +48,10 @@ application.register('appointment-date', AppointmentDateController)
 application.register('autocomplete', AutocompleteController)
 application.register('collapse', CollapseController)
 application.register('conditional-field', ConditionalFieldController)
+application.register(
+  'confirm-item-accessories',
+  ConfirmItemAccessoriesController
+)
```
phinze, line 54: I tried moving this back on its own line just cause it stands out and learned it's the linter that makes it go to multi-line. Silly linter. 😂

phinze (approved):
Code looks good and I took it for a spin locally and it all seems to work well!

> I added the functionality when checking in/out items through an appointment, but the original issue says to add this to the lending UI and the appointment page. Am I missing something here?

I'm not positive, but I think the "Lending UI" is referring to the way librarians can directly lend an item to a member from their page:

2025-04-20 at 11 35 07@2x

Similarly, there's a "Return" button on the member page for loaned items. I imagine this is how Librarians lend and return things for folks who show up without an appointment. I imagine the accessories reminders would be just as useful here as when doing the appointment flow.

Approving since I think this is fine to merge as-is if you like. The nice thing is that the functionality is fully opt-in since it'll only show up when somebody adds accessories to an item.

crismali:
Thanks @phinze!

PR #1900, opened 2025-04-04
Add membership_type field to member/membership exports

jim:
# What it does

Adds a new field for each year in the members/membership exported CSV. This can be inferred from the other columns, but having an explicit column and value makes importing into other systems easier.

# UI Change Screenshot

image


crismali (approved):
👍 👍 👍 👍 👍 👍 👍 👍 👍 👍

PR #1905, opened 2025-04-07
Use Google Calendar events to determine loan due dates

phinze:
# What it does

For loan and renewal due date due date calculation, we've historically used an algorithm that was based on a hard coded list of weekdays that the library was open.

This changes to use Google Calendar synced appointment slots to determine the next open day.

# Why it is important

Now we don't need to update a static list when the schedule changes. As long as the linked Google Calendar events are correct, the next open day will be selected as the due date.

# Implementation notes

After spending some time with versions of the code that raised an error if no subsequent open day is found, I ended up deciding it was simpler to have the implementation return the passed-in day and log an error instead of blowing up. This means the tests don't have to change significantly, but we'll still get notified if the Google Calendar events run out and we're not properly calculating due dates.


crismali (approved):
Just in _time_ 🕚 👍

PR #1907, opened 2025-04-08
Deleted .standard_todo.yml and added comments to disable the lines in the migrations that were causing warnings

crismali:
# What it does

Adds inline comments to take care of standard rb lint errors in the migrations. Because we can drop the todo file we no longer get output about it when running standard.

# Why it is important

Reduces a little noise.

# Implementation notes

It doesn't make any of the migrations any more reversible than they were. I think the only "substantial" change was changing an `each` to a `find_each`.


phinze (approved):
Yay less noise! 🤫

PR #1906, opened 2025-04-08
Fix flakey unpulled items system tests

crismali:
# What it does

# Why it is important

# UI Change Screenshot

# Implementation notes

PR #1909, opened 2025-04-10
Prevent borrow policy approval errors in devdata rake task

crismali:
# What it does

Fixes more things up in devdata.rake similar to #1906. I must've gotten lucky running bin/setup when testing that PR.

# Why it is important

Quality of life for everyone

# Implementation notes

Just spreading that scope around.


phinze (approved):
🧼 👍

PR #1920, opened 2025-04-29
1737: Add pinned notes to member profiles

fgonza23:
# What it does

Implements support for pinning notes to the top of a member’s profile. Pinned notes are shown in a high-visibility section above the Notes tab, with all notes (pinned and unpinned) accessible within the tabbed table view.

# Why it is important

- Highlights critical member information at the top of the profile
- Adds a checkbox to allow pinning when creating or editing a note
- Organizes all member notes in a clean, structured tabbed layout for easier access

# UI Change Screenshot

Screenshot 2025-04-29 at 5 32 47 PM

# Implementation notes

Updated the NotesController to support both Turbo Stream and full HTML fallback responses.
Created Turbo Stream templates to dynamically update the pinned notes section after creating or deleting a note.

- Known limitation: the Notes tab does not currently auto-refresh after note creation/deletion; requires manual page refresh.

### DIFF
--- a/app/controllers/admin/members/base_controller.rb
+++ b/app/controllers/admin/members/base_controller.rb
@@ -13,6 +13,7 @@ module Admin
 
       def load_member
         @member = Member.find(params[:member_id])
+        @notes = @member.notes.pinned_first
       end
     end
   end
--- a/app/controllers/admin/members/notes_controller.rb
+++ b/app/controllers/admin/members/notes_controller.rb
@@ -21,6 +21,9 @@ module Admin
         load_note
 
         if @note.update(note_params)
+          if @note.pinned?
+            @member.notes.where.not(id: @note.id).update_all(pinned: false)
+          end
           redirect_to [:admin, @member, anchor: dom_id(@note)], status: :see_other
         else
           render :edit, status: :unprocessable_entity
@@ -47,7 +50,7 @@ module Admin
       private
 
       def note_params
-        params.require(:note).permit(:body)
+        params.require(:note).permit(:body, :pinned)
       end
 
       def load_note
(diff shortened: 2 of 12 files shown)


fgonza23:
Hey @jim -Pinned notes now appear at the top of the member profile, and all notes are listed in the Notes tab.

I also began integrating Turbo Streams to update the UI dynamically — the pinned notes section updates correctly, and the new note form resets without a full page reload. However, the Notes tab section doesn’t yet update automatically and still requires a manual refresh. That part appears to need a more targeted Turbo Stream or frame setup.

phinze (changes requested):
Thanks for your work on this @fgonza23! This looks like it's going in the right direction 🚀 

I have a couple of smaller comments inline, but there's one main issue I see testing this out, which is the update screen doesn't seem to redirect or show a flash message when editing a note. Also I can reproduce the thing you noted which is after creating a note it doesn't show up until after a refresh. I think this is all quirky behavior coming from the turbo streams pieces.

I wonder if it might make sense to implement this first without turbo streams and then follow up adding that as an enhancement?

phinze:
I also enabled the test suite to run which shows a few test failures. The linter issues can be fixed automatically by [getting the linters set up in your development environment](https://github.com/chicago-tool-library/circulate/blob/main/README.md#code-formatting-and-linting). And the [controller test issue](https://github.com/chicago-tool-library/circulate/actions/runs/14742709101/job/41563454001?pr=1920#step:16:48) looks like it should reproduce locally. Let me know if you have any issues or questions digging into these!

fgonza23:
Thanks for the detailed review, @phinze, I appreciate the clear guidance!

fgonza23:
Thanks for the review, @phinze . I added a flash message to the update redirect and removed the unused `get :all` route. The controller test now explicitly requests HTML to avoid the 204 from the Turbo Stream response.

The Notes tab still requires a manual refresh after creating or deleting a note. I left the Turbo Stream blocks in place for now—let me know if you'd prefer I remove them entirely in this PR.

Issue #1857, opened 2025-02-20
Drop items.description column


jim:
We don't actually use this column because we use `has_rich_text :description` in `Item`, which stores that content in a different table with the other active text content. So we can drop this column to avoid confusion.

PR #1922, opened 2025-04-30
Drop the description column from the items table

crismali:
# What it does

It removes the `description` column from `Item` since we use a rich text description for it instead.

# Why it is important

#1857 Kind of confusing to use `item.description` and to not be sure if you're getting the column or the rich text.

# Implementation notes

Tried to keep it easy-breezy.


phinze (approved):
Looks both easy and breezy

PR #1923, opened 2025-05-02
Phinze/dust off nix on mac

phinze:
# What it does

Makes the Nix development setup work on Darwin machines again

# Why it is important

Not very important! Just allows me to hop between Linux and Mac machines. Also we had one other person try out the Nix setup and note that it was failing on Chromium issues

# Implementation Notes

The only non-nix-specific code change here is to eager load vips for the image_processing gem. I think this should be safe, and I'll make sure it's working ok on staging before rolling it out.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -5,8 +5,7 @@
 # For details on the Nix development setup, see NIX.md.
 #
 ## shellcheck shell=bash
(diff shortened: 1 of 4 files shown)


phinze:
Self-merging a la https://github.com/chicago-tool-library/circulate/pull/1888#issuecomment-2738075672

Issue #874, opened 2021-11-23
Add place in member portal to view policies


jim:
**Is your feature request related to a problem? Please describe.**
There is currently no way for members to view the policies they agree to when they sign up for a membership.

**Describe the solution you'd like**
Add a page (or more, as needed) to the member portal that display the information that is shown when uses sign up.

We'll likely need to wait until #889 is done so that there is a place to put this info. We could add a menu item called "Agreements" to the Account menu once there is one.

lundersaur:
Hi @jim I am trying to start looking into this. One thing I am struggling with as a new contributor is understanding how I can associate a Member with a Document through an AgreementAcceptance. I cannot seem to determine how I would know what Document a Member has agreed to -- the AgreementAcceptance has the `member_id`, but seemingly nothing that reference what was agreed to. Can you help me understand what the expected flow is?

Relatedly, does this feature need to account for versions of Documents? E.g. a Document is published stating that pickups are not available when the Cubs are World Series Winners as a joke and everyone signs up pre 2016, but then the Cubs win and the Document is revised to just say "No pickups on Sundays" but everyone who registered prior agreed to the original Document.

jim:
@lundersaur Good questions. The agreement that is being agreed to is currently whatever we show on the screen during signup, which is [`Document.agreement`](https://github.com/chicago-tool-library/circulate/blob/main/app/controllers/signup/documents_controller.rb#L12) which [ends up being the `Document` with a `code` of `agreement`](https://github.com/chicago-tool-library/circulate/blob/f7c6bafc5d51a0c89c9b534d6601bd9ab3f37cb4/app/models/document.rb#L18).

I think having versions of Documents (or at least the agreement one) is probably a good idea. But we could get started here by just adding pages that let us view the text from `Document.agreement` somewhere in the member profile. Next steps would probably be to also add a page to view the text from `Document.rules` and then probably some way to version documents.

I don't think tracking the version of the rules that a person sees is all that important (although we can confirm with staff). My understanding is that the agreement is basically a legal agreement about liability, whereas the rules are not as much a formal agreement and more likely to change over time. I think we'd probably always want members to see the latest version of the rules if they go to view them.

PR #1938, opened 2025-05-29
Refactor Document model to use 'enum' for 'code'

lundersaur:
# What it does

Replaces some existing logic around the `code` field to use the Rails `enum` helper. This gives extra functionality in terms of scopes and value checkers.

# Why it is important

Rather than needing individual methods added for each Document `code` type, `enum` will give us the methods for free.

# UI Change Screenshot

n/a

# Implementation notes

This is not a requested feature, but was something I encountered working on https://github.com/chicago-tool-library/circulate/issues/874 and so thought I would raise for review.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -9,8 +9,6 @@ class Document < ApplicationRecord
     borrow_policy: "borrow_policy",
   }, validate: true
 
-  scope :coded, ->(code) { where(code: code) }
-
   acts_as_tenant :library
 
   def to_param
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -6,7 +6,7 @@ class Document < ApplicationRecord
   enum :code, {
     agreement: "agreement",
     code_of_conduct: "code_of_conduct",
-    borrow_policy: "borrow_policy",
+    borrow_policy: "borrow_policy"
   }, validate: true
 
   acts_as_tenant :library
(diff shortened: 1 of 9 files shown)


app/models/document.rb
```diff
@@ -3,19 +3,17 @@ class Document < ApplicationRecord
 
   validates :name, :summary, presence: true
 
+  enum :code, {
+    agreement: "agreement",
+    code_of_conduct: "code_of_conduct",
+    borrow_policy: "borrow_policy",
+  }, validate: true
+
   scope :coded, ->(code) { where(code: code) }
```
jim, line 12: I wonder if we can get rid of this scope now as well?

jim (approved):
A nice refactor!

app/models/document.rb
lundersaur: I think so -- didn't do it originally since adding the `enum` wasn't specifically requested, but if you're okay with switching I'll remove the `coded` `scope` as well.

lundersaur: Huh, looks like there is one spot left. Not entirely sure what it's doing from first glance, will need to look a little more.

From https://github.com/chicago-tool-library/circulate/blob/main/app/controllers/admin/documents_controller.rb#L38:
```
def set_document
  @document = Document.coded(params[:id]).first!
end
```

Maybe can approach it with https://github.com/chicago-tool-library/circulate/blob/main/app/controllers/documents_controller.rb#L3 ?

PR #1942, opened 2025-06-05
1737: Add pinned notes to member profiles

### LINKED ISSUE #1737: Member notes can be pinned
Here's the feature request:

> We would also love a secondary notes area on member pages. We'd like to keep the existing, high-visibility area for notes (#1 in photo) but also a more hidden area (#2 in photo) maybe near the other member tabs to have more sensitive info about members if-needed

Image

My proposed solution is:

> For the notes, since there can already be multiple notes added to a member, what if we had a way to "pin" a note and that would cause it to appear at the top (in position #1). Other notes could be shown in a tab (position #2).

To do this, we'll need to:

* add a new `pin` field to notes
* update the forms to allow the new field to be edited
* add a new notes tab to the member page to display all notes
* only show pinned notes at the top of the member page

fgonza23:
## What it does

Implements support for pinning notes to the top of a member’s profile, based on the original feature request in #1737.

- Adds a "Pinned notes" section at the top of the profile 
- Adds a "Notes" tab to display all notes
- Adds a `pinned` field to notes

## Why it is important

- Allows staff to highlight critical member information in a high-visibility location
- Organizes all member notes

## Implementation notes

- Uses Turbo Streams

## Linked issue

Closes #1737

### DIFF
--- a/app/controllers/admin/members/base_controller.rb
+++ b/app/controllers/admin/members/base_controller.rb
@@ -13,6 +13,7 @@ module Admin
 
       def load_member
         @member = Member.find(params[:member_id])
+        @notes = @member.notes.pinned_first
       end
     end
   end
--- a/db/migrate/20250423201733_add_pinned_to_notes.rb
+++ b/db/migrate/20250423201733_add_pinned_to_notes.rb
@@ -1,0 +1,5 @@
+class AddPinnedToNotes < ActiveRecord::Migration[8.0]
+  def change
+    add_column :notes, :pinned, :boolean
+  end
+end
--- a/app/controllers/admin/members/notes_controller.rb
+++ b/app/controllers/admin/members/notes_controller.rb
@@ -21,6 +21,9 @@ module Admin
         load_note
 
         if @note.update(note_params)
+          if @note.pinned?
+            @member.notes.where.not(id: @note.id).update_all(pinned: false)
+          end
           redirect_to [:admin, @member, anchor: dom_id(@note)], status: :see_other
         else
           render :edit, status: :unprocessable_entity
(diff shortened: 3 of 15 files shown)


fgonza23:
Hey @phinze  I’ve redone this PR using a clean branch. 
The feature now matches the original issue 1737

db/migrate/20250423201733_add_pinned_to_notes.rb
```diff
@@ -0,0 +1,5 @@
+class AddPinnedToNotes < ActiveRecord::Migration[8.0]
+  def change
+    add_column :notes, :pinned, :boolean
```
crismali, line 3:
```suggestion
    add_column :notes, :pinned, :boolean, default: false, null: false
```

app/controllers/admin/members/base_controller.rb
```diff
@@ -13,6 +13,7 @@ def are_appointments_enabled?
 
       def load_member
         @member = Member.find(params[:member_id])
+        @notes = @member.notes.pinned_first
```
crismali, line 16: As discussed this should probably be in a `load_notes` method

phinze: Would look something like this:
```ruby
before_action :load_notes
def load_notes
  @notes = @member.notes.pinned_first
end
```

fgonza23: I believe I made that change in my latest commit. just wanted to confirm if it usually updates here automatically?

crismali: @fgonza23 Looks like this change didn't make it through. Could you give it another try?

PR #1946, opened 2025-06-11
Added includes to appointment check ins/outs to speed things up a bit

crismali:
# What it does

Checking items out and returning them has felt a little sluggish lately so this adds some eager loading to help speed things up.

# Why it is important

I believe this was mentioned at the last stakeholders meeting

# Implementation notes

The real change is just the addition of some `includes` calls to eager load associations. I also changed the names of some block variables in the templates because the original names were throwing me off a little.

### DIFF
--- a/app/controllers/admin/appointments_controller.rb
+++ b/app/controllers/admin/appointments_controller.rb
@@ -77,19 +77,19 @@ module Admin
     end
 
     helper_method def items_available_to_add_to_pickup
-      Item.available.eager_load(:borrow_policy)
+      Item.available.includes(:borrow_policy)
     end
 
     helper_method def items_available_to_add_to_dropoff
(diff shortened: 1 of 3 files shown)


app/controllers/admin/appointments_controller.rb
```diff
@@ -77,19 +77,19 @@ def load_appointment
     end
 
     helper_method def items_available_to_add_to_pickup
-      Item.available.eager_load(:borrow_policy)
+      Item.available.includes(:borrow_policy)
```
phinze, line 80: I've got a separate PR trying to address that this returns nearly the whole item collection to render into that dropdown, which I'm suspicious is also contributing to page slowness.

crismali: Definitely suspicious

PR #1947, opened 2025-06-12
Add ability for member to view Documents agreed to

### LINKED ISSUE #874: Add place in member portal to view policies

lundersaur:
# What it does

Resolves #874: feat: Add ability for a member to view the most recent Document `'agreement'`, `'code_of_conduct'`, and `'borrow_policy'`.

# Why it is important

There is currently no way for members to view the policies they agree to when they sign up for a membership.

# UI Change Screenshot

Screenshot 2025-06-11 at 21 28 24

# Implementation notes

There is no way to tie a particular Document to a Member agreeing, so we're unable to provide a member with when they agreed to a particular document (or a version of a document) for now, but that would be a good enhancement in the future.

The view/controller method are very similar currently, but opted to leave them distinct so that iteration on each version can be done without affecting other versions.

### DIFF
--- a/app/controllers/account/members_controller.rb
+++ b/app/controllers/account/members_controller.rb
@@ -4,6 +4,18 @@ module Account
       @member = current_member
     end
 
+    def agreement
+      @document = Document.where(code: 'agreement').last
+    end
+
+    def code_of_conduct
+      @document = Document.where(code: 'code_of_conduct').last
+    end
+    
+    def borrow_policy
+      @document = Document.where(code: 'borrow_policy').last
+    end
+
     def edit
     end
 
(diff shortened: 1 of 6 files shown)


jim (approved):
This code looks good!

We should probably check with staff to verify that the documents are titled in a good way for members to see. For example, I'm not sure that "borrow policy" is the right term, it might need to be something like "borrowing rules" or something.

jim:
@lundersaur I missed that there are some linting errors. Running `bundle exec standardrb --fix` should take care of those 😄.

Issue #1840, opened 2025-02-12
Maintenance ticket tagging UI


jim:
It would be nice to add a way to add/remove tags without typing them. We could probably use the same autocomplete we use for item categories? Or if there's a better approach, that's OK as well. We could also show a list of the most commonly used ones underneath the field, and when folks click on a tag, we add/remove it from the text input? Definitely open to ideas here of course!

PR #1952, opened 2025-06-24
Add tags select and tag creation to maintenance tickets

renwah:
# What it does

Uses the `tags_editor_controller` we use for Item categories to select tags for maintenance tickets. I also changed the tags_editor_controller slightly to use the "create" setting from selectize as well. 

# Why it is important

ref: #1840 

It's nice to be able to reuse tags easily to be more consistent!

# UI Change Screenshot
Current form:
Screenshot 2025-06-24 at 2 22 40 PM

New form:

https://github.com/user-attachments/assets/a3462aff-0c30-458c-b411-fa7ce4dd53c1

# Implementation notes

I think the tag_list_select name is a little clunky/open to suggestions about how to distinguish between the tag_select method that's used for item categories!

### DIFF
--- a/app/controllers/admin/items/tickets_controller.rb
+++ b/app/controllers/admin/items/tickets_controller.rb
@@ -57,7 +57,7 @@ module Admin
       end
 
       def ticket_params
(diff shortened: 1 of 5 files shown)


crismali (approved):
Looks great!

Issue #1934, opened 2025-05-17
When updating an item to retired, appointments for holds are not cancelled


phinze:
**Describe the bug**

We had a situation where:

* A member placed a hold on an item
* A few days later, that item was discovered to be missing and marked as retired
* The member still had a pick up appointment for the item

**Screenshots**

Circulate developers can view the example of this happening on [this slack thread](https://chicagotoollibrary.slack.com/archives/C05RB4B5M1N/p1747491248435269).

**Other notes**

We have [code that is meant to cancel holds on items when they're retired](https://github.com/chicago-tool-library/circulate/blob/main/app/models/item.rb#L218-L222), but I wonder if it's not handling appointments correctly.

lundersaur:
Hi @phinze, I am hoping to start contributing to Circulate but do not have access to the Slack currently -- would you be able to help me with access?

PR #1955, opened 2025-07-02
Cancel appointments for holds that are ended when an item is unavailable

renwah:
# What it does

- In the Hold model, adds a method to remove a hold from an appointment, cancelling the appointment if there are no other items
- In the Item model, updates `clear_holds_if_inactive` to trigger holds on the item to be removed from an appointment if it exists. 

# Why it is important
Addresses #1934 
# UI Change Screenshot

# Implementation notes

- Thinking about this, it might make sense to add a notification here, but wasn't sure if we wanted to avoid notifications for some reason!
- The way I implemented this ended up interacting badly with consumable items, so that if you were the librarian checking out the last of a consumable item, you'd see an error screen once the item was "loaned", retiring the item once it was consumed (the scenario in the consumable_concern_test I edited). I just fudged the test a little by adding another item to avoid this situation in the test, since looking at the description for the consumable borrow policy, we haven't had consumable items since the seed share a few years ago. If we're still using consumable items, I'm happy to look more into how to reconcile these issues!

### DIFF
--- a/app/models/hold.rb
+++ b/app/models/hold.rb
@@ -53,7 +53,7 @@ class Hold < ApplicationRecord
     ended? || expired?(now)
   end
 
-  # A hold that was picked up
+  # A hold that was picked up or a hold on an item that was retired
   def ended?
     ended_at.present?
   end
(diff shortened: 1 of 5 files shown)


crismali (approved):
Looks good! Regarding the edge case for consumable items, we're going to talk about that at the next stakeholders meeting but that's more to figure out the future of marking items as consumable.

renwah:
Oh good to know about the consumable items! Thanks for reviewing.

PR #1961, opened 2025-07-22
allow searching by different number formats

### LINKED ISSUE #1949: allow searching by "B-####"
**Describe the bug**
Suppose there's an item with number B-1234. You can find that item by searching "1234". However, if you search "B-1234" or "B1234" you won't get any results. It would be a better user experience all of those searches returned the B-1234 item.

I don't know exactly where in the codebase this should be implemented, but it's probably not too difficult. Just keep in mind that we have "B" and "C" prefixes and there can be three or four digits in the number (and maybe five digits in the future).

Related: if you look at the [zero search terms query](https://app.chicagotoollibrary.org/blazer/queries/47-zero-search-results?n=200) in Blazer, you'll see some "B-####" searches. The most common is "b-3663" with 9 searches. Pretty low volume, indicating this not a huge issue.

I suspect we're getting these searches because there are signs in the library that say something like "type the number on the tool into the search bar".

djcunningham0:
# What it does

Fixes #1949. Currently, to find item B-1234 you would have to search "1234". "B-1234" or "B1234" would return no results. Now, any of those options will return tool B-1234.

# Why it is important

It's not that important, just a tiny user experience improvement. We have physical signs at the library saying you can search by number, but it isn't clear that you should exclude the letter so we occasionally get searches for "B-1234".

# Implementation notes

I just added one line to preprocess the query in the `search_and_order_by_availability` scope, but I don't know if that's the best approach. I just used regex to match the "X-####" or "X####" patterns.

**Potential problems/weirdness:**
* If there are any search terms that follow the pattern "X###" (one letter followed by some numbers) that match tool names or descriptions, this change will negatively impact those tools. I can't think of any examples but I haven't looked into the search logs.
* This solution doesn't require you to use the *right* letter. You could search "A-1234", "B-1234", or "Z-1234", and all of those would return tool B-1234. This is probably fine.

### DIFF
--- a/app/models/item.rb
+++ b/app/models/item.rb
@@ -76,7 +76,9 @@ class Item < ApplicationRecord
   scope :by_name, -> { order(name: :asc) }
 
   scope :search_and_order_by_availability, ->(query) {
-    item_scope = search_by_anything(query)
+    # allow searching for all number formats, e.g., "B-1234" -> "1234", or "B1234" -> "1234"
+    processed_query = if query.match(/^[A-Z]-?(\d+)$/i) then $1 else query end
+    item_scope = search_by_anything(processed_query)
       .with_pg_search_rank
       .with(active_hold_counts: Hold.active.group(:item_id).select(:item_id, "COUNT(*) as active_hold_count"))
       .joins(


jim (approved):
This seems like a nice improvement, and it's scoped narrowly enough by the regex that I don't think it could cause many issues.

PR #1970, opened 2025-07-28
Upgraded Ruby to 3.4.5

crismali:
# What it does

Bumps Ruby up a little bit

# Why it is important

[Good to keep things current, security updates, etc.](https://github.com/ruby/ruby/releases/tag/v3_4_5)

# Implementation notes

Really did the least


jim (approved):
I think probably the nix stuff is also out of date. We could maybe have someone make a script that we run after changing Ruby versions that could update it? Don't need to block on that though, may as well get this in.

PR #1985, opened 2025-08-16
Added borrow policy approval statuses to the member profile page and added the ability to request approval

crismali:
# What it does

Adds the approval statuses for all borrow policies that require approval to the member's profile page and provides a button for them to request approval if they haven't previously.

# Why it is important

#1984 

# UI Change Screenshot

When no approval request exists:
Screenshot 2025-08-16 at 1 18 13 PM

Requested:
Screenshot 2025-08-16 at 1 14 54 PM

Revoked:
Screenshot 2025-08-16 at 1 16 26 PM

Rejected:
Screenshot 2025-08-16 at 1 16 07 PM

Approved:
Screenshot 2025-08-16 at 1 15 45 PM

# Implementation notes

I accidentally committed and pushed this straight to `main`, reverted that, and now this is reverting the revert commit 😅 

I removed a couple of assertions around where requesting approval redirects to since now it will just redirect to the referrer (or root if there is not referrer).

### DIFF
--- a/test/system/member_profile_test.rb
+++ b/test/system/member_profile_test.rb
@@ -7,11 +7,55 @@ class MemberProfileTest < ApplicationSystemTestCase
     login_as @user
   end
 
+  def borrow_policy_code_name(borrow_policy)
+    "#{borrow_policy.code}-Tools"
+  end
+
   test "member can view profile" do
     visit account_member_url
 
     assert_content @member.full_name
     assert_content @member.number
+  end
+
+  test "a member can see their approval statuses for borrow policies that require approval" do
+    no_approval_borrow_policy = create(:borrow_policy, requires_approval: false)
+    requires_approval_borrow_policy = create(:borrow_policy, :requires_approval)
+
+    approved = create(:borrow_policy_approval, :approved, member: @member)
+    rejected = create(:borrow_policy_approval, :rejected, member: @member)
+    requested = create(:borrow_policy_approval, :requested, member: @member)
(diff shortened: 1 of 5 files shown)


test/system/member_profile_test.rb
```diff
@@ -7,13 +7,57 @@ def setup
...
+    requested = create(:borrow_policy_approval, :requested, member: @member)
+    revoked = create(:borrow_policy_approval, :revoked, member: @member)
+
+    visit account_member_url
+
+    refute_content borrow_policy_code_name(no_approval_borrow_policy)
+    assert_content(/#{borrow_policy_code_name(approved.borrow_policy)}:\s+Approved/)
+    assert_content(/#{borrow_policy_code_name(rejected.borrow_policy)}:\s+Rejected/)
+    assert_content(/#{borrow_policy_code_name(requested.borrow_policy)}:\s+Requested/)
+    assert_content(/#{borrow_policy_code_name(revoked.borrow_policy)}:\s+Revoked/)
```
phinze, line 36: "Revoked" seems fine for the backend, but I think we might want something gentler for the UI text... maybe "Not Approved"?

phinze: Not blocking though, so approving the PR overall! ✅

phinze (approved):

test/system/member_profile_test.rb
phinze: Matching mailers would be:

* `rejected` - "Not approved"
* `revoked` - "No longer approved"

PR #1989, opened 2025-08-18
A member that has been rejected, had their access to C-Tools revoked or is new can make a new request after 2 months

crismali:
# What it does

Prevents people who have been members for less than two months from requesting approval for C-Tools and allows members who have had their approval rejected or revoked more than two months ago request approval again.

# Why it is important

#1980 and #1981

# UI Change Screenshot

Nothing should be visually different.

# Implementation notes

### DIFF
--- a/app/views/account/members/show.html.erb
+++ b/app/views/account/members/show.html.erb
@@ -84,6 +84,7 @@
       </dd>
 
       <% @borrow_policies_and_approvals.each do |borrow_policy, approval| %>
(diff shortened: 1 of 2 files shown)


jim (approved):
I think pulling the logic for this into an object makes sense. We should probably find a different name for this thing since “borrow policy” is so overloaded already. I’m trying to think of something better, but so far haven’t been able to.  We don’t need to block on this.

PR #1990, opened 2025-08-19
Added a small message letting users know they have been approved for C-Tools

crismali:
# What it does

Puts a small message about being approved for C-Tools on the homepage if they were approved in the last two weeks.

# Why it is important

#1979

# UI Change Screenshot

Screenshot 2025-08-19 at 3 15 36 PM

# Implementation notes

Once #1983 lands it might make sense to make the message link to the filtered list of tools.

### DIFF
--- a/app/controllers/account/home_controller.rb
+++ b/app/controllers/account/home_controller.rb
@@ -5,6 +5,7 @@ module Account
       @loans = current_member.loans.checked_out
       @ready_holds_count = current_member.holds.active.started.count
       @waiting_holds_count = current_member.holds.active.waiting.count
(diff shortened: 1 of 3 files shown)


app/controllers/account/home_controller.rb
```diff
@@ -5,6 +5,7 @@ def index
       @loans = current_member.loans.checked_out
       @ready_holds_count = current_member.holds.active.started.count
       @waiting_holds_count = current_member.holds.active.waiting.count
+      @recent_approvals = current_member.borrow_policy_approvals.approved.where(updated_at: 2.weeks.ago..)
```
phinze, line 8: Since we have plenty of room on this page, I say we just leave the approval in place all the time rather than scoping to only recent approvals

phinze (approved):

Issue #1999, opened 2025-09-14
Librarians should be able to grant access to C-Tools even when the member hasn't requested it


crismali:
> Correct me if I'm wrong, but I don't believe there is currently a way to grant a member C-Tool access without them requesting it. This would be nice to do! For example, a very reliable new member wants to borrow a C-Tool, but I can't override the 2month probation period.

PR #2000, opened 2025-09-14
Librarians should be able to grant access to C-Tools even when the member hasn't requested it

crismali:
# What it does

Adds a button to the admin member show page that will approve a member to borrow tools under a borrow policy immediately.

# Why it is important

Requested by Tessa recently, captured in [#1999](https://github.com/chicago-tool-library/circulate/issues/1999)

# UI Change Screenshot

When a member hasn't requested approval:
Screenshot 2025-09-14 at 4 20 04 PM

After approving the member:
Screenshot 2025-09-14 at 4 20 24 PM

# Implementation notes

It also sends an email to let the member know they've been approved


phinze (approved):
Nice!

Issue #2015, opened 2025-10-08
Add field to item for retire reason


jim:
> Hey all! Curious if we can add some options to tool status drop-down. Currently it's:
> * Pending
> * Active
> * Maintenance
> * Missing
> * Retired
> Could we add:
> "Retired - Not Returned"
> "Retired - Broken"
> "Retired - Upgraded" (edited) 

I'm thinking we add a second field, `retire_reason`, and only show/set it when the item's status is "retired". That keeps any logic that looks for items with a `status` of `retired` unchanged, but still adds context that we can use in reports, etc.

[Original request](https://chicagotoollibrary.slack.com/archives/C05RB4B5M1N/p1759344700285449)

PR #2030, opened 2025-10-25
Set retired_reason when creating Tickets or TicketUpdates

jim:
# What it does

Extends the work from #2015 to allow for an Item's `retire_reason` to be set from Ticket or TicketUpdate forms.

# Why it is important

This allows a staff member to accurately record why an item was retired without having to remember to complete additional steps.

# UI Change Screenshot

Screenshot 2025-10-25 at 12 05 31 PM
Screenshot 2025-10-25 at 12 04 08 PM

# Implementation Notes

I added some basic system tests for maintenance tickets since we didn't have any yet.

### DIFF
--- a/app/controllers/admin/items/tickets_controller.rb
+++ b/app/controllers/admin/items/tickets_controller.rb
@@ -21,7 +21,7 @@ module Admin
       end
 
       def create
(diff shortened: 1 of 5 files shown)


crismali (approved):
🔥

PR #2034, opened 2025-10-29
Replace datalist with server-backed autocomplete for checkout

phinze:
# What it does

Replaces the HTML5 datalist on the appointment checkout page with a server-backed autocomplete.

The "Add an item to the appointment for check-out" form now fetches matching items on-demand via AJAX as the user types, instead of rendering all ~5800 items into the DOM on page load.

# Why it is important

The datalist approach was loading thousands of `<option>` elements on every appointment page load, causing performance issues. The new autocomplete only loads ~20 items per search query.

# UI Change Screenshot

N/A - reproduces existing behavior with better performance

# Implementation notes

- Follows the existing autocomplete pattern used elsewhere (item names, brands, sizes)
- New endpoint: `GET /admin/ui/available_items?q=search_term`
- Uses `Item.search_by_anything()` to match user input
- Returns up to 20 results, properly scoped to current library tenant
- Uses existing `SpectreFormBuilder#autocomplete_text_field` helper
- Stimulus autocomplete controller + Awesomplete library handle the UI
- Moved the "Add an item" form outside the `.panel-body` container to prevent the Awesomplete dropdown from being clipped by the `overflow-y: auto` CSS property. The form is now wrapped in a `.panel-body-no-overflow` div that maintains consistent padding without the overflow constraint.

### DIFF
--- a/app/assets/stylesheets/overrides.scss
+++ b/app/assets/stylesheets/overrides.scss
@@ -123,6 +123,11 @@ trix-toolbar {
   z-index: 10;
 }
 
+// Panel body styling without overflow constraint for autocomplete dropdowns
+.panel-body-no-overflow {
+  padding: 0 variables.$layout-spacing-lg;
+}
+
 .off-canvas {
   height: auto;
   .off-canvas-content {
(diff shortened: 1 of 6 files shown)


jim (approved):
>  instead of rendering all ~5800 items into the DOM on page load

It's kind of wild that we got away with this for as long as we did.

I'm happy to see that the autocomplete stuff we had laying around mostly just worked out of the box.

PR #2067, opened 2026-01-03
use blend of tsearch and trigram

djcunningham0:
# What it does

This PR makes some changes to our `pg_search` configuration which should (slightly) improve the quality of results for lots of queries. 

Specifically, I added `trigram` search and blended the results with the current full-text search (`tsearch`) results. Trigram searching (docs [here](https://www.postgresql.org/docs/current/pgtrgm.html) and [here](https://github.com/Casecommons/pg_search?tab=readme-ov-file#trigram-trigram-search) calculates word "similarity" based on overlapping trigrams (3-letter sequences).

Empirically, adding trigram search alongside tsearch has the following impact:
- Blended tsearch + trigram results are more relevant than either tsearch or trigram on their own.
- Trigram can work with minor typos or misspellings, e.g., "taable".

# Why it is important

Search quality is a known problem. Addresses #1240

# UI Change Screenshot

Below are some screenshots of some common queries with different results after the change. Usually the new results are (subjectively) better.

(Note: screenshots are from the app running locally, where I didn't have item images and modified the HTML of the search page)

**table saw** (small improvement)
image

**router** (better first result)

image

**digging** (recent example from the Slack channel; top three results are better under new settings)

image

**cooler** (all top results are better)

image

**miter saw** (***mixed;*** first 2 results are better, but the rest are probably worse)

image

# Implementation notes

### evaluation

- In addition to subjectively evaluating of some queries, I also attempted to evaluate objectively using our search log data. I found that these new `pg_search` settings improved our overall search relevance metrics across many queries.
  - **TODO:** link to analysis or add documentation

- In addition to adding trigram search, I explored changing the weights (A, B, C, D) we're using in the tsearch config. **I found that the current weights are (nearly) optimal.** (Some other combinations performed *slightly* better on the objective metrics, but it's highly dependent on which queries you evaluate on and how you infer the relevance of results.)

- I tested the [other tsearch settings](https://github.com/Casecommons/pg_search?tab=readme-ov-file#tsearch-full-text-search) and found that **our current settings are optimal.** Specifically:
  - We should keep `prefix=True`. Setting `prefix=False` degrades the quality of a few very common search terms (such as "drill" and "saw").
  - All options for `normalization` appeared to degrade our results (but I didn't explore this one thoroughly)
  - Setting `any_word=True` significantly degraded the results for many queries

### trigram configuration

- I configured the trigram search to only look at the `name` attribute. I found this gave better, more consistent results than using trigram for all of the same fields as tsearch.

- I went back and forth between setting `sort_only` to true or false. Ultimately I decided false is probably better, but we could revisit in the future.
  - `sort_only=True`: Same results as current behavior (only the items retrieved by tsearch), but reranked using the tsearch + trigram scores
  - **`sort_only=False`**: Pull in additional results if their trigram similarity is greater than `threshold`. **This allows it to work with minor misspellings** that would be missed by tsearch (e.g., "taable"). The downside is it may add irrelevant results to the end for some queries (if you decrease `threshold` it'll add more results)

### DIFF
--- a/test/system/items_sorting_test.rb
+++ b/test/system/items_sorting_test.rb
@@ -155,7 +155,7 @@ class ItemSortingTest < ApplicationSystemTestCase
 
     assert_equal 3, item_links.size
 
-    expected_item_names_order = [@drill1, @drill2, @drill3].map(&:name)
+    expected_item_names_order = [@drill2, @drill1, @drill3].map(&:name)
 
     assert_equal expected_item_names_order, item_links
   end


jim:
@djcunningham0 Sounds good. I think we can create a migration that sets up the extension if it doesn't already exist.

PR #2069, opened 2026-01-06
Enforce User-Member relationship by auto-creating Users when admin creates Members

phinze:
# What it does

- When an admin creates a Member, a User account is now automatically created and linked
- The new User is pre-confirmed (admin has verified them) and receives a password reset email
- Removes the ability to create Users directly from the admin UI (removes "New User" button and create action)
- Admins can still edit existing Users to change roles (promote to staff/admin)

# Why it is important

Fixes a bug where users could end up with a User account but no associated Member record, causing 500 errors when they tried to access the member portal.

**Root cause:** Admin::UsersController allowed creating Users with `role: member` without any associated Member record. When these users logged in and were redirected to `/account`, the app crashed trying to call `.loans` on a nil `current_member`.

This enforces a new invariant: **every User has a Member**. Users are now just authentication records; Members are the primary entity.

# UI Change Screenshot

The "New User" button is removed from Settings > Users. Admins should create Members instead (which auto-creates the User).

# Implementation notes

- `Admin::MembersController#create` now wraps Member + User creation in a transaction
- Uses `skip_confirmation!` since admin-created users are pre-verified
- Sends password reset email so the member can set their own password
- Removed `new` and `create` actions from `Admin::UsersController`
- 12 other orphaned users with this issue exist in production (to be cleaned up separately)

### DIFF
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -38,6 +38,7 @@ class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
   end
 
   teardown do
+    Warden.test_reset!
     ActsAsTenant.test_tenant = nil
     ActionController::Base.allow_forgery_protection = false
   end


jim (approved):
Nice fix. TBH I had forgotten there was even a way to create users manually.

PR #2171, opened 2026-05-16
Stop chasing CSRF noise on Devise login and signup

phinze:
# What it does

Skips CSRF on `Users::SessionsController#create` and `Signup::MembersController#create`. The application_controller rescue and AppSignal reporting stay in place for every other form.

# Why it is important

These two endpoints have been generating most of our `InvalidAuthenticityToken` noise for a long time: about 1,300 on the sessions controller and 30 on signup. The `NonSessionCookieStore` round helped a lot but didn't fully stop the bleed.

I sampled the still-failing requests today, and the leftover noise is almost entirely things CSRF can't defend against. Bots POSTing scraped login forms, Safari users whose csrf_token cookie got pruned by ITP, and a few folks who left a tab open too long. Not a single real CSRF attack in the bunch.

Login is public, login-CSRF is low-impact for a library app, and Devise rotates the session on sign-in. Signup is public too and `#create` doesn't read `current_user`, so spam prevention belongs in rate-limiting, not CSRF.

# UI Change Screenshot

No UI changes. Failed attempts now show the usual Devise/form validation errors instead of the "There was an issue with your submission, please try again." retry flash.

# Implementation notes

The `Users::SessionsController#create` override is a `super` passthrough so `Rails/LexicallyScopedActionFilter` doesn't complain about a `skip_before_action` targeting an inherited action.

The CSRF rescue test moves off the sessions controller (since the rescue can't fire there anymore) and exercises the rescue via `Users::ConfirmationsController` instead.

### DIFF
--- a/app/controllers/signup/members_controller.rb
+++ b/app/controllers/signup/members_controller.rb
@@ -1,6 +1,11 @@
 module Signup
   class MembersController < BaseController
     before_action :is_membership_enabled?
+    # Skip CSRF to stop InvalidAuthenticityToken errors that hurt real users
+    # (long signup form, lost data on retry). Signup is public and #create
+    # doesn't read current_user, so spam prevention belongs in rate-limiting,
+    # not CSRF. See application_controller.rb for prior chapters.
+    skip_before_action :verify_authenticity_token, only: :create
 
     def new
       if session[:member_id]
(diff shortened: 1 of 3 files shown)


jim (approved):
Rationale makes sense to me. It will nice to not have this creating noise anymore!

PR #2173, opened 2026-05-17
Use bookmark icons for the save for later functionality

crismali:
# What it does

Replaces the old save for later buttons' text with styled bookmark icons and html `title` attributes for information on hover.

# Why it is important

Makes things a little easier on the eyes and more compact.

# UI Change Screenshot

Items index with saved and not saved items:
Screenshot 2026-05-17 at 3 58 17 PM

Saved for items index:
Screenshot 2026-05-17 at 3 58 29 PM

Saved for later item show:
Screenshot 2026-05-17 at 3 58 04 PM

Not saved for later item show:
Screenshot 2026-05-17 at 3 57 58 PM

## New flash messages

Unsaving an item on the index page:
Screenshot 2026-05-30 at 3 35 12 PM

Saving an item on the index page:
Screenshot 2026-05-30 at 3 35 26 PM

Removing an item from the saved for later page:
Screenshot 2026-05-30 at 3 35 51 PM

# Implementation notes

I added an svg as a partial which is possibly a little weird, but it made it much easier to style. Also, I had to add the svg in the first place because the feather icon for bookmark as Circulate is set up now is some tricky css with borders.

### DIFF
--- a/app/views/items/_for_later_list_item_show.html.erb
+++ b/app/views/items/_for_later_list_item_show.html.erb
@@ -1,7 +1,7 @@
 <% if FeatureFlags.for_later_lists_enabled? %>
   <span id="for_later_list_item_show">
     <% if for_later_list_item.present? %>
-      <%= button_to "Don't need it", account_for_later_list_item_path(for_later_list_item),
+      <%= button_to "Remove from saved for later", account_for_later_list_item_path(for_later_list_item),
             method: :delete, class: "btn" %>
     <% else %>
       <%= button_to "Save for later", account_for_later_list_items_path,
--- a/app/views/items/_for_later_list_item_show.html.erb
+++ b/app/views/items/_for_later_list_item_show.html.erb
@@ -1,11 +1,17 @@
 <% if FeatureFlags.for_later_lists_enabled? %>
   <span id="for_later_list_item_show">
     <% if for_later_list_item.present? %>
-      <%= button_to "Remove from saved for later", account_for_later_list_item_path(for_later_list_item),
-            method: :delete, class: "btn" %>
+      <%= button_to account_for_later_list_item_path(for_later_list_item),
+            method: :delete, class: "btn remove-from-save-for-later-button", title: "Remove from save for later list" do %>
+          <%= render partial: "shared/icons/bookmark" %>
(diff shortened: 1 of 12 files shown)


app/views/items/_for_later_list_item_show.html.erb
```diff
@@ -1,11 +1,17 @@
 <% if FeatureFlags.for_later_lists_enabled? %>
   <span id="for_later_list_item_show">
     <% if for_later_list_item.present? %>
-      <%= button_to "Don't need it", account_for_later_list_item_path(for_later_list_item),
-            method: :delete, class: "btn" %>
+      <%= button_to account_for_later_list_item_path(for_later_list_item),
+            method: :delete, class: "btn remove-from-save-for-later-button", title: "Remove from save for later list" do %>
```
phinze, line 5: Small mismatch: `title="Remove from save for later list"` but the visually-hidden span says `"Remove from saved for later"`. Same in `_for_later_list_items_index.html.erb`. Picking one keeps things tidy for screen reader users.

phinze (approved):
Took it for a spin locally. The bookmark swap feels lighter and friendlier than a row of "Don't need it" buttons. The inline SVG partial reads as the right call: the feather-icon helper made fill/stroke styling fiddly, so keeping the markup inline trades a tiny bit of weirdness for a lot less SCSS pain.

A couple of UX things felt like follow-on polish rather than blockers:

- Removing from the for-later list is a little sudden. No confirm, no undo. Maybe a flash with an undo, or a confirm step.
- The bookmark's destination isn't very discoverable. A member could save things and not realize there's a page waiting for them.

Happy to file follow-ups if useful. One tiny inline below.

app/views/items/_for_later_list_item_show.html.erb
crismali: good catch!

crismali:
Thanks for the feedback! I've gone ahead and added a quick confirm when removing an item from the saved for later index page. I think I can put together a quick toast message for adding/removing from other pages and may add that functionality to the saved for later index page as well when I get a moment.

crismali:
I've added flash messages and set them up to work with turbo streams 👍

PR #2189, opened 2026-06-10
Add a new item search page

### LINKED ISSUE #1741: Sortable/filterable item listing for staff
Start with items. Columns we want to have:

* Smaller image (thumbnail) (no sorting)
* Item number
* Status
* Name
* Size/Strength
* \# of holds
* Brand/Model
* Category (no sorting)

We want to be able to click to sort by that column.

We'd like to be able to search specific fields for a value. It's OK to do this inline in the table or to use a different page to setup the search. We can also start with a text-based search syntax like `name:"drill" size:"1/2"` and iterate from there.

jim:
# What it does

Adds a new admin page with a way to search items by typing queries in a text field. By default, queries are searched in all columns. They can be scoped to specific fields using syntax like `brand=dewalt`.

# Why it is important

Fixes https://github.com/chicago-tool-library/circulate/issues/1741.

# UI Change Screenshot

Screenshot 2026-06-09 at 7 14 43 PM

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -229,6 +229,7 @@ GEM
       net-http-persistent (>= 4.0.4, < 5)
     faraday-retry (2.3.2)
       faraday (~> 2.0)
+    ffi (1.17.3-aarch64-linux-gnu)
     ffi (1.17.3-arm64-darwin)
     ffi (1.17.3-x86_64-darwin)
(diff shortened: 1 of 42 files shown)


crismali (approved):
looks good, I'm wondering about the Gemfile.lock changes though. They seem like a weird artifact of something but I'm not sure what.

PR #2190, opened 2026-06-10
Add an oldest item column to overdue items report

jim:
# What it does

Adds a new column to the Members with Overdue Items report that lists the oldest due date for that member's items. This column is also added to the CSV that we export for external analysis.

# Why it is important

Staff feature request to make it easier to sort the exported CSV.


phinze (approved):
looks good! nice and straightforward

PR #2197, opened 2026-06-24
Don't emit logs while running tests

jim:
# What it does

This makes it so running tests isn't so noisy. I'm leaving them on in dev/prod while we sort out how things actually work.


phinze (approved):
👍 easy win

PR #2207, opened 2026-07-23
Remove the legacy admin appointments view

### LINKED ISSUE #2204: Remove the legacy admin appointments view

Production currently has `FEATURE_NEW_APPOINTMENTS_PAGE=on`, but the old appointments view remains available through the `?new=false` override. Every appointment UI change therefore has to be implemented across parallel templates, styles, Turbo responses, and tests even though normal production traffic uses only the newer view. Issue #2201 surfaced this again by requiring the same behavior in both implementations.

Make the current appointments page the only admin appointments view. Remove the feature flag and query parameter override, delete `index_orig.html.erb` and `_appointment_orig.html.erb`, and collapse the pull and completion controllers onto their current Turbo responses.

The remaining templates should no longer pass `new=true`, and the legacy `.admin-appointments-orig` styles and feature-flag-specific tests can be removed. The current desktop and mobile behavior, including moving appointments between pending, pulled, and completed sections, should remain covered.

**Additional context**

The feature flag was introduced in #1416. This cleanup was identified while implementing #2201.

phinze:
# What it does

Removes the `FEATURE_NEW_APPOINTMENTS_PAGE` flag and `?new=false` escape hatch, deletes the original appointments templates and styles, and collapses appointment pull and completion rendering onto the current responsive view. Appointment actions no longer carry the `new` parameter.

# Why it is important

Production already uses the responsive appointments view. Keeping the original implementation meant changes like #2205 had to be built and maintained across two parallel sets of templates, styles, controller branches, and tests.

Closes #2204

# UI Change Screenshot

No visual change. The existing responsive appointments page is now unconditional.

# Implementation notes

This PR is stacked on #2205 so the cleanup can delete the legacy code touched there. Once #2205 merges, this branch can be rebased onto `main`.

The core appointment system tests were previously forced onto the legacy view by `.env.test`. They now exercise the supported desktop markup and continue to cover moving appointments between pending, pulled, and completed sections through Turbo.


jim (commented):
Love a bit of cleanup!

jim (approved):

PR #2220, opened 2026-08-02
Refactor item status code and add to item search page

jim:
* Consolidates most of the status display and search code into a single file and names a concept that was always present: borrow status (which is if an item is available, checked out, etc). There are less nested conditionals in a view helper as a result. It also means that we're more consistent about how we were rendering status values for admins.
* Creates two new helper methods to make working with the statuses in views clearer. We now have `item_status_label`, `member_item_status_label`, and `borrow_status_label`.
* Adds tooltips to the two status labels in admin views to make it easier to understand what they mean.
* Finally, adds these status labels to the new item search view.


app/views/admin/items/searches/show.html.erb
```diff
@@ -61,8 +61,9 @@
             <td>
               <%= link_to item.name, admin_item_path(item) %>
             </td>
-            <td>
-              <%= item.status.humanize %>
+            <td class="no-wrap">
```
phinze, line 64: Tiny thing: this `no-wrap` doesn't actually attach. The only rule for it is `.items-search th.no-wrap` in `admin.scss`, so a `td` slips right past. Adding `td.no-wrap` to that selector should do it.

phinze (approved):
Great cleanup! Separating borrow status from item status makes everything a lot clearer.

PR #2222, opened 2026-08-02
Create pending memberships when a member chooses "in person" on the payment screen

### LINKED ISSUE #2168: Show a different message when partway through a renewal
When you go through the renewal process, and have completed it and gone back to the membership page, it still shows the same "your membership has expired" box. Might be because the new membership hasn't been started yet.

In any case, it shouldn't be there. I tested this on staging with the "pay in person" button and could reproduce it.

jim:
Fixes #2168 by resolving a gap in how we were modeling the membership signup and renewal process. After a member completed renewal "in person" (#skip), no membership record was created, so the "your membership expired" banner never cleared. Fixing that surfaced two follow-on issues — how staff finish those renewals, and a payment double-charge — which this branch also addresses.

The main change is that we always create a `Membership` object when going through either the signup or renewal flows (previously, we didn't if you chose the "pay in person" option). Unifying the flow removes a bunch of edge cases from the app going forward.

### DIFF
--- a/app/controllers/renewal/payments_controller.rb
+++ b/app/controllers/renewal/payments_controller.rb
@@ -30,7 +30,10 @@ module Renewal
     end
 
     def skip
-      # completing in person
+      # completing in person: create a pending membership so the renewal banner
+      # clears right away. Staff record the payment and start it when the member
+      # comes in to pay.
+      Membership.create_for_member(@member, start_membership: false) unless @member.pending_membership
       MemberMailer.with(member: @member).renewal_message.deliver_later
       redirect_to renewal_confirmation_url, status: :see_other
     end
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -99,7 +99,10 @@ class Membership < ApplicationRecord
       membership = member.memberships.create!(library: member.library, membership_type:)
     end
 
-    if amount > 0
+    # Record the payment at most once per membership. A membership can reach here
+    # already paid — e.g. an online signup creates a pending, paid membership that
+    # staff later start in person — and must not be charged a second time.
+    if amount > 0 && !Adjustment.exists?(adjustable: membership)
       Adjustment.record_membership(membership, amount)
       Adjustment.record_member_payment(member, amount, source, square_transaction_id)
     end
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -87,8 +87,11 @@ class Membership < ApplicationRecord
     if start_membership
       if (membership = member.pending_membership)
         # Finalize a membership that was left pending, e.g. a renewal the member
-        # chose to complete in person.
-        membership.start!(now)
+        # chose to complete in person. If the member renewed early and still has
+        # an active membership, start this one when that one ends so the dates
+        # don't overlap; otherwise start it now.
+        start_date = member.active_membership&.ended_at || now
+        membership.start!(start_date)
       else
         start_date = next_start_date_for_member(member, now: now)
         raise PendingMembership.new("member with pending membership can't start a new membership") unless start_date
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -85,10 +85,16 @@ class Membership < ApplicationRecord
     membership_type = member.memberships.present? ? "renewal" : "initial"
 
     if start_membership
-      start_date = next_start_date_for_member(member, now: now)
-      raise PendingMembership.new("member with pending membership can't start a new membership") unless start_date
+      if (membership = member.pending_membership)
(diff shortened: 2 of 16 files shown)


crismali (approved):

app/controllers/renewal/payments_controller.rb
```diff
@@ -30,7 +30,10 @@ def create
     end
 
     def skip
-      # completing in person
+      # completing in person: create a pending membership so the renewal banner
+      # clears right away. Staff record the payment and start it when the member
+      # comes in to pay.
+      Membership.create_for_member(@member, start_membership: false) unless @member.pending_membership
```
phinze, line 36: I think this creates a 500 on the member's own profile page for the early-renewal case. After this runs, a member can have an active membership *and* a pending one, and since `last_membership` orders `ended_at DESC NULLS FIRST`, the pending row becomes `last_membership`. `app/views/account/members/show.html.erb:49` does `date_with_time_title(@member.last_membership.ended_at)` inside `if @member.active_membership`, so `ended_at` is nil and `nil.to_fs(:short_date)` raises. The admin equivalent in `_member_details.html.erb` is fine because `display_date` nil-guards.

The fix is likely just switching that view to `@member.active_membership.ended_at`, since it's already inside the `if @member.active_membership` branch. CI doesn't catch it because the new system test visits `account_home_url` (loans), not `account_member_url`.

test/system/membership_renewal_test.rb
```diff
@@ -99,6 +99,14 @@ def complete_square_checkout
       assert_includes html, "Thank you for renewing"
       refute_includes html, "Your payment of"
     end
+
+    @member.reload
+    assert_equal 2, @member.memberships.count
+    assert @member.pending_membership, "expected a pending membership after completing in person"
+
+    # The renewal banner should no longer appear now that a (pending) renewal exists.
+    visit account_home_url
```
phinze, line 108: Worth a conscious call: the same mechanism that clears the banner here also suppresses it indefinitely. If the member picks "in person" and never comes in, their active membership lapses, `borrow?` goes false, and the account page shows no renewal prompt because the pending row is still `last_membership`. That's probably the right tradeoff (the banner would just loop them into a flow that no-ops now), but it means staff are the only ones who'll notice a stale pending renewal. If that bothers you, a small follow-up could show a different message when `last_membership.pending?` ("your renewal is pending until you pay in person") so the member still gets a nudge without being sent back through the renewal flow.

app/models/membership.rb
```diff
@@ -85,15 +85,27 @@ def self.create_for_member(member, amount: 0, source: nil, start_membership: fal
...
+      end
     else
       membership = member.memberships.create!(library: member.library, membership_type:)
     end
 
-    if amount > 0
+    # Record the payment at most once per membership. A membership can reach here
+    # already paid — e.g. an online signup creates a pending, paid membership that
+    # staff later start in person — and must not be charged a second time.
+    if amount > 0 && !Adjustment.exists?(adjustable: membership)
```
phinze: Small thing: when this guard trips with `amount > 0`, the payment is dropped silently, with no adjustment and no error. The view hides the payment radios when the pending membership is already paid, so reaching this takes a hand-crafted request, and it's fine as defense in depth. If you'd rather it be loud, raising here would make a misrouted cash payment impossible to lose quietly.

app/models/membership.rb
```diff
@@ -85,15 +85,27 @@ def self.create_for_member(member, amount: 0, source: nil, start_membership: fal
...
+        # chose to complete in person. If the member renewed early and still has
+        # an active membership, start this one when that one ends so the dates
+        # don't overlap; otherwise start it now.
+        start_date = member.active_membership&.ended_at || now
+        membership.start!(start_date)
+      else
+        start_date = next_start_date_for_member(member, now: now)
+        raise PendingMembership.new("member with pending membership can't start a new membership") unless start_date
+
+        membership = member.memberships.create!(started_at: start_date, ended_at: start_date + 365.days, library: member.library, membership_type:)
```
phinze: Not new to this PR, but now that the two branches sit side by side: `start!` gives 364 days and this path gives 365. Might be a nice moment to pick one.

phinze (approved):
Looks good overall! A few inlines, only important one is that 500 one.

app/models/membership.rb
jim: Good call, how 'bout we go with `1.year` instead!

test/system/membership_renewal_test.rb
jim: Added a message for the member to remind them to come in to complete the process.

PR #2225, opened 2026-08-13
Limit C-tools per member across holds and loans

### LINKED ISSUE #1850: Limit the number of items on an appointment
> Would it be possible to limit the number of items on an appointment? Ideally we would like to make the limits:
1 "C" tool
10 "B" tools
no limit on A tools

* We'll need a way to limit the number of items with a given borrow policy on a single appointment.
    * Add a new `int` field to `BorrowPolicy` called `per_appointment_limit`. A value of `0` will mean no limit. The default is `0` (this can be set in the database).
    * Update appointment hold validations to take this into account. Make sure the correct error is returned in the controller. We want this limitation to only apply to members.

phinze:
# What it does

Limits members to two C-tools total across active holds and checked-out loans. When someone reaches the limit, the blocked hold message links to the C-tools currently counting toward it and says whether each one is checked out or on hold.

The limit is configurable per borrow policy, with `0` meaning unlimited. This migration sets C-tools to 2 and leaves other policies unlimited. Librarian-created holds remain exempt for special cases.

# Why it is important

Members could reserve more C-tools than the lending policy allows, then only find out when librarians reviewed the day's orders. Enforcing the simple “2 C-tools at a time” rule when a member places a hold prevents that surprise without adding a second appointment-level rule.

Closes #1850.

# UI Change Screenshot

c-tool-hold-limit-updated

# Implementation notes

The validation counts active holds and checked-out loans using the item's borrow policy. Returned loans, inactive holds, and items under other policies do not count. Renewals keep the same loan in place and do not change the total.

Member hold creation locks the member row while checking the limit, so simultaneous requests cannot both slip through. The flash lists at most five items and summarizes any remainder to keep the cookie-backed session bounded.

Staff bypass is based on the hold creator's role, so holds placed through the librarian workflow can exceed the configured limit.

### DIFF
--- a/app/controllers/account/holds_controller.rb
+++ b/app/controllers/account/holds_controller.rb
@@ -1,5 +1,7 @@
 module Account
   class HoldsController < BaseController
+    LIMIT_MESSAGE_ITEM_COUNT = 5
+
     include Pagy::Backend
 
     def index
@@ -22,7 +24,12 @@ module Account
           @new_hold.start! if @new_hold.ready_for_pickup?
           redirect_to item_path(@item), success: "Hold placed.", status: :see_other
         else
-          redirect_to item_path(@item), error: "Something went wrong!", status: :see_other
+          error = if @new_hold.errors.of_kind?(:base, :maximum_items_per_member)
+            borrow_policy_limit_message(@new_hold)
+          else
+            @new_hold.errors.full_messages.to_sentence
+          end
+          redirect_to item_path(@item), error:, status: :see_other
         end
       end
     end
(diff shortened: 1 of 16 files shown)


app/models/hold.rb
```diff
@@ -31,7 +31,7 @@ class Hold < ApplicationRecord
     message: "is required when started_at is set"
   }, if: :started_at?
 
-  validate :ensure_items_are_holdable, on: :create
+  validate :ensure_items_are_holdable, :ensure_member_is_within_borrow_policy_limit, on: :create
```
crismali, line 34: Is there a situation where a librarian might want to create a hold for a member even if it violates this rule?

app/controllers/account/holds_controller.rb
```diff
@@ -37,5 +44,28 @@ def destroy
...
+    private
+
+    def borrow_policy_limit_message(hold)
+      borrow_policy = hold.item.borrow_policy
+      tool_name = "#{borrow_policy.code}-Tool".pluralize(borrow_policy.maximum_items_per_member)
+      introduction = helpers.tag.p(
+        "You can have a maximum of #{borrow_policy.maximum_items_per_member} #{tool_name} checked out or on hold at one time. These items count toward your limit:"
+      )
+      counted_items = hold.existing_items_counted_toward_limit
+      items = counted_items.first(LIMIT_MESSAGE_ITEM_COUNT).map do |item_holder|
```
crismali, line 57: Just curious, but why not just show a message for every counted item?

crismali (approved):
looks good, just a couple of questions

app/models/hold.rb
phinze: Yep, that’s exactly the exception here. The validation skips the limit when the hold creator has the staff role, and the admin hold flows set the librarian as creator. We’ve got coverage for both the model behavior and the librarian workflow.

app/controllers/account/holds_controller.rb
phinze: Mostly to keep the flash bounded. It’s stored in our cookie-backed session, which tops out at 4 KB. Staff can intentionally put a member over the limit, and existing data may already be over it, so rendering every linked item could turn this friendly error into a `CookieOverflow`. The usual two-C-tool case still shows both items; the five-item cap only affects outliers.

